PR #1, opened 2018-06-11
Add SOURCE_COMMIT env.

schmmd:


schmmd:
@joelgrus the extra commits were for triggering TeamCity.

PR #5, opened 2018-06-22
remove /build directory from git repo

joelgrus:
oyvindt pointed this out to me


schmmd:
Yikes!

@OyvindTafjord OT FTW!

PR #9, opened 2018-08-02
Dependency demo

DeNeutoy:
Adds:

- A demo for the dependency parser
- `DemoModel`, which was removed from allennlp by accident when we removed the demo
- Refactor to structure the demo as a proper python module, so we don't need to mess with paths
- Instructions to run the demo locally


DeNeutoy:
This is what it looks like at the moment:

image

image

I will make it a bit prettier, but none of the changes will affect this PR (they will all be in the `Predictor` in allennlp), so this is ready for review

joelgrus:
it looks like the CI is failing because maybe the demo wants things that aren't in the latest release? how do we manage that?

schmmd:
@joelgrus we just need to update the `FROM` to the specific commit we want to depend on: https://hub.docker.com/r/allennlp/commit/tags/

joelgrus:
maybe we should have tagged commits that point to tagged releases and everything else point to `:latest`?

schmmd:
I don't like the idea of everything pointing to `:latest` because then changes to github.com/allenai/allennlp could break the demo without CI catching it.  If we wanted to go that route--I think it would take some more thought and changes to our CI.

PR #15, opened 2018-08-20
Update allennlp dependency to fix dependency parser

schmmd:


schmmd:
@matt-gardner thanks--although I guess it doesn't really matter since @DeNeutoy doesn't want what's currently staged to be released either.

schmmd:
@DeNeutoy I merged this.  I'll have you double check that staging looks good before a production release.

PR #22, opened 2018-08-22
NER Facelift / Introducing <Highlight />

aaronsarnat:
ner

This PR:

- Introduces general `<Highlight />` component
- Updates NER demo, leverages new `<Highlight />` component

### DIFF
--- a/demo/src/components/NamedEntityComponent.js
+++ b/demo/src/components/NamedEntityComponent.js
@@ -169,11 +169,11 @@ class NerOutput extends React.Component {
 
     // Defining an empty array for building a list of formatted token objects.
     let formattedTokens = [];
-    // Defining an empty string to store temporary span text.
+    // Defining an empty string to store temporary span text (this field is used to build up the entire text in a single BIL span).
     let spanStr = "";
     // Iterate through array of tags from response data.
     tags.forEach(function (tag, i) {
-      // Defining an empty object var to store temporary token data.
+      // Defining an empty object to store temporary token data.
       let tokenObj = {};
       if (tag === "O") { // If this tag is not part of an entity:
         // Build token object using this token's word and set entity to null.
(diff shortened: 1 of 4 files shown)


demo/src/components/NamedEntityComponent.js
```diff
@@ -113,117 +113,109 @@ class NerInput extends React.Component {
...
-
+    // "B" = "Beginning" (first token in a sequence of tokens comprising an entity)
+    // "I" = "Inside" (token in a sequence of tokens (that isn't first or last in its sequence) comprising an entity)
+    // "L" = "Last" (last token in a sequence of tokens comprising an entity)
+    // "O" = "Outside" (token that isn't associated with any entities)
+    // "U" = "Unit" (A single token representing a single entity)
+
+    // Defining an empty array for building a list of formatted token objects.
+    let formattedTokens = [];
+    // Defining an empty string to store temporary span text.
```
schmmd: I would expand on your comment here.  Specifically this field is used to build up the entire text in a single BIL span.

demo/src/components/NamedEntityComponent.js
```diff
@@ -113,117 +113,109 @@ class NerInput extends React.Component {
...
+        formattedTokens.push(tokenObj);
+      } else if (tag[0] === "U") { // If this tag is a unit token:
+        // Build token object using this token's word and entity.
+        tokenObj = {
+          text: words[i],
+          entity: tag.slice(2) // tag value with "U-" stripped from the beginning
+        }
+        // Append array of formatted token objects with this token object.
+        formattedTokens.push(tokenObj);
+      } else if (tag[0] === "B") { // If this tag is beginning of a span:
```
schmmd: This is a fine solution.  I would have built the nested loop solution, but you could argue either approach.  Each has its benefits.

aaronsarnat: Cool, thanks! :) Mainly the reason I didn't do the nested loop is I couldn't get it to work right on the first try, so I gave this approach a shot and it worked.

demo/src/components/NamedEntityComponent.js
aaronsarnat: Will do. I'll make this change and then merge.

aaronsarnat: Done

aaronsarnat:
@schmmd I addressed your comment about the comment. Tested again just now and everything seems to work. All green checks, so I'm going to merge...

PR #25, opened 2018-08-23
Adding (hidden) demo for wikitables parser

matt-gardner:
This adds a demo for the wikitables parser.  It's currently hidden, as it could be made to look nicer, and we're still wanting to improve the model a bit.  I'm doing this now so that @kl2806 has an example to look at for building a demo for his in-progress ATIS model.

### DIFF
--- a/app.py
+++ b/app.py
@@ -40,6 +40,7 @@ class ServerError(Exception):
 
     def __init__(self, message, status_code=None, payload=None):
         Exception.__init__(self)
+        logger.error(message)
         self.message = message
         if status_code is not None:
             self.status_code = status_code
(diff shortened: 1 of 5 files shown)


matt-gardner:
@schmmd, who's the right person to review this PR?  And is there a way to see the demo that's produced by this, or does that just go to staging after the PR is merged?

schmmd:
@matt-gardner you can assign it to me or Joel.  I wasn't sure it was ready for review.

Will we need the JVM dependency when we ultimately have the semantic parser demo?  ;-(

It's pushed to staging once it's merged.  If you want to see the demo, it's best to run the Dockerfile locally.

matt-gardner:
Yes, we'll need the jvm dependency, as it's required for wikitables.  I'm surprised it wasn't already there - we're using `Dockerfile.pip` for what gets pushed to the `allennlp` docker repository?

schmmd:
@matt-gardner Dockerfile.pip just installs AllenNLP via pip.  We do bake the JVM into our Dockerfile that is used for continuous integration.

app.py
```diff
@@ -40,6 +40,7 @@ class ServerError(Exception):
 
     def __init__(self, message, status_code=None, payload=None):
         Exception.__init__(self)
+        logger.error(message)
```
schmmd: Was this for debugging?

schmmd: Great!

PR #27, opened 2018-08-23
update parser with working one

DeNeutoy:


DeNeutoy:
The parser in the demo uses Penn TreeBank (stanford) style dependencies. Here is a parser which uses Universal Dependencies:

https://s3-us-west-2.amazonaws.com/allennlp/models/biaffine-dependency-parser-ud-2018.08.23.tar.gz

It works in exactly the same way, you could just switch the model location and it would work.

schmmd:
@DeNeutoy cool, thanks.  I'm still inclined to remove the spans so the text is guarenteed to be in order--but I'll play around with staging a bit first.

DeNeutoy:
@schmmd you will find it really hard I think to find an example which will be non-projective for english now the parser is correct - they seem like very heavy handed english, e.g:

John saw a dog yesterday which was a yorkshire terrier.

But now in 98% of cases it should be ok

PR #30, opened 2018-08-24
UI Improvements to Coref Demo

aaronsarnat:
image

---

This PR
---

- Updates Coref demo experience to use new `<Highlight />` component
- Improves `<Highlight />` component


aaronsarnat:
@schmmd assigning to you for review. FYI I have some minor CSS tweaks I still want to make before I merge this (esp. relating to nested refs), but I figured I could have you start looking at the logic while I make small edits to the stylesheet.

demo/src/components/CorefComponent.js
```diff
@@ -116,62 +116,133 @@ class CorefOutput extends React.Component {
...
-      var clusteredDocument = doc.map((word, wordIndex) => {
-        var membershipClusters = [];
-        clusters.forEach((cluster, clusterIndex) => {
-          cluster.forEach((span) => {
-            if (wordIndex >= span[0] && wordIndex <= span[1]) {
-              membershipClusters.push(clusterIndex);
+      const { activeIds, isClicking, selectedId } = this.state;
+      const { tokens, clusters } = this.props;
+
+      // Span tree data transform code courtesy of Michael S.
```
schmmd: Consider putting this code into a helper function that just transforms the JSON structure.  That would make it easier to test if we find any problems.

aaronsarnat: good idea. i'll take a look at that

aaronsarnat:
@schmmd PTAL. I feel good about the state this PR is in now and am ready to merge, but thought it was worth getting another look at it from you (or whoever) since I made a number of changes since you approved this:

- Fixed nested highlight interaction bug where the wrong cluster was being selected
- Refactored highlight component structure to support cleaner-looking layout (side-aligned labels now fill height of container, things aren't quite as scrunched, and the new structure should be more robust when dealing with edge cases) 
- Modularized the `<Highlight />` component a bit more
- Browser tested on the demo's most used browsers

---

Here's a screen video of the latest (sped up a little):

highlight-ux

PR #32, opened 2018-08-27
Attempt to fix broken highlight styling by un-modularizing CSS

aaronsarnat:
This PR moves the styles from the modularized highlight CSS file to main.css along with all the other CSS for the demo.

Something about modularizing CSS (by importing a css file at the component level) created a broken `<link>` tag (resource returned 404) instead of an inline `<style>` tag. Hunch is that it's some outdated React-related dependency in the Docker file.


schmmd:
Did you verify that this fixes the locally @aaronsarnat ?  You can follow the `docker` commands in the README to bring up an environment.

aaronsarnat:
@schmmd following the [`README`](https://github.com/allenai/allennlp-demo/blob/master/README.md) instructions (both for Docker method and local Conda environment method), this is what I'm seeing locally:

coref

ner

Issue #14, opened 2018-08-09
Enter doesn't submit a query to process results


schmmd:
You need to manually click on "Run".  Ideally you could hit enter as well.

schmmd:
Here's how we fixed this in a similar repo: https://github.com/allenai/demo-me-elmo/commit/f7a8fb927f6894a276274999d9932b18c6c2a657#diff-1f5d33d628ea1b5f92fa869e4ad9ae23R58

PR #41, opened 2018-08-28
Adding Enter-to-submit capability on all demos

aaronsarnat:
Fixes [Enter doesn't submit a query to process results (#14)](https://github.com/allenai/allennlp-demo/issues/14)

---

Tested all fields on all demos.

### DIFF
--- a/demo/src/components/Button.js
+++ b/demo/src/components/Button.js
@@ -4,18 +4,16 @@ import React from 'react';
   <Button /> Component
 *******************************************************************************/
 
-class Button extends React.Component {
-    render() {
-      const { enabled, runModel, inputs } = this.props;
+export default class Button extends React.Component {
+  render() {
+    const { enabled, runModel, inputs } = this.props;
 
-      return (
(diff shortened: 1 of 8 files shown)


schmmd:
Cool, looks good but I wonder--is there any way to handle this in a more general fashion so we don't need this code for each page?  @joelgrus any ideas here?

aaronsarnat:
@schmmd that thought crossed my mind as well, but there is already a ton of duplicated code in the same vein, and I wasn't sure if I should try to introduce a new paradigm for this task. We don't really have a place for general util functions currently, but we probably should.

TBH I think the whole demo front-end could use a refactor / refresh, as it's a bit of a mess currently, but that seems like a separate effort that I believe you were saying we probably don't want to worry about in Q3.

That said, if you want me to work on generalizing this particular function, I am happy take another stab.

joelgrus:
I agree that (1) this code probably should be written only once and (2) that probably can't happen until we do a larger refactor

schmmd:
Thanks--if you have any specific ideas on a refactor I'd love to hear them :https://github.com/allenai/allennlp-demo/issues/43

PR #45, opened 2018-08-31
Multi NER

DeNeutoy:
This is a first step to getting multiple model predictions in the same demo. There's one thing I don't know how to do - Make the model selection dropdown "sticky" so that it doesn't refresh.

The fine-grained NER model is currently SOTA I believe, with dev set performance on Ontonotes 5.0 of 88.2 F1.

image

### DIFF
--- a/demo/src/components/Button.js
+++ b/demo/src/components/Button.js
@@ -6,10 +6,10 @@ import React from 'react';
 
 export default class Button extends React.Component {
   render() {
-    const { enabled, runModel, inputs } = this.props;
+    const { enabled, runModel, inputs, modelEndpoint} = this.props;
 
     return (
-    <button id="input--mc-submit" type="button" disabled={!enabled} className="btn btn--icon-disclosure" onClick={ (event) => { runModel(event, inputs) }}>Run
+    <button id="input--mc-submit" type="button" disabled={!enabled} className="btn btn--icon-disclosure" onClick={ (event) => { runModel(event, inputs, modelEndpoint) }}>Run
       <svg>
         <use xlinkHref="#icon__disclosure"></use>
       </svg>
--- a/demo/src/components/NamedEntityComponent.js
+++ b/demo/src/components/NamedEntityComponent.js
@@ -351,7 +351,6 @@ class _NerComponent extends React.Component {
     this.setState({outputState: "working"});
 
     var payload = {sentence: inputs.sentenceValue};
-    console.log(modelEndpoint)
     fetch(`${API_ROOT}/predict/${modelEndpoint}`, {
       method: 'POST',
       headers: {
--- a/demo/src/components/NamedEntityComponent.js
+++ b/demo/src/components/NamedEntityComponent.js
@@ -22,6 +22,12 @@ const nerSentences = [
   "When I told John that I wanted to move to Alaska, he warned me that I'd have trouble finding a Starbucks there."
 ];
 
+const nerModels = {
+  "ner": "named-entity-recognition",
+  "fine-grained-ner": "fine-grained-named-entity-recognition"
+};
+
+
 const title = "Named Entity Recognition";
 const description = (
   <span>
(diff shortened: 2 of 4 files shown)


DeNeutoy:
FYI I didn't permanently add a "mute" sign on top of the demo. I considered it but in the end I couldn't get it to fit with the rest of the UI theme. Perhaps something for Q4.

demo/src/components/NamedEntityComponent.js
```diff
@@ -254,12 +347,12 @@ class _NerComponent extends React.Component {
...
   }
 
-  runNerModel(event, inputs) {
+  runNerModel(event, inputs, modelEndpoint) {
     this.setState({outputState: "working"});
 
     var payload = {sentence: inputs.sentenceValue};
-
-    fetch(`${API_ROOT}/predict/named-entity-recognition`, {
+    console.log(modelEndpoint)
```
joelgrus: get rid of the console.log ?

demo/src/components/Button.js
```diff
@@ -6,10 +6,10 @@ import React from 'react';
 
 export default class Button extends React.Component {
   render() {
-    const { enabled, runModel, inputs } = this.props;
+    const { enabled, runModel, inputs, modelEndpoint} = this.props;
 
     return (
-    <button id="input--mc-submit" type="button" disabled={!enabled} className="btn btn--icon-disclosure" onClick={ (event) => { runModel(event, inputs) }}>Run
+    <button id="input--mc-submit" type="button" disabled={!enabled} className="btn btn--icon-disclosure" onClick={ (event) => { runModel(event, inputs, modelEndpoint) }}>Run
```
joelgrus: it is absolutely horrific that you can just pass extra arguments to a javascript function and it will ignore them

DeNeutoy: I know

DeNeutoy: I just tried this and it worked, is there a better way? I kind of feel like a button shouldn't own this state anyway, like it should just be automatically piped from the props of the parent component or something

joelgrus: I think it's good for now, when we refactor everything we should just do so in a way that solves this problem (`runModel` in a base class, everyone has to tell it what endpoint to use, etc...)

aaronsarnat: Re: refactor, we could consider integrating TypeScript. All the cool kids seem to be doing it nowadays.

Nontrivial change, but if we're rewriting the app anyway, that would be a good time to consider it. It would help take some of the mystery out of things like passing extra arguments that sometimes get ignored.

demo/src/components/NamedEntityComponent.js
```diff
@@ -151,7 +178,73 @@ class TokenSpan extends React.Component {
       "MISC": {
         tooltip: "Miscellaneous",
         color: "gray"
-      }
+      },
+      "PERSON": {
```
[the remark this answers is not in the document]
DeNeutoy: The problem is, the model i've added has a different set of tags - it's easier to just add them all here rather than try to dis-ambiguate them

DeNeutoy:
@aaronsarnat The sticky thing I was talking about:
When I choose a model from the dropdown and run it on some input, I have to select the model again if I want to run another sentence. It seems natural that this should stay locked when a user selects it, but I don't know how to do that. I'll merge this and open an issue.

matt-gardner:
@DeNeutoy, for a dropdown, you need to have which model is selected be part of the react component's state (I didn't look too closely at the code, but I think you might have done this), and you need to have that state get rendered when you create the select box.  I think that's as simple as making one of the options `selected`.

demo/src/components/NamedEntityComponent.js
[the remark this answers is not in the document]
aaronsarnat: :+1: Fair enough, I was just curious.

demo/src/components/NamedEntityComponent.js
```diff
@@ -102,8 +116,20 @@ class NerInput extends React.Component {
     const callHandleKeyDown = (e) => { this.handleKeyDown(e, nerInputs)};
 
     return (
-      <div className="model__content">
+      <div className="model__content" >
         <ModelIntro title={title} description={description} />
+
+        <div className="form__instructions"><span>Choose Model: </span>
+          <select disabled={outputState === "working"} onChange={this.handleModelChange}>
+            <option>default ner model</option>
```
aaronsarnat: image

@DeNeutoy do you actually want these 3 options? Seems like you'd only want the NER and fine-grained options where NER is selected by default.

schmmd:
Just another heads up--the demo restarts due to a memory leak.  It becomes more and more pressing to address this issue with each extra model.  I'll certainly argue for prioritizing this in Q4--but in the meantime we might need to increase machines size / revert a model if we cross a threshold.

PR #48, opened 2018-09-08
Add Open Information Extraction to the demo

gabrielStanovsky:
Follows closely the SRL demo, with the addition of multi-word predicates.

I wasn't sure if should also add a test to `server_test.py`, is it deprecated?


schmmd:
@gabrielStanovsky `server_tests.py` is still run by CI.  Ideally you would add a test there similar to `test_textual_entailment`.

PR #55, opened 2018-09-14
Add note about Docker memory limit to demo README

aaronsarnat:


aaronsarnat:
Thanks again @DeNeutoy for figuring out what was preventing me from running the demo! Assigning to you for review.

PR #66, opened 2018-09-18
Hide Event2Mind demo in main menu.

brendan-ai2:
- We're not ready for it to be fully visible in prod.
- Direct URL still available to share.


brendan-ai2:
No big deal! Thanks!

PR #67, opened 2018-09-18
Fix the SHA in the info route.

schmmd:
Previously we were bringing in SOURCE_COMMIT from the base image instead of the build argument, and so the SHA we present corresponds to allenai/allennlp instead of allenai/allennlp-demo.


schmmd:
Works for me locally--but the real test will be whether staging ends up with the right SHA in the /info route.  Force-merging since this is a blocker for me presently--I'll address any CR comments that come in--I promise!

brendan-ai2:
LGTM!

schmmd:
Great--also reporting back that it worked ;-)

Issue #70, opened 2018-09-20
Fix bugs in Event2Mind demo.


schmmd:
Currently the demo has some glaring bugs.  We want to fix this so that it's somewhat usable--even though it won't be fully usable until we improve the design.

- [x] Currently the input text is lost when something is submitted.
- [x] On the right, the user can select between "Tree" and "Text" but that option does nothing.

aaronsarnat:
I can fix these

aaronsarnat:
@schmmd FYI there's a PR for fixing this: https://github.com/allenai/allennlp-demo/pull/71

PR #71, opened 2018-09-20
Fixing Event2Mind demo UI bugs

aaronsarnat:
This PR fixes [Fix bugs in Event2Mind demo. #70](https://github.com/allenai/allennlp-demo/issues/70)

### DIFF
--- a/demo/src/components/Event2MindComponent.js
+++ b/demo/src/components/Event2MindComponent.js
@@ -224,23 +224,28 @@ class _Event2MindComponent extends React.Component {
             sentence={sentence} />
         </PaneLeft>
         <PaneRight outputState={this.state.outputState}>
-          <ul className="visualization-types">
-            {Object.keys(VisualizationType).map(tpe => {
-              const vizType = VisualizationType[tpe];
-              const className = (
-                visualizationType === vizType
-                  ? 'visualization-types__active-type'
-                  : null
-              );
-              return (
-                <li key={vizType} className={className}>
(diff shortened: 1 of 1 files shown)


demo/src/components/Event2MindComponent.js
```diff
@@ -206,7 +206,7 @@ class _Event2MindComponent extends React.Component {
     const { requestData, responseData } = this.props;
     const { visualizationType } = this.state;
 
-    const sentence = requestData && requestData.sentence;
+    const sentence = requestData && requestData.source;
```
aaronsarnat: @brendan-ai2 FYI this is what was breaking the input text after running the model. This data structure uses the field name of 'source' instead of 'sentence'.

demo/src/components/Event2MindComponent.js
```diff
@@ -147,7 +147,7 @@ class Event2MindOutput extends React.Component {
 *******************************************************************************/
 
 const VisualizationType = {
-  TREE: 'Tree',
+  DIAGRAM: 'Diagram',
```
aaronsarnat: @brendan-ai2 what do you think of calling the visual tab `Diagram`? I'm open to suggestions...

demo/src/components/Event2MindComponent.js
brendan-ai2: Good catch. Thanks!

demo/src/components/Event2MindComponent.js
brendan-ai2: Sounds reasonable!

PR #73, opened 2018-09-24
Text to SQL demo, make WikiTables live, update allennlp commit

kl2806:


schmmd:
My only concern is adding a new model and increasing pressure on the memory leak.  That said, full steam ahead for now!

Issue #44, opened 2018-08-28
Update the Attention Visualization


aaronsarnat:
We have a visualization of attention, which has helped researchers significantly in debugging their models and understanding how they are working. The current demo is reasonable, but a bit rough.  Particularly with the doubly-nested folding menus that animate open and closed and longer words.

**Aspirational design**

[![image](https://user-images.githubusercontent.com/8367927/46327384-3d6be100-c5b6-11e8-8b7b-d50ce3ee2145.png)](https://drive.google.com/file/d/1WoCAjcUfT0SvZY2aNJuhoFeXGgnvo8ZR/view?usp=sharing)

---

**Issues:**
- [ ] Accordion UX is poor. We need a Drawer UX instead.
- [x] Hard-coded table header labels are causing overlap bugs
- [x] Need more space between between attention description and table

**Features:**
- [ ] sticky header row
- [x] heatmap needs to support linear vs log scale
- [ ] Improve hover tooltip UX (currently have to wait for tooltip which is annoying)

---

**Design Update** (10/1/18): see https://github.com/allenai/allennlp-demo/issues/44#issuecomment-426135733

schmmd:
We have a "hidden demo" available at http://demo.allennlp.org/wikitables-parser/MzAyNzMw.

@matt-gardner do you have any ideas for roughly how we could provide a better visualization here?

schmmd:
@matt-gardner is what's showing up in the "answer" box actually what you want to see for the answer?  I'm a bit confused by the output.

matt-gardner:
Yeah, something about the executor broke in between when I originally wrote the demo and putting it in here.  The answers are currently all wrong.

But, other than that, the content being shown in the demo is roughly the right thing to show, I think, it could just be presented in a nicer way.  Not really sure on designs for how to make it look better, though.

matt-gardner:
Oh, another important issue: one of the inputs is a table.  We're currently hacking something together to get a table input to the model, using a big text field, but if there were some nicer way of doing this, that'd be much better.

schmmd:
@matt-gardner thanks--that gives me enough to work on for now.  Just wanted to know whether you had something particular and different in mind.

schmmd:
@matt-gardner @aaronsarnat and I are pretty unsure how to best visualize the logical forms coming out of the semantic parser.  We would like to start by improving how they look and ignore the table input for now.

For reference, here's one output example that's also "formatted" by me (potentially incorrectly).

```
[...]

PR #77, opened 2018-09-25
Adding general <SyntaxHighlight /> component and implementing it in WikiTables demo

aaronsarnat:
image

---

**This PR:**

- Represents part 1 of 3 addressing [Visualize semantic parsing output (#44)](https://github.com/allenai/allennlp-demo/issues/44)
- Fixes [Add syntax highlighting to demo (#76)](https://github.com/allenai/allennlp-demo/issues/76)
- Implements syntax highlighting in `WikiTables` demo.
- Disables annoying spellcheck on table input `<textarea>` on `WikiTables` demo.

### DIFF
--- a/demo/src/components/WikiTablesComponent.js
+++ b/demo/src/components/WikiTablesComponent.js
@@ -155,10 +155,7 @@ class WikiTablesOutput extends React.Component {
           </div>
 
           <div className="form__field">
-            <label>Logical Form (Raw Data)</label>
-            <pre><code>{logicalForm}</code></pre>
-
-            <label>Logical Form (Formatted)</label>
+            <label>Logical Form</label>
             <SyntaxHighlight language="lisp">{logicalForm.split('fb:').join('').split('.').join('-')}</SyntaxHighlight>
           </div>
 
--- a/demo/src/components/WikiTablesComponent.js
+++ b/demo/src/components/WikiTablesComponent.js
@@ -3,10 +3,10 @@ import HeatMap from './heatmap/HeatMap'
 import Collapsible from 'react-collapsible'
 import { API_ROOT } from '../api-config';
 import { withRouter } from 'react-router-dom';
-import {PaneLeft, PaneRight} from './Pane'
-import Button from './Button'
-import ModelIntro from './ModelIntro'
-
+import {PaneLeft, PaneRight} from './Pane';
+import Button from './Button';
+import ModelIntro from './ModelIntro';
+import SyntaxHighlight from './highlight/SyntaxHighlight.js';
 
 /*******************************************************************************
   <McInput /> Component
@@ -156,7 +156,7 @@ class WikiTablesOutput extends React.Component {
 
           <div className="form__field">
             <label>Logical Form</label>
-            <div className="model__content__summary">{ logicalForm }</div>
+            <SyntaxHighlight language="lisp">{logicalForm}</SyntaxHighlight>
           </div>
 
           <div className="form__field">
--- a/demo/src/components/WikiTablesComponent.js
+++ b/demo/src/components/WikiTablesComponent.js
@@ -155,8 +155,11 @@ class WikiTablesOutput extends React.Component {
           </div>
 
           <div className="form__field">
-            <label>Logical Form</label>
-            <SyntaxHighlight>{logicalForm}</SyntaxHighlight>
+            <label>Logical Form (Raw Data)</label>
+            <pre><code>{logicalForm}</code></pre>
+
+            <label>Logical Form (Formatted)</label>
+            <SyntaxHighlight language="lisp">{logicalForm.split('fb:').join('').split('.').join('-')}</SyntaxHighlight>
           </div>
 
           <div className="form__field">
--- a/demo/src/components/WikiTablesComponent.js
+++ b/demo/src/components/WikiTablesComponent.js
@@ -156,7 +156,7 @@ class WikiTablesOutput extends React.Component {
 
           <div className="form__field">
             <label>Logical Form</label>
-            <SyntaxHighlight language="lisp">{logicalForm}</SyntaxHighlight>
+            <SyntaxHighlight>{logicalForm}</SyntaxHighlight>
           </div>
 
           <div className="form__field">
--- a/demo/src/components/WikiTablesComponent.js
+++ b/demo/src/components/WikiTablesComponent.js
@@ -124,7 +124,7 @@ render() {
             </div>
             <div className="form__field">
             <label htmlFor="#input--mc-passage">Table</label>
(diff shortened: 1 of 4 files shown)


demo/src/components/WikiTablesComponent.js
```diff
@@ -156,7 +156,7 @@ class WikiTablesOutput extends React.Component {
 
           <div className="form__field">
             <label>Logical Form</label>
-            <div className="model__content__summary">{ logicalForm }</div>
+            <SyntaxHighlight>{logicalForm}</SyntaxHighlight>
```
aaronsarnat: @matt-gardner this is how syntax highlighting is invoked. I didn't see any language identifier in the output data, but luckily the syntax highlighting library this branch is using has the ability to auto-detect language. Thus, we should be able to support SQL, Java, etc. without having to explicitly set the language.

aaronsarnat:
@matt-gardner @schmmd this is an incremental improvement to the semantic parsing demo. Attention visualization will be the bigger chunk of work, but that seems pretty well contained as its own thing.

As for the non-attention output, some issues I wanted to call out:

1. **Answer** - it seems the answer output is still broken. I did not attempt to monkey with this, as such things are above my pay grade. :)
    image

2. **Formatting** - While we have syntax highlighting, we don't have newline and indent formatting (a.k.a. "beautify" / "prettify" / "un-minify" processing). This is a nontrivial problem to solve on the UI side. I spent some time hunting for libraries that do this automatically, but came up short, unfortunately. How important do you think formatting is, and are we certain it would need to happen on the front-end as opposed to being expressed in the data?

3. **Tooltips** - There was [some talk](https://github.com/allenai/allennlp-demo/issues/44#issuecomment-422506260) of hover tooltips on some of the logical form content. This PR isn't attempting to add any of that. Can I assume that if we still want to add this functionality, it would need to come later, as there is currently no supporting data for that?

demo/src/components/WikiTablesComponent.js
matt-gardner: Looking at the code, looks like the answer is yes.  What happens if you set this to lisp?  Does it look any better?

matt-gardner:
1. I have fixed the `allennlp` library so it should be good, but I don't think the demo has been updated to use that commit yet.  I wouldn't worry about it at this point.  I'll be sure that that part's working before I making anything live.

2. I don't have a strong opinion on whether this should live in python or javascript; I'd probably do it where it's easier (preferably wherever a library already existing to do it).  If it's in python, it should a utility called in the `Predictor` class, not in the `Model` (just as a reference to anyone who might be trying to do this at some point in the future).  I wouldn't say this is super high priority for me.

3. This can definitely come later, and probably never.  I think it'd be more work than it's worth to make this happen.

schmmd:
2.  I think we should do this--but only if we can use an off the shelf library.  Since you couldn't figure out how to do it in javascript--let's not worry about it for now.

3.  Seems fine.

demo/src/components/WikiTablesComponent.js
aaronsarnat: @matt-gardner Funny you ask, I actually had [explicitly-set language](https://github.com/allenai/allennlp-demo/blob/e1ecc6758018f1c80a491c75331c34b2304d63fb/demo/src/components/WikiTablesComponent.js#L159) until my latest commit. :)

I removed it for two reasons:

1. I wanted to future-proof the UI for multi-language support (e.g. SQL, Java, etc.) without the need to add language identification to the `responseData`.

2. The auto-detect output is identical to the explicitly-set LISP output:

    image

<br>

That said, we can still do explicit language if that feels cleaner to you. In the short-term I can just hardcode it in like I had it before, but at some point we'll need to include language identification in the data if/when we want to support additional languages.

---

As for your comment about the quality of the highlighting, these are the style options provided by the library I'm currently using:

https://highlightjs.org/static/demo/

First choose `lisp` from the "Language categories" menu in the top/left. Then choose a style from the "Styles" menu below it. The style that was used in the screenshots is `Atom One Light`. I also thought `Vs` looked nice, but if you have a suggestion for a better style to use, please let me know!

matt-gardner: Hmm, that's too bad, I guess the colons and periods that we have as part of our identifiers messes with the syntax highlighting.  What if you try something like `logical_form.replace('fb:', '').replace('.', '-')`.  Does that look any better?

aaronsarnat: I'm confused. What should the logical form output look like? Right now all the syntax highlighter is doing is changing text colors.

image

aaronsarnat:
Thanks @matt-gardner and @schmmd.

1. Understood. I won't worry about this.

2. I hunted for a react library that unminifies code strings, but couldn't find one. I pinged the `#ui-design-and-eng` slack channel at AI2, and nobody there was aware of such a library either. Hopefully there's a python library that does this!

3. Sounds good. I won't worry about tooltips for this rev.

demo/src/components/WikiTablesComponent.js
aaronsarnat: Oh is it the inconsistency in how `fb:` is colored? I just noticed that...

aaronsarnat:
```jsx
<SyntaxHighlight>{logicalForm.replace('fb:', '').replace('.', '-')}</SyntaxHighlight>
```

image

matt-gardner: Sorry, I didn't know that javascript's `replace` doesn't replace all instances.  Is there a global replace version?  That _might_ make things better.

matt-gardner:
For (1) - "unminifying" seems significantly harder than what we would need here.  All we want to do is put a horizontal limit on the code and do some formatting.

demo/src/components/WikiTablesComponent.js
aaronsarnat: I didn't know that either! But yeah, there's an easy way to replace all instances.

Raw Data
---

```JSON
{
  "logical_form" : "((reverse fb:cell.cell.date) ((reverse fb:row.row.season) (fb:row.row.position fb:cell.1st)))"
}
```

---

Javascript
---

```JSX
<SyntaxHighlight language="lisp">
  {logicalForm.split('fb:').join('').split('.').join('-')}
</SyntaxHighlight>
```

---

Output
---

image

aaronsarnat:
Just to clarify, by "unminify" i mean only whitespace formatting. A.k.a. "beautify" or "prettify". Basically adding newlines and indents where appropriate.

Were you wanting something else? If so, might it be as simple as doing a regex over a string?

demo/src/components/WikiTablesComponent.js
aaronsarnat: @matt-gardner this output is using your character replacement suggestion, the `vs` code style, and `lisp` as explicitly-set language:

image

Is this what we want in terms of syntax highlighting, or should I start looking for alternative libraries?

aaronsarnat:
Thanks @matt-gardner! I'll merge this as soon as TeamCity gives me the green check.

PR #84, opened 2018-10-02
Adding normalization type to <HeatMap />

aaronsarnat:
@matt-gardner this PR:

- Plumbs in an optional `normalization` prop that gets passed down from model component to where opacity is calculated in `<DataGrid />`.
- Fixes some minor JS-driven CSS syntax bugs that were throwing console warnings.

---

Hopefully, this will help: search `DataGrid.jsx` and  `WikiTablesComponent.js` for instances of<br> `TODO(matt-gardner)`.
<br>

### DIFF
--- a/demo/src/components/heatmap/DataGrid.jsx
+++ b/demo/src/components/heatmap/DataGrid.jsx
@@ -16,44 +16,56 @@ const DataGrid = ({
   height,
   normalization,
 }) => {
-  var opacity;
-  if (normalization === "none") {
-    opacity = data;
-  } else if (normalization === "linear") {
+  
+  let opacity;
+
+  function normalizeLinear() {
     const flatArray = data.reduce((i, o) => [...o, ...i], []);
     const max = Math.max(...flatArray);
     const min = Math.min(...flatArray);
-    if (max == min) {
+    if (max === min) {
       opacity = data;
     } else {
       opacity = data.map((x_list) => x_list.map((x) => ((x - min) / (max - min))));
     }
-  } else if (normalization === "log-global") {
-    const exped = data.map((x_list) => x_list.map((x) => Math.exp(x)));
-    const flatArray = exped.reduce((i, o) => [...o, ...i], []);
-    const sum = flatArray.reduce((a, b) => a + b, 0)
-    opacity = exped.map((x_list) => x_list.map((x) => x / sum));
-  } else if (normalization === "log-per-row") {
-    const exped = data.map((x_list) => x_list.map((x) => Math.exp(x)));
-    opacity = exped.map((x_list) => {
-      const sum = x_list.reduce((a, b) => a + b, 0)
-      return x_list.map((x) => x / sum)
-    });
-  } else if (normalization === "log-per-row-with-zero") {
-    const exped = data.map((x_list) => x_list.map((x) => Math.exp(x)));
-    opacity = exped.map((x_list) => {
-      const sum = x_list.reduce((a, b) => a + b, 0) + Math.exp(0)
-      return x_list.map((x) => x / sum)
-    });
   }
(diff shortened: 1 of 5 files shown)


matt-gardner:
@aaronsarnat, try this out.  I had to play around with how I did the normalization a bit before I got something that I liked, but I'm happy with this.  If you think it's ok, you can remove the "NOT FOR MERGE" bit in the PR description.

aaronsarnat:
@matt-gardner thanks for making this change. 2 questions:

1. What would be the default normalization type? Linear? As-is we're breaking the other models that use `<HeatMap />`. We either want to have a catch-all default to support cases where there's no `normalization` data received, or make `normalization` a required field and explicitly set that value on all instances of `<HeatMap />` in the demo. I'd just need some guidance there before I impose any normalization types on the other models.

2. Just want to double-check, does this output look correct to you?

    image

demo/src/components/heatmap/DataGrid.jsx
```diff
@@ -14,12 +14,61 @@ const DataGrid = ({
...
+    }
+    case "linear": {
+      normalizeLinear();
+      break;
+    }
+    default: {
+      normalizeLinear();
+    }
+  }
+
```
aaronsarnat: @matt-gardner can you take a look at this change I made? The idea was that we could support explicit setting of normalization to `"linear"` while also supporting a default where no parameter is passed.

demo/src/components/heatmap/HeatMap.jsx
```diff
@@ -3,18 +3,26 @@
  * columns.  This is done with a series of divs, where the first div has the column labels (the
  * `XLabels` component) and the rest of the divs each have a row label and then colored boxes for
  * each cell, where the colors are determined by the data values (the `DataGrid` component).
+ *
+ * Supported normalization types:
+ *    "none",
+ *    "log-global",
+ *    "log-per-row",
+ *    "log-per-row-with-zero",
+ *    "linear" (default)
```
aaronsarnat: @matt-gardner I added a list of supported normalization values. Do you think we'd want to add some descriptions to these?

demo/src/components/heatmap/DataGrid.jsx
aaronsarnat: (this also fixes what was breaking MC and TE)

aaronsarnat:
Thanks @matt-gardner for all your help. I just made a couple of final changes:

1. Reverted the switch statement to if/else. React was throwing a warning about not having a `default` case anymore. Same logic, just using the structure you had originally.

2. Explicitly set linear normalization for MC and TE models. The output on those was defaulting to "none" which was causing the shading to look different on those models than what is currently on master. I didn't make any changes to WikiTables.

Just got green checks, so am going to merge now...

matt-gardner:
None is better than linear, for all models.

aaronsarnat:
Ok, I can take out the linear normalization for MC and TE. I was just noticing that linear looked like what is currently on master. None looked a little more washed out since there weren't any cells with opacity reaching 1.

Will revert my last commit and merge

aaronsarnat:
> I made the default "none" on purpose - the default of "linear" was actually wrong, and it's just better to have no normalization - the inputs to this component are generally already-normalized probability distributions, and we shouldn't be mucking with them.

Got it. Thanks for the clarification, and sorry for the confusion. I saw your comment and fixed those before I merged. Anything without an explicit normalization value should be defaulting to `"none"`, like you had it:

http://staging.allennlp.org/machine-comprehension/MzgxOTUy
http://staging.allennlp.org/textual-entailment/MzgxOTUz
http://staging.allennlp.org/wikitables-parser/MzgxOTU3

PR #100, opened 2018-10-05
Fix arguments to HeatMap in AtisComponent.js

matt-gardner:
I should have realized that the wrong import also made it so the arguments changed.  Oops again...


matt-gardner:
Just merging, as the build passes and it's really simple changes.

Issue #97, opened 2018-10-04
Add custom tooltip for displaying HeatMap cell values


aaronsarnat:
![image](https://user-images.githubusercontent.com/8367927/46628319-09327c00-caf2-11e8-9b92-8f8aac11d4cd.png)

- Same look and feel as `<Highlight />` tooltip (e.g. NER)
- Instantly shown on hover (no need to wait several seconds for browser tooltip to show up)
- Could support additional metadata beyond cell value (TBD)

---

Note:

- For initial release, we don't want to show normalization type in the tooltip. Only the value.
- May want to round floating points to nearest hundredth or thousandth (TBD)
- [Crosshair UI](https://github.com/allenai/allennlp-demo/issues/96) is being tracked in a separate issue

aaronsarnat:
This task is based on the recent aspirational design for the new Attention Visualization:

[![image](https://user-images.githubusercontent.com/8367927/46327384-3d6be100-c5b6-11e8-8b7b-d50ce3ee2145.png)](https://drive.google.com/file/d/1WoCAjcUfT0SvZY2aNJuhoFeXGgnvo8ZR/view?usp=sharing)

aaronsarnat:
@matt-gardner @schmmd Note that despite the mockup, I'm not going to add the normalization type to the tooltip (as discussed on a separate thread with Matt).

However, I wanted to ask about rounding of floating point numbers. Michael called out the possibility of rounding to the nearest hundredth (or thousandth?). I wanted to get a consensus on this before I spend any time on rounding. Do we want to:

1. Round to the nearest _____th
2. Not round at all
3. Not round for this PR, but revisit later in a separate issue

Thanks guys! This and crosshairs are my focus for the day, which I will make sure I get shipped before I pivot to PRIOR, but if we don't have consensus before I roll off of AllenNLP then I will default to option 3.

On a related note, having a custom tooltip will allow us to include whatever metadata we want. Something to think about for future iterations.

matt-gardner:
I'd probably default to no rounding.  These are often probability distributions, and some have very small values.  Rounding 0.0001 to 0 would be very bad, as those are very different values in a probability distribution (that is, seeing a 0 in a probability distribution typically means something is broken in my model, while 0.0001 is just fine).

Issue #96, opened 2018-10-04
Show crosshairs when hovering HeatMap


aaronsarnat:
![attention-crosshairs](https://user-images.githubusercontent.com/8367927/46628075-4fd3a680-caf1-11e8-8e3a-fad1957d70f3.gif)

---

When user hovers a cell in the HeatMap, they should have an indication of which row and column they are referencing.

Despite the aspirational design above, this issue does not deal with:
- [Tooltips](https://github.com/allenai/allennlp-demo/issues/97)
- [Angled Column Headers](https://github.com/allenai/allennlp-demo/issues/95)

aaronsarnat:
This task is based on the recent aspirational design for the new Attention Visualization:

[![image](https://user-images.githubusercontent.com/8367927/46327384-3d6be100-c5b6-11e8-8b7b-d50ce3ee2145.png)](https://drive.google.com/file/d/1WoCAjcUfT0SvZY2aNJuhoFeXGgnvo8ZR/view?usp=sharing)

PR #105, opened 2018-10-10
Adding crosshairs and tooltip to HeatMap

aaronsarnat:
heatmap

---

**This PR:**
- Fixes [Show crosshairs when hovering HeatMap (#96)](https://github.com/allenai/allennlp-demo/issues/96)
- Fixes [Add custom tooltip for displaying HeatMap cell values (#97)](https://github.com/allenai/allennlp-demo/issues/97)
- Introduces a new 3rd-party component, [`react-tooltip`](https://www.npmjs.com/package/react-tooltip) (was necessary to solve container overflow and window edge masking issues)
- Adds row and column metadata to cell value tooltip
- Removes grid lines
- Changes how opacity is applied (now as an alpha on an RGB background color value)
- Compacts data grid padding by a couple of pixels
- Adds rudimentary support for additional heatmap colors

### DIFF
--- a/demo/package-lock.json
+++ b/demo/package-lock.json
@@ -13877,6 +13877,16 @@
         "refractor": "^2.4.1"
       }
     },
+    "react-tooltip": {
+      "version": "3.8.4",
+      "resolved": "https://registry.npmjs.org/react-tooltip/-/react-tooltip-3.8.4.tgz",
+      "integrity": "sha512-i9rP5eihRNXFqYtyw58fhy+oT7OAEox/WC7XJOSED57s31nLU+uOJai5TgADbwNzJ7xamPkijYF36IfrUGJgcQ==",
+      "requires": {
+        "classnames": "^2.2.5",
+        "prop-types": "^15.6.0",
+        "sanitize-html-react": "^1.13.0"
+      }
+    },
     "read-all-stream": {
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/read-all-stream/-/read-all-stream-3.1.0.tgz",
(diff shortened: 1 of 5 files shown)


aaronsarnat:
@matt-gardner assigning to you for review. @schmmd FYI.

I'll address any pressing PR feedback before merging this, but otherwise I must now pivot to PRIOR for the foreseeable future... This attention visualization stuff is fun though, and all the additional features relating to the aspirational design now have [issues](https://github.com/allenai/allennlp-demo/issues?q=is%3Aopen+is%3Aissue+label%3A%22Attention+Viz%22) associated with them. Here's hoping I get to pick up where we left off sooner than later...

matt-gardner:
I tried this out locally, and it looks great!  The one issue I had was that it's painfully slow with the ATIS parser demo.  The others were just fine.  I tried to dig around to see if there was a way to speed it up, and I'm pretty sure the problem is with the `Collapsible` component.  Because of how it works, we're rendering _way_ more stuff than we actually need to render at a time, even when it's hidden, and the ATIS model has a ton of actions, so this is particularly problematic.  It would be better to have that component take a function that gets rendered on expand, or something.

But, the plan is to switch to a drawer, anyway, and remove the use of the `Collapsible` component, so I didn't spend too much time trying to make it better.  We should make a note, somewhere, though, that for the drawer, you want to render the active component in the drawer on the fly, not have some things rendered but invisible.  The list of items in the drawer could be huge, and we don't want to render them until they're needed.

aaronsarnat:
Thanks @matt-gardner ! Yeah, I noticed some sluggishness as well. As-is we're rendering a lot of markup to create these heatmaps. I created an [issue](https://github.com/allenai/allennlp-demo/issues/106) for optimizing some of the html/css structure which may help a bit, but you're right in that the drawer upgrade will allow us to render a single table at a time, on the fly, and on demand. Hard to predict how dramatically that will improve interaction performance, but it certainly can't hurt!

Going to go ahead and merge this. Looking forward to beefing this up when I'm able to come back to AllenNLP.

PR #108, opened 2018-10-26
Update Event2Mind to account for multi-GPU fixes.

brendan-ai2:
- The model (in terms of code) as used by the predictor is apparently incompatible after the multi-GPU fixes with the serialized model stored on S3.


brendan-ai2:
Note to self: Bump the numbers on the website.

brendan-ai2:
Thanks! I used the `Event2Mind` model to validate that the multi-GPU code was working as expected. Resulted in a small change in how the model was written, nothing at the architecture level.  See https://github.com/allenai/allennlp/commit/2a8bd63829ed6e84ecd011e9dd8d134db41d93dd#diff-8f018044da0180b1f5c2aa4a0368d1b8. In short, we need to use `ModuleDict`, `ModuleList`, etc. instead of using `add_module` unless we're very careful with how we subsequently access the submodules. This is due to the manner in which the model is copied and modified to run on the different GPUs. (Half-shallow, half-deep)

PR #113, opened 2018-11-15
Log requester IPs to the database

schmmd:


schmmd:
@joelgrus yes ;-)

And I've written data to it locally to test my changes.

PR #114, opened 2018-11-19
Use ProxyFix to properly set the requester ip.

schmmd:


schmmd:
I think it's part of Flask.  Let me double-check the build would catch a failure before merging (my environment might be messy).

PR #121, opened 2018-12-18
Don't cache models in demo Dockerfile

schmmd:
Our images are getting huge (6 GB).  This removes most of the data in our images and allows users to use their local cache (by mounting in the appropriate directory) when running the demo locally.


brendan-ai2:
Seems like a nice change, @schmmd. Could you elaborate on precisely what modifications to make to use the local cache?

schmmd:
You would use `--mount` to bring in your local `~/.allennlp/` into the Docker image.

PR #128, opened 2018-12-20
add quarel demo

joelgrus:
for the most part this is a straight port of Oyvind's existing demo into the new refactored world.

there was an `appendDefault` piece in the original that I didn't entirely understand, but I think it was trying to be more general and that the correponding behavior here is correct.

also, the trained quarel model referenced a trained tagger model that used a deprecated parameter, so I had to update the tagger mode and then the quarel model. this code reflects that too.

-- 

there is a section at the end where the fine-grained NER model is commented out, it needs to be similarly updated, but downloading and uploading these models is a pain on home network, so I'll fix that tomorrow, and I won't merge until that's done. in the meantime, that has nothing to do with the quarel demo itself, so that can still be reviewed.

### DIFF
--- a/demo/src/components/demos/QuarelZero.js
+++ b/demo/src/components/demos/QuarelZero.js
@@ -111,6 +111,18 @@ const qrspecDefault = "[friction, -speed, -smoothness, -distance, +heat]\n[speed
 
 const entitycuesDefault = "friction: resistance, traction\nspeed: velocity, pace, fast, slow, faster, slower, slowly, quickly, rapidly\ndistance: length, way, far, near, further, longer, shorter, long, short, farther, furthest\nheat: temperature, warmth, smoke, hot, hotter, cold, colder\nsmoothness: slickness, roughness, rough, smooth, rougher, smoother, bumpy, slicker\nacceleration: \namountSweat: sweat, sweaty\napparentSize: size, large, small, larger, smaller\nbreakability: brittleness, brittle, break, solid\nbrightness: bright, shiny, faint\nexerciseIntensity: excercise, run, walk\nflexibility: flexible, stiff, rigid\ngravity: \nloudness: loud, faint, louder, fainter\nmass: weight, heavy, light, heavier, lighter, massive\nstrength: power, strong, weak, stronger, weaker\nthickness: thick, thin, thicker, thinner, skinny\ntime: long, short\nweight: mass, heavy, light, heavier, lighter"
 
+const appendDefault = (example, fieldName, defaults) => {
+    if (example[fieldName] && example[fieldName].slice(-1) == '*') {
+        // Ends with *, so don't append anything and get rid of the '*'
+        example[fieldName] = example[fieldName].slice(0, -1)
+    } else if (example[fieldName]) {
+        example[fieldName] += "\n" + defaults
+    } else {
+        example[fieldName] = defaults
+    }
+}
+
+
 const examples = [
     {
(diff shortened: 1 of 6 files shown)


demo/src/components/demos/QuarelZero.js
```diff
@@ -0,0 +1,179 @@
...
+    {name: "question", label: "Question", type: "TEXT_AREA",
+     placeholder: `E.g. "A hockey puck slides a lot longer on a frozen lake then on a slushy lake. This means the surface of the _____ is smoother (A) frozen lake (B) slushy lake"`},
+     {name: "qrspec", label: "Qualitative Relations", type: "TEXT_AREA", optional: true,
+     placeholder: `[friction, -speed, +heat]\n[weight, acceleration]`},
+     {name: "entitycues", label: "Attribute Cues", type: "TEXT_AREA", optional: true,
+     placeholder: `friction: resistance\nspeed: velocity, fast, slow`}
+]
+
+const ActionInfo = ({ action, question_tokens }) => {
+    const question_attention = action['question_attention'].map(x => [x])
```
schmmd: What is this doing?

matt-gardner:
I'm pretty sure the `appendDefault` bit is important - it's what puts the rest of the default qualitative relations into the model.  If you look [here](http://aristo-demos-adhoc.dev.ai2/quarel-q-parsing-zero) and select the first example, the example in the code only has the sugar/diabetes relation, but when you select it you get the whole list.  It's cumbersome to add the whole list every time.  An easy way around this is to just have those qualitative relations hard-coded in every example (possibly all saved in a class variable to make that easier).  At least, I'm pretty sure that's what that's doing.  @OyvindTafjord might have more to say, though.

demo/src/components/demos/QuarelZero.js
```diff
@@ -0,0 +1,189 @@
...
+import { API_ROOT } from '../../api-config';
+
+const title = "Qualitative Relations Story Question Answering"
+
+const description = (
+<span>
+  Answer story questions about qualitative relations (QuaRel dataset) while adding new relations
+  without retraining. This uses the QuaSP+Zero semantic parser described in <i>QuaRel: A Dataset and Models for
+  Answering Questions about Qualitative Relationships</i> (in submission). The first few examples use
+  new relations, the rest are from the validation set.
```
OyvindTafjord: Could we change "in submission" to "AAAI 2019", linked to https://arxiv.org/abs/1811.08048  (or http://data.allenai.org/quarel/)?

demo/src/components/demos/QuarelZero.js
OyvindTafjord: This piece is basically just copied from https://github.com/allenai/allennlp-demo/blob/master/demo/src/components/demos/WikiTables.js#L40

OyvindTafjord:
One small comment about the paper reference in the intro. Also note that the tagger model which the parser depends on will not be cached inside the docker image by the `cache_models.py` script which might be annoying for deployment purposes?

joelgrus:
ok, I think this should be good, I made the following changes

* fixed the "appendDefault" behavior, I think it's doing the right thing now
* changed the description as requested
* fixed the fine-grained-ner model and added it back

it doesn't look like the dockerfile is actually running cache_models right now (I think @schmmd took that out because our docker images are too big), but when that script runs it not only downloads but hydrates each model, which means it should also download transitively required models.

PR #136, opened 2019-02-11
Configure sentry in both Python and Javascript.

schmmd:
https://sentry.io/ is an error aggregation software.  The vision team has used it to good effect with Iconary and I wanted to evaluate it once it has aggregated errors from general AllenNLP usage.  The hope is it will help us hone in on Python and Javascript exceptions that our users experience but may be unknown to us.


brendan-ai2:
Could you please update the description with a link to some "Sentry" docs? We need to leave some breadcrumbs for people.

PR #137, opened 2019-02-11
Change installation instructions for clearer options on how to install AllenNLP

schmmd:


schmmd:
@joelgrus does this look OK?  I get confused by the current setup every time I use it.

Issue #37, opened 2018-08-28
Memory leak in the demo


schmmd:
The AllenNLP demo has a memory leak that causes pods to fail in Kubernetes.  Recently we added the dependency parser and pods were restarting every few minutes, so I doubled the memory available on each machine.  Regardless, I'm still seeing a few restarts overnight.

```
$ kubectl get pods
NAME                                     READY     STATUS    RESTARTS   AGE
allennlp-demo-prod-597b67546-mglpq       2/2       Running   3          18h
allennlp-demo-prod-597b67546-tfdnj       2/2       Running   2          18h
allennlp-demo-prod-597b67546-xzvjh       2/2       Running   4          18h
```

The user isn't seeing any impact because the pods restart when they crash and we have three replicas, but we are presently addressing the situation by spending more money (on replicas) and our ability to add more models is limited.

```
$ kubectl describe pod allennlp-demo-prod-597b67546-mglpq
Name:           allennlp-demo-prod-597b67546-mglpq
Namespace:      default
Node:           gke-allennlp-demo-cluster-new-pool-bdd08a8c-z0c5/10.9.254.14
Start Time:     Mon, 27 Aug 2018 13:46:54 -0700
Labels:         app=allennlp-demo-prod
                pod-template-hash=153623102
Annotations:    kubernetes.io/limit-ranger=LimitRanger plugin set: cpu request for container cloudsql-proxy
Status:         Running
IP:             10.4.14.10
Controlled By:  ReplicaSet/allennlp-demo-prod-597b67546
Containers:
  allennlp:
    Container ID:   docker://61c5625be2673d770391807f03e0060eb3948076a216fe75a8f1f616f048de2c
    Image:          allennlp/demo:e68ab81a7a52a1ee13831f90358f2698e9c90209
    Image ID:       docker-pullable://allennlp/demo@sha256:c16c141bba76bd4b73015ffb22739210e2ebc7c8427a70d22a60884579194761
    Port:           <none>
    State:          Running
      Started:      Tue, 28 Aug 2018 08:04:03 -0700
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Tue, 28 Aug 2018 05:06:44 -0700
      Finished:     Tue, 28 Aug 2018 08:04:02 -0700
    Ready:          True
    Restart Count:  3
    Limits:
      cpu:     1
      memory:  3000Mi
    Requests:
      cpu:      1
      memory:   3000Mi
[...]

PR #139, opened 2019-02-12
Update DB with request before predicting.

brendan-ai2:
- This will allow us to track down "queries of death".
- In service of https://github.com/allenai/allennlp-demo/issues/37.
- Also includes a script and instructions for running a local development DB.


brendan-ai2:
Do we have a copy of the SQL needed to start up a local copy of the DB for development? I wrote a unit test for this change, but I should probably validate against an actual postgres before pushing to prod. Though maybe staging is sufficient?

schmmd:
@brendan-ai2 staging uses the prod database.  I don't have such a query.

brendan-ai2:
I'll try to extract it from the DB with `pg_dump`.

schmmd:
I.e. we could extend https://github.com/allenai/allennlp-demo/blob/d1b5ce0651074b2ca5cc4f9fa4ec8d050b199eb5/app.py#L259 by logging the requests as they come in (say, here https://github.com/allenai/allennlp-demo/blob/d1b5ce0651074b2ca5cc4f9fa4ec8d050b199eb5/app.py#L178).

schmmd:
@brendan-ai2 that should work.  There are only 1 or 2 tables as I recall.

brendan-ai2:
Thanks for the review! Yeah, I had the same thought about logging, but initially decided against as it makes the logs slightly harder to read. Super minor issue though.

I managed to get the schema with `pg_dump`. I'll clean that up and get it working for creating a local DB tomorrow when I'm back in the office.

brendan-ai2:
Updated with a script + instructions on setting up a DB locally for development. Sure enough, this caught a typo in my SQL. Going to go ahead and merge now that that's fixed!

PR #140, opened 2019-02-12
initial push of style updates

jonborchardt:
1- made all demos horizontal vs vertical
2- squished content a bit when screen size is constrained (tracking to look ok on 1280x768)
3- made model description collapsable
4- using accordion instead of collapse for better look and utility of heatmaps
5- added ability to filter heatmaps to the words that get higher scores

you should pull down and build to compare to the current.
(is there a way to deploy this branch?)

image

image


demo/ADDING_A_DEMO.md
```diff
@@ -46,6 +46,16 @@ const description = (
...
 ```
 
+add the descriptionEllipsed as just JSX (or some other element):
+
+```js
+const descriptionEllipsed = (
+    <span>
+        Machine Comprehensi… // Note: ending in a …
+    </span>
+)
```
codeviking: Are the descriptions already JSX, or are they text only?

If they're text only, what if we just wrote a function that "ellipsized" the text for them by truncating (without splitting words) to a given character limit (as close to it w/o breaking words) and appended the ellipsis.

The code for this is already in S2 https://github.com/allenai/scholar/blob/master/online/webapp/app/js/util.js#L160, which you could use if that'll do the trick.

demo/src/components/HeatMap.js
```diff
@@ -45,57 +52,103 @@ export default class HeatMap extends React.Component {
...
       }
-    } else if (normalization === "none") {
-      opacity = data;
+    } else {
+      opacity = newProps.data;
+    }
+
+    const flatArray = opacity.reduce((i, o) => [...o, ...i], []);
+    this.setState({opacity: opacity, minFilterOpacity: Math.min(...flatArray), maxFilterOpacity: Math.max(...flatArray)},
+      () => this.setShowRowsAndColumns(this.state));
```
codeviking: Woah, I'm not even going to pretend to follow this, as it's a lot.  That said as I understand it you migrated the effect of this method into `state`, and possibly extended things a bit, which seems like a step in the right direction.

demo/src/components/demos/DependencyParser.js
```diff
@@ -19,6 +19,12 @@ const description = (
   </span>
 )
 
+const descriptionEllipsed = (
+<span>
+    A dependency parser analyzes the grammatical structure of a sentence, establishing relationships between "hea…
+</span>
```
codeviking: I wonder if we should try to include the full word, rather than truncating mid word (that's what the referenced ellipsize impl) does.

demo/src/components/demos/Nlvr.js
```diff
@@ -26,6 +31,12 @@ const description = (
   </span>
 );
 
+const descriptionEllipsed = (
+  <span>
+    Semantic parsing maps natural language to machine language.  This page demonstrates a semantic parsing mod…
+  </span>
+)
```
codeviking: Weird that Github's parser is having issues here, any idea what's causing that? Is there possibly an unclosed tag somewhere in the mix?

schmmd:
@jonborchardt we don't have the functionality to deploy branches easily.  I mean--I can do it without too much trouble but it would take some custom work.  I feel great about your changes!

demo/src/css/model.css
```diff
@@ -24,6 +24,13 @@
   color: rgba(35, 35, 35, .75);
 }
 
+.model__content p {
+  margin: 0.5em 0 0 0;
+}
+.model__content p span {
+  width: 1000px;
+}
```
codeviking: Is the demo responsive? If so this will likely cause issues for mobile, unless it's overridden elsewhere.

I tend to avoid `width` where possible, that said it might be necessary here.

jonborchardt:
@schmmd thanks, these changes are easy to alter/revert if we end up wanting something different later.

demo/ADDING_A_DEMO.md
jonborchardt: they are jsx... not text.... :-(

demo/src/components/HeatMap.js
jonborchardt: correct, most of this is migration to state

demo/src/components/demos/DependencyParser.js
jonborchardt: ill, pull back to the last full word (by hand)

demo/src/css/model.css
jonborchardt: oops, i think this is a temp thing i wanted to delete.... thanks for spotting it

yes, it is responsive...

demo/src/components/demos/Nlvr.js
jonborchardt: ill check, but i dont think so

demo/src/css/model.css
jonborchardt: removed

demo/src/components/demos/Nlvr.js
jonborchardt: looks fine in code

demo/src/components/demos/DependencyParser.js
jonborchardt: done

joelgrus:
I personally find the new design pretty cramped, compare:

spacious:
screenshot from 2019-01-17 09-45-10

cramped:
screenshot from 2019-01-17 09-45-19

but I guess that was one of the design goals, so 🤷

joelgrus:
I personally find the new design pretty cramped, compare:

spacious:

screenshot from 2019-02-12 14-36-57


cramped:

screenshot from 2019-02-12 14-37-08

but I guess that was one of the goals, so 🤷

(sorry about the previous comment, I don't know to paste screenshots apparently)

joelgrus:
I do like the heatmap changes though

schmmd:
I do think the input is more cramped, but the benefit is that we have more space for the visualization and we can still see the input and output together.  @jonborchardt is there an easy way to give us more breathing room without sacrificing the goal of seeing both the input and the output (on most screens)?  If so--could we try that out?

If not, please merge, we'll try it out, and revise forward.

jonborchardt:
The proposed design allows for a min of 31% of the screen to be answer space (before scroll, on 768 screen, on most complex model )

image

The existing design does not look well in this size:

image

We can spread out the top and make the answer space less than 31%... if we want to.

jonborchardt:
I am going to add a few more pixels above the labels....

PR #143, opened 2019-02-15
make menu grouped and add styled components

jonborchardt:
current:
image

proposal:
image

with something closed
image

### DIFF
--- a/demo/config-overrides.js
+++ b/demo/config-overrides.js
@@ -1,0 +1,7 @@
+const rewireStyledComponents = require('react-app-rewire-styled-components');
+
+module.exports = function override(config, env) {
+  // do stuff with the webpack config...
+  config = rewireStyledComponents(config, env);
+  return config;
+}
--- a/demo/config-overrides.js
+++ b/demo/config-overrides.js
@@ -1,3 +1,4 @@
+// see: react-app-rewired
 const rewireStyledComponents = require('react-app-rewire-styled-components');
 
 module.exports = function override(config, env) {
--- a/demo/config-overrides.js
+++ b/demo/config-overrides.js
@@ -1,4 +1,10 @@
-// see: react-app-rewired
+// Using react-app-rewired to tweak the create-react-app webpack config(s) without using 'eject' and
+// without creating a fork of the react-scripts.
+//
+// This is needed for styled-components (react-app-rewire-styled-components)
+// but is generally useful if we want to add other configurations in the future.
+// see: https://github.com/timarney/react-app-rewired
+
 const rewireStyledComponents = require('react-app-rewire-styled-components');
 
 module.exports = function override(config, env) {
--- a/demo/package.json
+++ b/demo/package.json
@@ -8,6 +8,7 @@
     "react": "15.6.1",
     "react-accessible-accordion": "2.4.5",
     "react-app-rewire-styled-components": "3.0.2",
+    "react-app-rewired": "2.1.0",
     "react-dom": "15.6.1",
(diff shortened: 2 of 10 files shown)


matt-gardner:
You could imagine an "answer a question" demo that does not have a passage and does some retrieval first.  We don't currently have one of these, but there used to be one up at documentqa.allenai.org, and we might want to put one there again eventually.  Also, you could have visual question answering there at some point.  So, "answer a question" does make sense as a top-level thing, with "Reading Comprehension" as our only current subcategory, but with more potentially coming later.  Please rename it from "machine comprehension" to "reading comprehension", though.

jonborchardt:
i forgot to push some small tweaks to the code, ill do on monday... so dont push this yet...

codeviking:
Looks good, could we add a bit of padding to the category items.  They look a little squished to me! I'll take a look at the code now.

demo/config-overrides.js
```diff
@@ -0,0 +1,7 @@
+const rewireStyledComponents = require('react-app-rewire-styled-components');
+
+module.exports = function override(config, env) {
+  // do stuff with the webpack config...
+  config = rewireStyledComponents(config, env);
+  return config;
+}
```
codeviking: Any chance we can add a more detailed explanation here? I'm not sure what this does, and an explanation would be helpful, even if it's something like `// Not totally certain what this does, but it was required to use styled components.`...

demo/package.json
```diff
@@ -3,21 +3,26 @@
...
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test --env=jsdom",
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
+    "test": "react-app-rewired test --env=jsdom",
     "eject": "react-scripts eject"
+  },
+  "devDependencies": {
+    "react-app-rewired": "2.1.0"
```
codeviking: You might just keep this in the regular deps for simplicity.  The dev-dep vs dep distinction really only matters for modules that are published to npm (or if a build passes the flags that omit dev-deps).

demo/src/components/Menu.js
```diff
@@ -9,24 +16,34 @@ import '../css/Menu.css';
...
-                <span>{label}</span>
-              </Link>
-            </span>
-          </li>
-        )
-      }
+      const { selectedModel, expandedModelGroupIndexes, clearData, onExpandModelGroup, className } = this.props;
+
+      const ModelGroup = ({modelGroup, expanded, uuid}) => (
+        <MenuAccordionItem className={`accordion__item ${expanded?'expanded':''}`} expanded={expanded} uuid={uuid}>
```
codeviking: Are we using css to hide / show them? If we are, is there a benefit to doing so instead of just toggling whether the DOM is rendered or not?

I generally opt for conditional rendering, as I feel like it's both easier to understand (one doesn't have to be aware of any css definition) and it keeps the DOM smaller (which is good for performance, though this only matters for really big apps).

demo/config-overrides.js
jonborchardt: done

demo/package.json
jonborchardt: done

demo/src/components/Menu.js
jonborchardt: the accordion code shows the section in a collapsed form... so we dont just want to toggle them out of the dom. (i agree with you in the general case)

codeviking: ah! i missed that, sounds good!

PR #168, opened 2019-02-21
 Add maximum request lengths to the demo.

brendan-ai2:
- It seems like large requests are at least partially responsible for OOMs.
- See https://github.com/allenai/allennlp-demo/issues/37.
- Unrelated test fix.


brendan-ai2:
Thanks, Joel!

PR #169, opened 2019-02-22
added additional visualizations for matts new model

jonborchardt:
also added radiobuttons for model selection and other general ui updates
(this includes mock code that i will remove)

model selection is radio button and under input
image

added tooltips to say what the models are
image

viz when no explanation
image

viz when passage
image

viz when question
image

viz when counting
image

viz when arithmetic
image


demo/src/components/demos/NamedEntityRecognition.js
```diff
@@ -46,16 +46,28 @@ const descriptionEllipsed = (
...
 
-const nerModels = ["ner", "fine-grained-ner"]
-const nerEndpoints = {
-    "ner": "named-entity-recognition",
-    "fine-grained-ner": "fine-grained-named-entity-recognition"
-  };
+const taskModels = [
+  {
+    value: "ner",
+    desc: "A coarse-grained named entity recognition model refers to a NER model that can recognize<br/>\
```
jonborchardt: PR: are these descriptions good?

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -10,78 +10,1059 @@ import {
...
+  <span>
+    Reading comprehension is the task of answering questions about a passage of text to show that
+    the system…
+  </span>
+)
+
+const taskModels = [
+  {
+    value: "BiDAF",
+    desc: "Reimplementation of BiDAF (Seo et al, 2017), or Bi-Directional Attention Flow,<br/>\
```
jonborchardt: PR: are these descriptions ok?

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -10,78 +10,294 @@ import {
...
+  "BiDAF": "machine-comprehension", // TODO: we should rename tha back-end model to reading-comprehension
+  "Augmented QANet": "augmented-qanet-reading-comprehension"
+};
 
 const fields = [
   {name: "passage", label: "Passage", type: "TEXT_AREA",
    placeholder: `E.g. "Saturn is the sixth planet from the Sun and the second-largest in the Solar System, after Jupiter. It is a gas giant with an average radius about nine times that of Earth. Although it has only one-eighth the average density of Earth, with its larger volume Saturn is just over 95 times more massive. Saturn is named after the Roman god of agriculture; its astronomical symbol represents the god's sickle"`},
   {name: "question", label: "Question", type: "TEXT_INPUT",
    placeholder: `E.g. "What does Saturn’s astronomical symbol represent"`}
+  // {name: "model", label: "Model", type: "RADIO", options: taskModels, optional: true} // TODO: add when matts model is ready
```
jonborchardt: i think this is the only todo, for the FE once matts model is avail on the backend

demo/src/components/DemoInput.js
```diff
@@ -1,15 +1,19 @@
...
-    } else {
-        return text.substring(0, maxLen) + "..."
+    if(text) {
+        if (text.length <= maxLen) {
+            return text
+        } else {
+            return text.substring(0, maxLen) + "..."
+        }
     }
+    return "";
```
codeviking: This truncation function will split on spaces and/or mid word, which isn't ideal.

I think I linked this before, but this implementation will solve those issues, and should be copy paste-able:
https://github.com/allenai/scholar/blob/master/online/webapp/app/js/util.js#L160

demo/src/components/DemoInput.js
```diff
@@ -110,18 +121,36 @@ class DemoInput extends React.Component {
                 case "SELECT":
                     input = (
                         // If we have no value for this select, use the first option.
-                        <select value={this.state[field.name] || field.options[0]}
+                        <select value={this.state[field.name] || field.options[0].value}
```
codeviking: Can `options` ever be empty? In that case this would throw an error. Not sure if you need to handle that or not.

codeviking: Also, another thought. You could simplify this by just setting state in the constructor:

```
this.state = {};
this.state[field.name] = field.options.length > 0 ? field.options[0].value : null;
```

demo/src/components/DemoInput.js
```diff
@@ -110,18 +121,36 @@ class DemoInput extends React.Component {
...
                                 onChange={this.handleInputChange(field.name)}
                                 disabled={outputState === "working"}>
                             {
-                                field.options.map((value, idx) => (
-                                    <option key={idx} value={value}>{value}</option>
+                                field.options.map((opt, idx) => (
+                                    <option key={idx} value={opt.value}>{opt.value}</option>
```
codeviking: I believe it's best to avoid using the index alone as the key:
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

Could you just use value here? I imagine it'll never be duplicated in a select?

```
key={opt.value}
```

Looks like it's not a change you introduced, feel free to ignore.

demo/src/components/DemoInput.js
```diff
@@ -110,18 +121,36 @@ class DemoInput extends React.Component {
...
                                 ))
                             }
                         </select>
                     )
                     break
 
+                case "RADIO":
```
codeviking: I'm not sure what the larger switch / case structure is, but it looks like something I don't want to know more about.

Do you think it makes sense to maintain the generic `DemoInput` class? Should we just use composition in the actual components to add interface elements (i.e. just add a `RadioGroup` in the component itself).

I lean towards the composition route, but you'll have a better idea if the `DemoInput` abstraction is actually providing value.

demo/src/components/DemoInput.js
```diff
@@ -110,18 +121,36 @@ class DemoInput extends React.Component {
...
+                    input = (
+                        // If we have no value for this select, use the first option.
+                        <RadioGroup
+                            name={inputId}
+                            selectedValue={this.state[field.name] || field.options[0].value}
```
codeviking: As with above, this could be initialized in the constructor so it's simpler here.  A trade-off either way, so not something I care deeploy about.

demo/src/components/DemoInput.js
```diff
@@ -64,6 +68,13 @@ class DemoInput extends React.Component {
             this.setState(stateUpdate)
         }
 
+        // for radio input, the event is simply the value
+        this.handleRadioInputChange = name => e => {
+            let stateUpdate = {}
+            stateUpdate[name] = e;
```
codeviking: Isn't `e` the event?  Shouldn't we store only the `value`? The event object has references to DOM elements, so I'd worry about a potential memory leak here (we're storing an object in state that references DOM elements that could be thrown away, so we're building a link that might prevent the DOM from being gc'd when it's no longer present).

The memory leak isn't likely a big deal because the lifetime of the parent component isn't long -- but still feels a bit odd. I'd be tempted to either store the value only, or store a smaller object that only has what we need, for clarity.

```
stateUpdate[name ] = { value: e.value }
```

demo/src/components/Shared.js
```diff
@@ -0,0 +1,49 @@
...
+import ReactTooltip from 'react-tooltip';
+
+export const RadioGroup = styled(ReactRadioGroup.RadioGroup)`
+  && {
+    margin-top: 0;
+    label,
+    label input {
+      margin-top: 0;
+      cursor: pointer;
+      width: fit-content;
```
codeviking: Hadn't seen `fit-content` before. Interesting.

Looks like Firefox has some gotchas:
https://caniuse.com/#search=fit-content

Might want to make sure none of those cause problems.

demo/src/components/Shared.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    input[type='radio']:after,
+    input[type='radio']:checked:after {
+        width: 11px;
+        height: 11px;
```
codeviking: We may want to use `ems` here for sizing, so that if the user increases their default font-fize the input changes sizes.

I imagine the existing app has all sorts of accessibility issues, so it might be a task we save for later.

demo/src/components/Shared.js
```diff
@@ -0,0 +1,49 @@
...
+        border-radius: 11px;
+        top: -1px;
+        left: -2px;
+        position: relative;
```
codeviking: I usually find that good ole' flexbox prevents me from needing this hack anymore. Could we use flex on the parent and `align-items: center` to avoid this?

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -10,78 +10,294 @@ import {
...
+    let ret = numbers
+      .filter(n => n.sign !== 0)
+      .map(n => `${n.sign > 0 ? "+" : "-"} ${n.value}`)
+      .join(" ");
+    while(ret.charAt(0) === "+" || ret.charAt(0) === " ") {
+      ret = ret.substr(1);
+    }
+    return <span>{ret}</span>;
+  }
+  return null;
```
codeviking: I wonder of `Intl.NumberFormat` could help with this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

That way we'd be friendly to international numeric representations too -- though I'm not sure how important that is.

demo/src/components/demos/NamedEntityRecognition.js
schmmd: We should probably rename "ner" to something more meaningful.  @matt-gardner or @DeNeutoy would have a better suggestion than what I would come up (maybe `elmo-ner`.  This is the NER model at https://s3-us-west-2.amazonaws.com/allennlp/models/ner-model-2018.12.18.tar.gz which corresponds to the following text (I believe):

> "The named entity recognition model identifies named entities (people, locations, organizations, and miscellaneous) in the input text. This model is a reimplementation of the state-of-the-art NER model described in Deep contextualized word representations, and uses a biLSTM with CRF layer and ELMo embeddings. It was trained on the CoNLL-2003 NER dataset, and has test set F1 of 92.5 for a single run, compared to the reported 92.22 +/- 0.10 F1 across five seeds in the reference paper."

A possible (short) description:

> This named entity model is based on the NER model described in "Deep contextualized word representations" by Peters, et. al.  It uses a biLSTM with CRF layer and ELMo embeddings. It was trained on the CoNLL-2003 NER dataset

I don't know what definition is appropriate for fine-grained-ner.

demo/src/components/demos/ReadingComprehension.js
schmmd: Works for me.

demo/src/components/demos/NamedEntityRecognition.js
DeNeutoy: "The fine grained named entity recognition model identifies a broad range of 16 semantic types in the input text. This model is a reimplementation of [Lample (2016)](https://arxiv.org/abs/1603.01360)  and uses a biLSTM with a CRF layer, character embeddings and ELMo embeddings. It was trained on the Ontonotes 5.0 dataset, and has dev set F1 of 88.2."

demo/src/components/DemoInput.js
jonborchardt: replaced the code with the new function (we should move to Varnish.util later)

demo/src/components/DemoInput.js
jonborchardt: protected the value, but moving to state is a no go at this time

demo/src/components/DemoInput.js
jonborchardt: agreed, fixed

demo/src/components/DemoInput.js
jonborchardt: agreed, but will do later

demo/src/components/DemoInput.js
jonborchardt: agreed, but later

demo/src/components/DemoInput.js
jonborchardt: the code that returns this 'event' is just returning the value, i reworded and updated variable names to make clearer

demo/src/components/Shared.js
jonborchardt: the issues look ok

demo/src/components/Shared.js
jonborchardt: fixed in this place... but, yes, need a larger fix (as we move to styled components)

demo/src/components/Shared.js
jonborchardt: good idea, fixed

demo/src/components/demos/ReadingComprehension.js
jonborchardt: im not sure how this would help
the current impl allows the backend to pass the value/precision they want
e.g. five - 6 = -1 OR 1.111 - 0.111 = 1
the ui makes no change

demo/src/components/demos/NamedEntityRecognition.js
jonborchardt: updated, please recheck

demo/src/components/demos/NamedEntityRecognition.js
```diff
@@ -53,7 +53,7 @@ const taskModels = [
   },
   {
     name: "fine-grained-ner",
-    desc: "A fine-grained named entity recognition model refers to a NER model that can recognize<br/>and classify a large number of entity categories (e.g., hundreds of NE categories)"
+    desc: "This Model identifies a broad range of 16 semantic types in the input text.<br/>This model is a reimplementation of Lample (2016) and uses a biLSTM<br/>with a CRF layer, character embeddings and ELMo embeddings. It was<br/>trained on the Ontonotes 5.0 dataset, and has dev set F1 of 88.2."
```
DeNeutoy: 👍

PR #170, opened 2019-02-22
interactive beam search -> wikitables parser

joelgrus:
* add `BeamSearch` component
* introduce notion of "input-outputs" -- outputs in the sense that they depend on the results of a call to the model API; inputs in the sense that they can be used to launch a new call to the model API
* accordingly, introduce the notion of "dependent" input fields, that should be cleared using the normal submit path ("RUN" or enter in a text field), but that should not be cleared using the "input-output" submit path
* add interactive beam search to the wikitables parser demo

* in addition, I added a `BeamTree` visualization for beam search, but using the default settings it looks bad, so it's not actually in the demo anywhere.

it's possible some of the design choices I made here are not great, so let me know.


jonborchardt:
can you add a screen shot of the relevant changes?

demo/package.json
```diff
@@ -16,6 +16,7 @@
     "react-scripts": "1.0.13",
     "react-syntax-highlighter": "8.0.1",
     "react-tooltip": "3.8.4",
+    "react-vis": "^1.11.6",
```
jonborchardt: did you consider any other libraries?

demo/src/css/BeamSearch.css
```diff
@@ -0,0 +1,69 @@
+ul.beam-search {
```
jonborchardt: i would recommend using styled components...

demo/src/components/BeamSearch.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import '../css/BeamSearch.css';
+
+// Component representing an element of the sequence that has been fixed,
+// with an X to "unchoose" it.
+const Chosen = ({action, unchoose, idx}) => (
+    <li className="chosen" key={idx}>
```
jonborchardt: index should not be used for key unless it is all we have...

demo/src/components/BeamSearch.js
```diff
@@ -0,0 +1,104 @@
...
+        <span className="unchoose" role="img" aria-label="x" onClick={unchoose}>❌</span>
+    </li>
+)
+
+// Component representing an element of the sequence that can be selected,
+// with a dropdown containing all the possible choices.
+const ChoiceDropdown = ({predictedAction, choices, choose, idx}) => {
+
+    const options = choices.map(([probability, action], i) => (
+        <li className="choice" key={i} onClick={() => choose(action)}>
```
jonborchardt: can we use the name for the key?

demo/src/components/demos/WikiTables.js
```diff
@@ -39,7 +39,8 @@ const fields = [
   {name: "table", label: "Table", type: "TEXT_AREA",
     placeholder: `E.g. "Season\tLevel\tDivision\tSection\tPosition\tMovements\n1993\tTier 3\tDivision 2\tÖstra Svealand\t1st\tPromoted\n1994\tTier 2\tDivision 1\tNorra\t11th\tRelegation Playoffs\n"`},
   {name: "question", label: "Question", type: "TEXT_INPUT",
-    placeholder: `E.g. "What is the only year with the 1st position?"`}
+    placeholder: `E.g. "What is the only year with the 1st position?"`},
+  {name: "beamSearch", type: "BEAM_SEARCH", optional: true, dependentInputs: ['initial_sequence'], inputOutput: true}
```
jonborchardt: have you considered breaking this into two

inputFields and inputOutputFields
this would make the code in demoInput cleaner i think

joelgrus:
the beam:

screenshot from 2019-02-22 16-25-15

mousing over:

screenshot from 2019-02-22 16-25-47

once you've selected an initial path:

screenshot from 2019-02-22 16-27-46

I am very bad at CSS, which is why it doesn't look that good.

demo/src/components/DemoInput.js
```diff
@@ -102,19 +103,40 @@ class DemoInput extends React.Component {
...
                 e.stopPropagation();
-                runModel(this.state)
+                runModel(this.independentInputs())
             }
         }
+
+        // Some of the inputs (e.g. interactive beam search)
+        // depend on the previous outputs, so when we do a new run
+        // we need to clear them out.
+        this.independentInputs = () => {
```
jonborchardt: conssider rename
freshInputs
cleanInputs
originalInputs?

jonborchardt:
let me know if you want me to take a crack at the css and tree

demo/package.json
joelgrus: I considered ones that were closer to the d3 metal, but I had trouble getting them to work

demo/src/css/BeamSearch.css
joelgrus: done

demo/src/components/BeamSearch.js
joelgrus: fixed

demo/src/components/BeamSearch.js
joelgrus: fixed

demo/src/components/demos/WikiTables.js
joelgrus: I'm not convinced this would make DemoInput cleaner, I'm inclined to file this under "premature optimization" and revisit once we have another model with input-outputs

demo/src/components/DemoInput.js
joelgrus: done

joelgrus:
ok, I addressed most of your comments. 

if you want to take a crack at the CSS that would be great. (I think the tree is less important and can wait for a future PR).

the difficulty is that the allennlp changes needed to support this are not merged yet.

you basically have to create a new environment, install from https://github.com/joelgrus/allennlp/tree/interactive-beam-search , and then run the demo backend with that code.

demo/src/components/demos/WikiTables.js
jonborchardt: sounds good

jonborchardt:
looks good... if we want to look at the styles, we can do that later

PR #171, opened 2019-03-02
Adding NAQANet demo

matt-gardner:
This currently relies on a funny commit; I'll update it once a PR gets merged in the main allennlp repo.  This PR makes two main changes:
1. It adds a radio button to select between BiDAF and the new NAQANet for reading comprehension, and all of the backend stuff to make the new model work.
2. It adds multiple example select boxes, so you can suggest different kinds of examples to a user.  I don't think I did anything here that would break permalinks, but let me know if you think I did.  Here's a screenshot of the new UI:
screen shot 2019-03-01 at 5 30 35 pm

### DIFF
--- a/demo/src/components/DemoInput.js
+++ b/demo/src/components/DemoInput.js
@@ -57,30 +57,37 @@ class DemoInput extends React.Component {
         super(props)
 
         const { examples, fields, inputState, runModel } = props
+        if (Array.isArray(examples)) {
+          this.normalizedExamples = {'default': {'text': 'Choose an example...', 'examples': examples}}
+        } else {
+          this.normalizedExamples = examples
+        }
 
         // Populate state using (a copy of) provided values.
         this.state = inputState ? {...inputState} : {}
 
+
         // What happens when you change the example dropdown
-        this.handleExampleChange = e => {
+        this.handleExampleChange = exampleType => { return e => {
             const exampleId = e.target.value
             if (exampleId !== "") {
                 // Because the field names vary by model, we need to be indirect.
                 let stateUpdate = {}
+                const example = this.normalizedExamples[exampleType]['examples'][exampleId]
 
                 // For each field,
                 fields.forEach(({name}) => {
                     // if the chosen example has a value for that field,
-                    if (examples[exampleId][name] !== undefined) {
+                    if (example[name] !== undefined) {
                         // include it in the update.
-                        stateUpdate[name] = examples[exampleId][name];
+                        stateUpdate[name] = example[name];
                     }
                 })
 
                 // And now pass the updates to setState.
                 this.setState(stateUpdate)
             }
-        }
+        }}
 
         // What happens when you change an input. This works for text
         // inputs and also select inputs. The first argument is
@@ -227,24 +234,29 @@ class DemoInput extends React.Component {
             }
         })
 
-
         return (
             <div className="model__content answer">
                 <ModelIntro title={title} description={description} descriptionEllipsed={descriptionEllipsed}/>
                 <div className="form__instructions">
                     <span>Enter text or</span>
-                    <select
-                        disabled={outputState === "working"}
-                        onChange={this.handleExampleChange}>
-                            <option value="">Choose an example...</option>
-                            {
-                                this.props.examples.map((example, index) => {
-                                    return (
-                                        <option value={index} key={index}>{makeSnippet(example, fields)}</option>
-                                    )
-                                })
-                            }
-                    </select>
+                    {
+                        Object.keys(this.normalizedExamples).map((example_type) => {
+                            const example_data = this.normalizedExamples[example_type]
+                            return (<div><select
+                                disabled={outputState === "working"}
+                                style={{"max-width": "17em"}}
+                                onChange={this.handleExampleChange(example_type)}>
+                                    <option value="">{example_data['text']}</option>
+                                    {
+                                        example_data['examples'].map((example, index) => {
+                                            return (
+                                                <option value={index} key={index}>{makeSnippet(example, fields)}</option>
+                                            )
+                                        })
+                                    }
+                            </select></div>)
+                        })
+                    }
                 </div>
                 {inputs}
                 <div className="form__field form__field--btn">
--- a/demo/src/components/DemoInput.js
+++ b/demo/src/components/DemoInput.js
@@ -247,17 +247,7 @@ class DemoInput extends React.Component {
                             <option value="">Choose an example...</option>
                             {
                                 this.normalizedExamples.map((exampleInfo, groupIndex) => {
-                                    return (
-                                      exampleInfo[0] !== "default" && <optgroup label={exampleInfo[0]}>
-                                      {
-                                          exampleInfo[1].map((example, index) => {
-                                              return (
-                                                  <option value={groupIndex + "@@" + index} key={groupIndex + "@@" + index}>{makeSnippet(example, fields)}</option>
-                                              )
-                                          })
-                                      }
-                                      exampleInfo[0] !== "default" && </optgroup>
-                                    )
+                                    return OptionGroup(exampleInfo, groupIndex, fields)
                                 })
                             }
                     </select>
@@ -280,6 +270,28 @@ class DemoInput extends React.Component {
             </div>
         )
     }
+}
+
+function OptionGroup(exampleInfo, groupIndex, fields) {
+  const exampleType = exampleInfo[0]
+  const examples = exampleInfo[1]
+  if (exampleType === "default") {
+      return RenderOptions(examples, groupIndex, fields)
+  } else {
+      return (
+          <optgroup label={exampleType}>
+              {RenderOptions(examples, groupIndex, fields)}
+          </optgroup>
+      )
+  }
+}
+
+function RenderOptions(examples, groupIndex, fields) {
+    return examples.map((example, index) => {
+        return (
+            <option value={groupIndex + "@@" + index} key={groupIndex + "@@" + index}>{makeSnippet(example, fields)}</option>
+        )
+    })
 }
 
 export { DemoInput as default, truncateText }
--- a/demo/src/components/DemoInput.js
+++ b/demo/src/components/DemoInput.js
@@ -57,8 +57,8 @@ class DemoInput extends React.Component {
         super(props)
 
         const { examples, fields, inputState, runModel } = props
-        if (Array.isArray(examples)) {
-          this.normalizedExamples = {'default': {'text': 'Choose an example...', 'examples': examples}}
+        if (!Array.isArray(examples[0])) {
+          this.normalizedExamples = [["default", examples]]
         } else {
           this.normalizedExamples = examples
         }
@@ -66,14 +66,15 @@ class DemoInput extends React.Component {
         // Populate state using (a copy of) provided values.
         this.state = inputState ? {...inputState} : {}
 
-
         // What happens when you change the example dropdown
-        this.handleExampleChange = exampleType => { return e => {
-            const exampleId = e.target.value
+        this.handleExampleChange = e => {
+            const parts = e.target.value.split('@@')
+            const exampleGroup = parts[0]
+            const exampleId = parts[1]
             if (exampleId !== "") {
+                const example = this.normalizedExamples[exampleGroup][1][exampleId]
                 // Because the field names vary by model, we need to be indirect.
                 let stateUpdate = {}
-                const example = this.normalizedExamples[exampleType]['examples'][exampleId]
 
                 // For each field,
                 fields.forEach(({name}) => {
@@ -87,7 +88,7 @@ class DemoInput extends React.Component {
                 // And now pass the updates to setState.
                 this.setState(stateUpdate)
             }
-        }}
+        }
 
         // What happens when you change an input. This works for text
         // inputs and also select inputs. The first argument is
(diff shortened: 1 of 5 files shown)


demo/src/components/DemoInput.js
```diff
@@ -227,24 +234,29 @@ class DemoInput extends React.Component {
...
-                                    )
-                                })
-                            }
-                    </select>
+                    {
+                        Object.keys(this.normalizedExamples).map((example_type) => {
+                            const example_data = this.normalizedExamples[example_type]
+                            return (<div><select
+                                disabled={outputState === "working"}
+                                style={{"max-width": "17em"}}
```
jonborchardt: i recommend reverting this, and instead using groups in a single options dropdown
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_optgroup

it will be a much smaller change, simply add optionGroupLabels
```
<select>
  <optgroup label="Swedish Cars">
    <option value="volvo">Volvo</option>
    <option value="saab">Saab</option>
  </optgroup>
  <optgroup label="German Cars">
    <option value="mercedes">Mercedes</option>
    <option value="audi">Audi</option>
  </optgroup>
</select>
```

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -300,24 +300,105 @@ const Output = (props) => {
...
 }
 
-const examples = [
-  {
-    passage: "A reusable launch system (RLS, or reusable launch vehicle, RLV) is a launch system which is capable of launching a payload into space more than once. This contrasts with expendable launch systems, where each launch vehicle is launched once and then discarded. No completely reusable orbital launch system has ever been created. Two partially reusable launch systems were developed, the Space Shuttle and Falcon 9. The Space Shuttle was partially reusable: the orbiter (which included the Space Shuttle main engines and the Orbital Maneuvering System engines), and the two solid rocket boosters were reused after several months of refitting work for each launch. The external tank was discarded after each flight.",
-    question: "How many partially reusable launch systems were developed?",
+const examples = {
+  'span_extraction': {
+    'text': 'Choose a SQuAD-like example...',
+    'examples': [
```
jonborchardt: examples format would change to
```'examples': [{items:[], groupName?:string}]```
and where used, add any items of a group without a groiuipoName to the toplevel (no optiongroup)
```
<select>
  <option value="volvo">Volvo</option> // had no group name
  <option value="saab">Saab</option> // had no group name
  <optgroup label="German Cars">
    <option value="mercedes">Mercedes</option> // in german group
    <option value="audi">Audi</option> // in german group
  </optgroup>
</select>
```
so most other models will use:
```'examples': [{items:[1,2,3,etc], groupName: undefined}]```

which will be translated to waht we have now
```
<select>
  <option value="1">1</option>
  <option value="2">2</option>
</select>
```

**Comment by matt-gardner** [comment]
Thanks for the suggestion, I like this a lot better.  It now looks like this: 
screen shot 2019-03-02 at 7 17 19 am

It's quite possible I wrote not-very-good javascript in making this work, as I don't know much of what I'm doing with this UI stuff, but we can fix that on Monday.  Now I'm just waiting on https://github.com/allenai/allennlp/pull/2560 getting merged, at while point I'll update the commit this PR points to, and merge this.

File: demo/src/components/demos/ReadingComprehension.js [hunk 3]
```diff
@@ -301,23 +301,85 @@ const Output = (props) => {
...
-    question: "Who stars in The Matrix?",
-  },
-  {
-    passage: "Kerbal Space Program (KSP) is a space flight simulation video game developed and published by Squad for Microsoft Windows, OS X, Linux, PlayStation 4, Xbox One, with a Wii U version that was supposed to be released at a later date. The developers have stated that the gaming landscape has changed since that announcement and more details will be released soon. In the game, players direct a nascent space program, staffed and crewed by humanoid aliens known as \"Kerbals\". The game features a realistic orbital physics engine, allowing for various real-life orbital maneuvers such as Hohmann transfer orbits and bi-elliptic transfer orbits.",
-    question: "What does the physics engine allow for?",
-  }
-].map(ex => ({...ex, snippet: truncateText(ex.passage)}));
+  ['SQuAD-like Span Extraction', [
```
matt-gardner: I'm debating whether this should be "SQuAD-like Argument Finding" instead, or something like that.

jonborchardt:
Looks great! I can help with the js Monday!

matt-gardner:
Merging now; I'll check staging to make sure this worked before deploying.

demo/src/components/DemoInput.js
```diff
@@ -57,23 +57,31 @@ class DemoInput extends React.Component {
         super(props)
 
         const { examples, fields, inputState, runModel } = props
+        if (!Array.isArray(examples[0])) {
+          this.normalizedExamples = [["default", examples]]
```
jonborchardt: i am not a fan of this, it toggles the type
we are inviting runtime errors as we are not type safe
as we move to typescript, this logic will be hard to understand

demo/src/components/DemoInput.js
```diff
@@ -57,23 +57,31 @@ class DemoInput extends React.Component {
...
 
         // Populate state using (a copy of) provided values.
         this.state = inputState ? {...inputState} : {}
 
         // What happens when you change the example dropdown
         this.handleExampleChange = e => {
-            const exampleId = e.target.value
+            const parts = e.target.value.split('@@')
+            const exampleGroup = parts[0]
+            const exampleId = parts[1]
```
jonborchardt: need to test len of parts to be safe

demo/src/components/DemoInput.js
```diff
@@ -57,23 +57,31 @@ class DemoInput extends React.Component {
...
             if (exampleId !== "") {
+                const example = this.normalizedExamples[exampleGroup][1][exampleId]
```
jonborchardt: the index is unsafe

demo/src/components/DemoInput.js
```diff
@@ -57,23 +57,31 @@ class DemoInput extends React.Component {
...
                 let stateUpdate = {}
 
                 // For each field,
                 fields.forEach(({name}) => {
                     // if the chosen example has a value for that field,
-                    if (examples[exampleId][name] !== undefined) {
+                    if (example[name] !== undefined) {
                         // include it in the update.
-                        stateUpdate[name] = examples[exampleId][name];
+                        stateUpdate[name] = example[name];
```
jonborchardt: clearer:
stateUpdate[name] = example[name] ||  stateUpdate[name];

demo/src/components/DemoInput.js
```diff
@@ -266,4 +272,26 @@ class DemoInput extends React.Component {
     }
 }
 
+function OptionGroup(exampleInfo, groupIndex, fields) {
+  const exampleType = exampleInfo[0]
+  const examples = exampleInfo[1]
```
jonborchardt: need to check len

demo/src/components/DemoInput.js
```diff
@@ -266,4 +272,26 @@ class DemoInput extends React.Component {
...
+  if (exampleType === "default") {
```
jonborchardt: default is a magic string.... what does it mean?

jonborchardt: this should be the default if undefined

demo/src/components/DemoInput.js
```diff
@@ -266,4 +272,26 @@ class DemoInput extends React.Component {
...
+              {RenderOptions(examples, groupIndex, fields)}
+          </optgroup>
+      )
+  }
+}
+
+function RenderOptions(examples, groupIndex, fields) {
+    return examples.map((example, index) => {
+        return (
+            <option value={groupIndex + "@@" + index} key={groupIndex + "@@" + index}>{makeSnippet(example, fields)}</option>
```
jonborchardt: this looks brittle, can we pass this data in the click even instead of having it in the value?

demo/src/components/DemoInput.js
matt-gardner: It does not toggle any type in this, because it's writing to a new variable.  It assumes that `examples` has a union type in `props`.  I don't think it's a good idea to make all of the other demos change the way they specify their examples just so that we can do something fancy in the reading comprehension demo.  Any suggestions on what else to do?

demo/src/components/DemoInput.js
matt-gardner: This is logic that is entirely internal to this class, and it's something that I would handle with a test instead of an assert in python.  What's the right way to do this in javascript?

demo/src/components/DemoInput.js
matt-gardner: Same as above.  I constructed this object and all UI elements internally to the class, so I know that it is safe (and I'd use a test to make sure of that in python, not a runtime assert).

demo/src/components/DemoInput.js
matt-gardner: I'm not sure why that's clearer - I actually am not sure just looking at that line what it would do in this context, because I'm not super familiar with javascript, but the if statement and comments are pretty easy to understand. (And it's logic that I'm not changing in this PR.)

demo/src/components/DemoInput.js
matt-gardner: I don't understand what you mean here.

demo/src/components/DemoInput.js
jonborchardt: we don't want a divergence of how models do this.... they should be kept as similar as possible
and i think giving them this option is great, it allows them to make nicer examples if they want to

i will convert the others to [default:{originalExamples}]if you like

demo/src/components/DemoInput.js
matt-gardner: The trouble is that the event is `onChange` in the select box, not on the option.  I agree that it'd be nicer to do this some other way, but I don't know how to do that.  I tried a bunch of different things before finding this one that actually worked.

demo/src/components/DemoInput.js
matt-gardner: Same comment as above about this being internal to the class.

demo/src/components/DemoInput.js
jonborchardt: if(parts.length>1){
  logic
}
else handle error

jonborchardt: i would also suggest writing an encode and decode function to keep this logic contained

```
// psudo
class valueEncoder {
const sep = '@@'
encode(x,y) return `${x}${sep}${y};
decode(s) {
  spl = s.split(sep)
  return {x: spl.length?spl[0]:undefined, y:spl.length>1?slp[1]:undefined}
}
}
```

demo/src/components/DemoInput.js
matt-gardner: I guess I'm more a fan of "make the typical case easy", and it's unfortunate to me to complicate the other cases just to allow for fancier stuff.  But if you want to make the change (and figure out a way to make this a dictionary instead of a list), go for it.

jonborchardt: this format is odd
i would expect one of the following
this.normalizedExamples = [{"default": examples}]
this.normalizedExamples = [{label: "default", values: examples}]

as is you have an array of an array with items of the array having different types
this.normalizedExamples = [["default", [example1, example2, etc]]]
this is brittle IMO

demo/src/components/DemoInput.js
jonborchardt: a unittest would be fine, but we can also jist protect against it...

i think my suggestion on interface change to the data will make this easier to understand and less brittle.

demo/src/components/DemoInput.js
matt-gardner: Yes, I agree that this is brittle, I just wasn't sure how to preserve order and get a dictionary without being overly verbose.  Not sure why I didn't think of `[{"default": examples}]`.  That should work.  I'll try that.

demo/src/components/DemoInput.js
jonborchardt: fair enough, i am just showing more modern looking js
A = B or A
a gets b if it exists, else it remains a

demo/src/components/DemoInput.js
jonborchardt: i replied elsewhere, lets kill this thread and combine them

demo/src/components/DemoInput.js
jonborchardt: checking for magic strings... "default"
leads to runtime errors because we can fat finger the string...
safer to set a constant or change the data interface as i mention elsewhere

demo/src/components/DemoInput.js
jonborchardt: ahh.. you are right
adding the encoder will make this safer

demo/src/components/DemoInput.js
matt-gardner: Oh, I can't do this without changing all of the demos, because it's a mess to tell if you have the fancier input or the non-fancy input.

demo/src/components/DemoInput.js
matt-gardner: Ok, yes, that makes sense, I'll add encode/decode functions.

demo/src/components/DemoInput.js
jonborchardt: the trouble comes in when the next model makes a new case....
i think the longer term fix is to add a class that takes one or more sets of questions..... and makes the format for them

```class ExampleThing {
  function AddExample(question, answer, optionalGroupName)
  function GetExamples() // returns in a proper format (and does the map work for trimming etrc...)
}```

File: demo/src/components/DemoInput.js (hunk 6 above)
Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: Yeah, I agree with changing the data interface, but it's a larger change than I can do right now, so I will leave that one for later.

File: demo/src/components/DemoInput.js (hunk 7 above)
Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: I get the general pattern; the reason it's confusing me in this case is because `stateUpdate[name]` might be empty, and it seems odd to add things to a dictionary this way.  E.g., will your logic end up with more keys in the dictionary than the existing logic?

File: demo/src/components/DemoInput.js (hunk 9 above)
Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: What I didn't understand was "this should be the default if undefined".  I totally agree with not using magic strings, I should have done that differently.

File: demo/src/components/DemoInput.js (hunk 4 above)
Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: Some kind of function like that seems like a good idea.  I approve =).

Comment (jonborchardt) [reply demo/src/components/DemoInput.js]: i suggest updating the other examples, but if not, i think you can:
```
if (examples[0].question) { // top level item has a question (old interface)
          // convert to new interface
          this.normalizedExamples = [{"default": examples}]
```
then in your new model
```
examples = [{"sec1": exampleArray1}, {"sec2": exampleArray2}]
```

Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: Resolving this conversation, as it's the same as the first one above.

Comment (jonborchardt) [reply demo/src/components/DemoInput.js]: ok, i still suggest altering the interface to be [label:array, label:array] (see other comments)

Comment (matt-gardner) [reply demo/src/components/DemoInput.js]: Sorry, I resolved right as you posted.  The trouble with your suggestion here is that it will break the OpenIE, NER, Coref, and other demos.  They don't have questions.

File: demo/src/components/DemoInput.js (hunk 7 above)
Comment (jonborchardt) [reply demo/src/components/DemoInput.js]: ahh... worth checking... but it should be name:undefined.... which should not add a key... maybe im wrong
anyway, its fine as is...

File: demo/src/components/DemoInput.js (hunk 9 above)
Comment (jonborchardt) [reply demo/src/components/DemoInput.js]: ahhh....
instead of 
if(exampleType === "default")
i propose
if (!exampleType || exampleType === "default"))

File: demo/src/components/DemoInput.js (hunk 4 above)
Comment (jonborchardt) [reply demo/src/components/DemoInput.js]: ug, this is all messed up... so the bag coming back for different models is already inconsistent... which makes extending it a pain

i can offer:

```
if (!examples[0].default) { // (old interface)
          // convert to new interface
          this.normalizedExamples = [{"default": examples}]
```
then in your new model
```
examples = [{"default"}:[], {"sec1": exampleArray1}, {"sec2": exampleArray2}]
```
then in the code that makes groupLabes, dont make a group if the label is empty...
```
Object.keys(normalizedExamples ).map(k=> if(normalizedExamples[k].length)(/*wrap items in group*/) else /*dont wrap items*/
```

demo/src/components/DemoInput.js
matt-gardner: Done.

jonborchardt: BTW, im heading out...
tty on monday!

demo/src/components/DemoInput.js
matt-gardner: I think all of these issues go away if we have a class that handles adding examples.  I think that's the much cleaner solution here.  I left a TODO in the code to fix this, but I'm going to leave the current code as it is, because it works, and it doesn't make sense to change it if we're just going to change it again later anyway.

PR #173, opened 2019-03-03
Minor tweaks to the reading comprehension UI

matt-gardner:
This was based on feedback from DROP team members using what was in staging:

- I fixed a bug where the javascript displayed "no answer" when the model predicted a numerical value of 0
- I added "(trained on SQuAD/DROP)" to the model names
- I made the example drop down show questions instead of passages
- I disabled the slider and hid the attention by default

@jonborchardt, I don't think there's any particularly iffy javascript stuff in here, but I'd still appreciate a quick glance on Monday, even though I'm going to merge this now.

### DIFF
--- a/demo/src/components/demos/ReadingComprehension.js
+++ b/demo/src/components/demos/ReadingComprehension.js
@@ -30,18 +30,18 @@ const descriptionEllipsed = (
 
 const taskModels = [
   {
-    name: "BiDAF",
+    name: "BiDAF (trained on SQuAD)",
     desc: "Reimplementation of BiDAF (Seo et al, 2017), or Bi-Directional Attention Flow,<br/>a widely used MC baseline that achieved state-of-the-art accuracies on<br/>the SQuAD dataset (Wikipedia sentences) in early 2017."
   },
   {
-    name: "NAQANet",
+    name: "NAQANet (trained on DROP)",
     desc: "An augmented version of QANet that adds rudimentary numerical reasoning ability,<br/>trained on DROP (Dua et al., 2019), as published in the original DROP paper."
   }
 ]
 
 const taskEndpoints = {
-  "BiDAF": "machine-comprehension", // TODO: we should rename tha back-end model to reading-comprehension
-  "NAQANet": "naqanet-reading-comprehension"
+  "BiDAF (trained on SQuAD)": "machine-comprehension", // TODO: we should rename tha back-end model to reading-comprehension
+  "NAQANet (trained on DROP)": "naqanet-reading-comprehension"
 };
 
 const fields = [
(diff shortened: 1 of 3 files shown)


demo/src/components/demos/ReadingComprehension.js
```diff
@@ -202,7 +202,7 @@ const AnswerByType = ({requestData, responseData}) => {
 
       case "count": {
         const { count } = answer || {};
-        if(question && passage && count) {
+        if(question && passage && count.toString()) {
```
jonborchardt: this is unsafe, what is intended here
as is, if count is undefined, this will cause a runtime exception

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -230,7 +230,7 @@ const AnswerByType = ({requestData, responseData}) => {
 
       case "arithmetic": {
         const { numbers, value } = answer || {};
-        if(question && passage && numbers && value) {
+        if (question && passage && numbers && value.toString()) {
```
jonborchardt: same as above

demo/src/components/demos/ReadingComprehension.js
```diff
@@ -377,7 +377,7 @@ const examples = [
           passage: "Kannada language is the official language of Karnataka and spoken as a native language by about 66.54% of the people as of 2011. Other linguistic minorities in the state were Urdu (10.83%), Telugu language (5.84%), Tamil language (3.45%), Marathi language (3.38%), Hindi (3.3%), Tulu language (2.61%), Konkani language (1.29%), Malayalam (1.27%) and Kodava Takk (0.18%). In 2007 the state had a birth rate of 2.2%, a death rate of 0.7%, an infant mortality rate of 5.5% and a maternal mortality rate of 0.2%. The total fertility rate was 2.2.",
           question: "Which linguistic minority is larger, Hindi or Malayalam?"
         },
-    ].map(ex => ({...ex, snippet: truncateText(ex.passage)}))],
+    ].map(ex => ({...ex, snippet: truncateText(ex.question)}))],
```
jonborchardt: this map should be a function, since its used x times
(or do it in one map on the outer array, instead of in x inner arrays

demo/src/components/demos/ReadingComprehension.js
matt-gardner: So what's the right check here?  I don't know javascript well enough to know what to do.  The problem was that a numerical value of 0 gives incorrect behavior in the old logic.

demo/src/components/demos/ReadingComprehension.js
matt-gardner: Done (in a PR that's coming).

demo/src/components/demos/ReadingComprehension.js
jonborchardt: count !== undefined

demo/src/components/demos/ReadingComprehension.js
jonborchardt: value !== undefined

demo/src/components/demos/ReadingComprehension.js
matt-gardner: Done.

demo/src/components/demos/ReadingComprehension.js
matt-gardner: Done.

PR #174, opened 2019-03-03
Give the slider an option to always show all columns, add it back in

matt-gardner:
I misunderstood the feedback on the UI; this adds back in the slider, but makes all question words always visible (because the question is typically short).

### DIFF
--- a/demo/src/components/demos/ReadingComprehension.js
+++ b/demo/src/components/demos/ReadingComprehension.js
@@ -69,7 +69,7 @@ const Attention = ({passage_question_attention, question_tokens, passage_tokens}
                 </p>
                 <HeatMap
                   colLabels={question_tokens} rowLabels={passage_tokens}
-                  data={passage_question_attention} />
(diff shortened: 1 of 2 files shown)


demo/src/components/demos/ReadingComprehension.js
```diff
@@ -69,7 +69,7 @@ const Attention = ({passage_question_attention, question_tokens, passage_tokens}
                 </p>
                 <HeatMap
                   colLabels={question_tokens} rowLabels={passage_tokens}
-                  data={passage_question_attention} />
+                  data={passage_question_attention} includeSlider={true} showAllCols={true} />
```
jonborchardt: includeslider should be true by default IMO

matt-gardner: The majority of these sliders are in the semantic parsers, where they are not really helpful.  The slider is most helpful when you have really big matrices, and this is the only demo where we have those, and it's enabled here.

jonborchardt: ok

PR #177, opened 2019-03-03
Add reading-comprehension and nlvr-parser to routing

### LINKED ISSUE #176: Direct links for reading comprehension and nlvr parser 404
https://demo.allennlp.org/nlvr-parser
https://demo.allennlp.org/reading-comprehension

matt-gardner:
Fixes #176.


matt-gardner:
This is live on staging, and it appears to be fixed.  It's unfortunate that we have to merge PRs to really test this - I was pretty sure this solved the problem when I tested it locally, but it'd be nice to be able to see a deployed version of a PR _before_ merging it.

app.py
```diff
@@ -300,11 +300,13 @@ def info() -> Response:  # pylint: disable=unused-variable
     @app.route('/constituency-parsing')
     @app.route('/dependency-parsing')
     @app.route('/machine-comprehension')
```
jonborchardt: why are we keeping the old route? not sure the FE is set up to handle it...

app.py
```diff
@@ -313,11 +315,13 @@ def info() -> Response:  # pylint: disable=unused-variable
     @app.route('/constituency-parsing/<permalink>')
     @app.route('/dependency-parsing/<permalink>')
     @app.route('/machine-comprehension/<permalink>')
```
jonborchardt: same

app.py
matt-gardner: Because I didn't want to break anything.  I just checked what's on staging, and it looks like I didn't need to be worried.  Fixing this.

PR #179, opened 2019-03-11
Cache /Users/michael/.allennlp when running Docker commands.

schmmd:


schmmd:
@brendan-ai2 unlike https://github.com/allenai/allennlp/pull/2593 I think this change is essential (and also much better tested).

PR #183, opened 2019-03-26
allow single-model operation

joelgrus:
this PR has two changes (and some refactoring to enable them)

1. if an environment variable `ALLENNLP_MODEL_NAME` is set (e.g. to `machine-comprehension`) then the app will load only that model. Otherwise, it will load all the models in the `MODELS` variable.

2. there was some really clunky behavior where we manually added a path for each model to serve index.html at /model_name and /model_name/<permalink>. I made that automatic based on the keys of the models passed into the constructor.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,4 @@ ENV ALLENNLP_DEMO_SOURCE_COMMIT $SOURCE_COMMIT
 
 EXPOSE 8000
 
-ENV ALLENNLP_DEMO_DIRECTORY /stage/allennlp/demo
-
-ENTRYPOINT ["./app.py"]
+ENTRYPOINT ["./app.py", "--demo-dir /stage/allennlp/demo"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,4 +37,5 @@ ENV ALLENNLP_DEMO_SOURCE_COMMIT $SOURCE_COMMIT
 
 EXPOSE 8000
 
-ENTRYPOINT ["./app.py", "--demo-dir /stage/allennlp/demo"]
+ENTRYPOINT ["./app.py"]
+CMD ["--demo-dir", "/stage/allennlp/demo"]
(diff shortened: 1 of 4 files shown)


schmmd:
Also, if the models were passed as a CLI argument we could support a list in addition to a single model--which might be nice for development sometimes (you could exclude the really slow models).

joelgrus:
take a look at it now (and in particular at the Dockerfile, which I'm not 100% sure I made the right change to) (I tried building it but ran into the jessie error)

Dockerfile
```diff
@@ -39,6 +39,4 @@ ENV ALLENNLP_DEMO_SOURCE_COMMIT $SOURCE_COMMIT
 
 EXPOSE 8000
 
-ENV ALLENNLP_DEMO_DIRECTORY /stage/allennlp/demo
-
-ENTRYPOINT ["./app.py"]
+ENTRYPOINT ["./app.py", "--demo-dir /stage/allennlp/demo"]
```
schmmd: This stuff is obnoxiously confusing.  The above would work if you split your argument in to, but ideally you would do the following:

```
ENTRYPOINT ["./app.py"]
CMD ["--demo-dir", "/stage/allennlp/demo"]
```

This way, on the command line people can easily overwrite the arguments to Docker (`CMD`) but we make it clear what is our preferred `ENTRYPOINT`.

That said--I recommend testing it by doing a build and running the demo locally ;-)

app.py
```diff
@@ -341,4 +327,15 @@ def static_css_proxy(path: str) -> Response: # pylint: disable=unused-variable
     return app
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser("start the allennlp demo")
+    parser.add_argument('--port', type=int, default=8000, help='port to serve the demo on')
+    parser.add_argument('--demo-dir', type=str, default='demo/', help="directory where the demo HTML is located")
+    parser.add_argument('--cache-size', type=int, default=128, help="how many results to keep in memory")
+    parser.add_argument('--model-name', type=str, action='append', default=[], help='if specified, only load these models')
```
schmmd: I might change `--model-name` to just `--model` but it's up to you.

schmmd:
Thanks @joelgrus -- this is a huge improvement.

schmmd:
To avoid the `jessie` error when https://github.com/allenai/allennlp/pull/2647 is merged we should bump up the commit we're depending on.

joelgrus:
ok, I fixed the docker file and renamed the parameter to `model` and tested it by building and running the docker image locally

PR #184, opened 2019-03-26
Update NLVR model to one with a compatible vocabulary

matt-gardner:
This error was (somewhat surprisingly) unrelated to the QuaRel demo problem (though they were both introduced in the big semantic parser refactoring that I did a while ago).  I just needed to update the saved model to something that's compatible with new code.  There is, unfortunately, not a good place to add a test for this.  A sniff test is the right thing here, but there aren't any of those in this repo, and if we're splitting things out into separate services, anyway, this wouldn't have ever been a problem (because the cause of the demo breaking was that the code changed from underneath what the demo model was expecting).

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -88,7 +88,7 @@ MODELS = {
                 8177
         ),
         'nlvr-parser': DemoModel(
-                'https://s3-us-west-2.amazonaws.com/allennlp/models/nlvr-erm-model-2018-12-18.tar.gz',
+                'https://s3-us-west-2.amazonaws.com/allennlp/models/nlvr-erm-model-2018-12-18-rule-vocabulary-updated.tar.gz',
                 'nlvr-parser',
                 1136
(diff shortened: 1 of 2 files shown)


brendan-ai2:
Sorry for not catching this yesterday. Feel free to ping on Slack for bugfix PRs.

@matt-gardner, I think we do have a class of tests that could have caught this. See for instance https://github.com/allenai/allennlp-demo/blob/master/tests/server_test.py#L128. They test model fixtures against the current code. Not as good as sniff tests against the full model, but definitely better than nothing.

matt-gardner:
Hmm, you're right.  For some reason I was thinking that because it was a separate fixture that wasn't getting updated, it wouldn't have caught this, but you're right that it would have.

It still will be obviated by the work Joel has planned for Q2 to split these out, so they won't depend on current code, anyway.

PR #185, opened 2019-03-27
Updating allennlp commit to fix quarel demo

### LINKED ISSUE #182: the quarel demo is broken
none of the examples work for me, they certainly used to work, I'm not sure if it's because of a backend change or a frontend change :unamused:

matt-gardner:
Fixes #182.  See https://github.com/allenai/allennlp/pull/2648.


joelgrus:
we should update it to a sha from after we merge the "update dockerfile" PR, since until then it won't even build

matt-gardner:
Thanks, I thought there was something like that, but lost track of it when I submitted this.

PR #188, opened 2019-04-22
add option to run the demo with no models

joelgrus:
this will allow us to run a "front-end" pod

### DIFF
--- a/app.py
+++ b/app.py
@@ -73,14 +73,11 @@ def main(demo_dir: str,
         logger.warning("demo db credentials not provided, so not using demo db")
 
     # If environment variables said to load only one model, only load that model.
-    if model_names:
-        models = {
-            name: demo_model
-            for name, demo_model in MODELS.items()
-            if name in model_names
-        }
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -332,10 +322,28 @@ def static_css_proxy(path: str) -> Response: # pylint: disable=unused-variable
     parser.add_argument('--demo-dir', type=str, default='demo/', help="directory where the demo HTML is located")
     parser.add_argument('--cache-size', type=int, default=128, help="how many results to keep in memory")
     parser.add_argument('--model', type=str, action='append', default=[], help='if specified, only load these models')
+    parser.add_argument('--no-models', dest='no_models', action='store_true', help='start just the front-end with no models')
```
schmmd: Please add `--model` and `--no-model` to a group and make them disjoint.  See https://docs.python.org/3/library/argparse.html#mutual-exclusion.  It won't complicate your code any but it will make for better user behavior and usage notes.

PR #191, opened 2019-04-27
Add interpretation visualization component

jens321:
The interpretation component added gives a rough idea of what visualizing interpretations (currently based on vanilla gradients) for tokens could look like in the UI. The current component expects tokens and normalized gradients for textual entailment.


matt-gardner:
@jonborchardt, here's the PR I was talking about.  To get something that's more easily re-usable across different demos, I think we should separate out a component that just takes the words and the weights, and returns a colored background.  It could take the particular colorscheme as a prop.  Then the SNLI demo can have the explanatory text and create two of these components with the right parameters.  So, basically, @jens321, your current `colorize` function should be split out into a component with some reasonable name, and that's the only part of this PR that we need.  The component can be added to the SNLI demo later, when we have the other pieces ready for it.

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import colormap from 'colormap'
+import assert from 'assert'
+
+class Interpretation extends React.Component {
```
jonborchardt: class name should be a bit more descriptive...

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+        }
+        } key={idx}>{token} </span>)
+    })
+
+    return result_string 
+  }
+
+  render() {
+
+    const { premise_tokens, hypothesis_tokens, premise_normalization, hypothesis_normalization } = this.props
```
jonborchardt: the allennlp demo has typescript support, so we should use it to define a props interface
```
interface Props {
  premiseTokens: string[];
  hypothesisTokens: string[];
  premiseNormalization: number[];
  hypothesisNormalization: number[];
  
  // break out to sep class, but if in one class add
  colorScheme: "RdBu" | "jet" | "hsv" | "etc...";
  nShades: number;
}
```

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+  render() {
+
+    const { premise_tokens, hypothesis_tokens, premise_normalization, hypothesis_normalization } = this.props 
+
+    const premise_color_map = this.colorize(premise_tokens, premise_normalization)
+    const hypothesis_color_map = this.colorize(hypothesis_tokens, hypothesis_normalization)
+
+    return (
+      <div>
+        <div className="model__content__summary">
```
jonborchardt: allennlp demo now supports styled components, so we should use that instead of inline classes

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+          The interpretation map below shows a <strong>red-blue colormap</strong> where:
+          <ul>
+            <li><strong>Red </strong>is a negative contribution to the prediction</li>
```
jonborchardt: consider showing the color (red or blue) instead or in addition to the word)

demo/src/components/Interpretation.jsx
jonborchardt: i agree with @matt-gardner 
consider breaking out color scheme and colormap from this class into its own class that can be parameterized
then in the props here, default to the colors you have, but let the user pass in a trumping color map

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+class Interpretation extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.colorize = this.colorize.bind(this);
+  }
+
+  colorize(tokens, grads) {
+    let colors = colormap({
+      colormap: 'RdBu',
```
jonborchardt: the color (RdBu) should be parameters

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+      result_string.push(<span style={
+        {
+          // There is 72 different shades in the spectrum 
+          backgroundColor: colors[Math.round(grads[idx] * 71)],
+          padding: "5px",
+          margin: "5px",
+          display: "inline-block",
+          borderRadius: "3px"
+        }
+        } key={idx}>{token} </span>)
```
jonborchardt: adding a class also allows users to alter the style via css

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+      colormap: 'RdBu',
+      format: 'hex'
+    })
+
+    // make sure tokens and grads have same length
+    assert.strictEqual(tokens.length, grads.length)
+
+    let result_string = [];
+    tokens.forEach((token, idx) => {
+      result_string.push(<span style={
```
jonborchardt: instead of inline style, add a class or use styled components

each style other than color should just be applied via css class
color you either need to do as is, or pass as parameter to styledcomponents

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  colorize(tokens, grads) {
+    let colors = colormap({
+      colormap: 'RdBu',
+      format: 'hex'
+    })
+
+    // make sure tokens and grads have same length
+    assert.strictEqual(tokens.length, grads.length)
```
jonborchardt: instead of asserting, you can have the caller map into the correct data format
```[{token, grad}, {token, grad}, ...]```

File: demo/src/components/Interpretation.jsx [hunk 9]
```diff
@@ -0,0 +1,69 @@
...
+  constructor(props) {
+    super(props);
+
+    this.colorize = this.colorize.bind(this);
+  }
+
+  colorize(tokens, grads) {
+    let colors = colormap({
+      colormap: 'RdBu',
+      format: 'hex'
```
jonborchardt: i would make a parameter for nshades
as a default, i would add nshades=20
72 is too many

demo/src/components/Interpretation.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    // make sure tokens and grads have same length
+    assert.strictEqual(tokens.length, grads.length)
+
+    let result_string = [];
+    tokens.forEach((token, idx) => {
+      result_string.push(<span style={
+        {
+          // There is 72 different shades in the spectrum 
+          backgroundColor: colors[Math.round(grads[idx] * 71)],
```
jonborchardt: use nshades parameter instead of hardcoding 71

jonborchardt: ```backgroundColor: grads[idx] ? colors[Math.round(grads[idx] * (props.nShades-1))] : 'transparent',```

demo/src/components/Interpretation.jsx
jonborchardt: color names would need to be part of the colorscheme passed in.... or dont show the color name :-P

demo/src/components/Interpretation.jsx
jens321: Just to make sure, do you mean these: https://www.styled-components.com ?

demo/src/components/Interpretation.jsx
jens321: When adding a class to this <span> element, that would not go inside this file right? I would have to define it inside the CSS files?

demo/src/components/Interpretation.jsx
jens321: When showing the color in addition to the word, could that, for example, be in the form of a background color that is representative of how it would look inside the actual visualization?

jens321:
@jonborchardt Thanks for all the feedback! I just added some clarification questions here and there to make sure I am understanding all the changes to be made correctly. It'd definitely be great if you could take another look once I'm done fixing the changes. Thanks again.

demo/src/components/Interpretation.jsx
jens321: Would this interface be declared in a separate file with a d.ts extension such as explained here: https://charleslbryant.gitbooks.io/hello-react-and-typescript/content/DefinePropsAndStateTypes.html ?

jonborchardt: Just put it directly above the class I the same file.

demo/src/components/Interpretation.jsx
jonborchardt: Maybe. You need to play and see what looks nice.

I would try a square preceding the text. Just see what looks best. Maybe I'm wrong 😎

jonborchardt: Maybe. You need to play and see what looks nice.

I would try a square preceding the text. Just see what looks best. Maybe I'm wrong 😎

demo/src/components/Interpretation.jsx
jonborchardt: Same file is probably ok, but a different file is also fine.... Not in the css.

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react'
+import colormap from 'colormap'
+import styled from 'styled-components';
+
+interface Props {
```
jonborchardt: Props can extend DefaultProps
```
interface Props extends DefaultProps {
  tokensWithWeights: {token: string, weight: number}[];
}
```

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  static defaultProps: DefaultProps = {
+    colorScheme: 'RdBu',
+    nShades: 20
+  }
+
+  colorize(tokensWithWeights: {token: string, weight: number}[]) {
+    let colors = colormap({
+      colormap: this.props.colorScheme,
```
jonborchardt: instead of passing in colormap and nshades as props
you can define a type for ColorMap and just pass it in
```
interface ColorMap {
  colormap: ColorScheme
  format: ColorMapFormat // define as 'hex' | 'bin' | ' or whatever the options are'
  nshades:  number
}

// then

interface Props {
  colormap: ColorMap
  tokensWithWeights: {token: string, weight: number}[];
}
```

// this can be combined with other suggestions via
```
DefaultProps {
 colormap: ColorScheme
}

Props extends DefaultProps {
    tokensWithWeights: {token: string, weight: number}[];
}
```

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,88 @@
...
+      format: 'hex',
+      nshades: this.props.nShades
+    })
+
+    let result_string: JSX.Element[] = [];
+    tokensWithWeights.forEach((obj: {token: string, weight: number}, idx: number) => {
+      result_string.push(
+        <ColorizedToken color={obj['weight'] ? colors[Math.round(obj['weight'] * (this.props.nShades - 1))] : 'transparent'}
```
jonborchardt: backgroundColor not color

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,88 @@
...
+    nShades: 20
+  }
+
+  colorize(tokensWithWeights: {token: string, weight: number}[]) {
+    let colors = colormap({
+      colormap: this.props.colorScheme,
+      format: 'hex',
+      nshades: this.props.nShades
+    })
+
```
jonborchardt: ensure nShades > 0

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,88 @@
...
+               "plasma" | "warm" | "rainbow-soft" | "bathymetry" |
+               "cdom" | "chlorophyll" | "density" | "freesurface-blue" |
+               "freesurface-red" | "oxygen" | "par" | "phase" | "salinity" |
+               "temperature" | "turbidity" | "velocity-blue" |
+               "velocity-green" | "cubehelix";
+  nShades: number;
+}
+
+const ColorizedToken = styled.span`
+  background-color: ${props => props.color};
```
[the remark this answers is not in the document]
jens321: For (1), what exact line/lines are you referring to?

demo/src/components/Interpretation.tsx
jens321: What would be a good way to ensure this? I was thinking either throwing an error via `throw`, using react error boundaries, or using asserts.

demo/src/components/Interpretation.tsx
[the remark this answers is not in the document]
jonborchardt: 22

demo/src/components/Interpretation.tsx
jonborchardt: nshades: Math.min(Math.max(this.props.nShades, 1), 72)

demo/src/components/Interpretation.tsx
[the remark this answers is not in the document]
jens321: Yes, line 22 does seem to work.

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,91 @@
...
+      colormap: this.props.colormap.colorScheme,
+      format: this.props.colormap.format,
+      nshades: Math.min(Math.max(this.props.colormap.nShades, 1), 72)
+    })
+
+    let result_string: JSX.Element[] = [];
+    tokensWithWeights.forEach((obj: {token: string, weight: number}, idx: number) => {
+      result_string.push(
+        <ColorizedToken backgroundColor={obj['weight'] ? colors[Math.round(obj['weight'] * (this.props.colormap.nShades - 1))] : 'transparent'}
+                        key={idx}>{obj['token']}
```
jonborchardt: obj.weight and obj.token
not obj['weight'] etc

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface Props extends DefaultProps {
+  tokensWithWeights: {token: string, weight: number}[];
+}
+
+interface DefaultProps {
+  colormap: ColorMap
+}
+
+interface ColorMap {
```
jonborchardt: colormap is a confusing name, since it is a function also..
lets rename this interface to ColorMapProps or something... naming is hard

demo/src/components/Interpretation.tsx
```diff
@@ -0,0 +1,91 @@
...
+  static defaultProps: DefaultProps = {
+    colormap: {
+      colorScheme: 'RdBu',
+      format: 'hex',
+      nShades: 20
+    }
+  }
+
+  colorize(tokensWithWeights: {token: string, weight: number}[]) {
+    let colors = colormap({
```
jonborchardt:
```
const {colormapProps} = this.props.colormapProps;
colormapProps.nShades =  Math.min(Math.max(colormapProps.nShades, 1), 72);
let colors = colormap(colormapProps);
```

jens321:
@jonborchardt I added the changes (I'll push them soon), but I wanted to make sure it works when someone instantiates this component in one of the demos and that there are no typescript errors etc. How would I best go about this? 
Currently, I have just been using the component in my own repo, but to integrate with typescript I needed a tsconfig.json and a file to `declare module 'colormap';`. I didn't see any `.ts` files in the allennlp-demo repo, so I wasn't sure if my current file will correctly integrate with the repo?
Thanks for clarifying.

jonborchardt:
I misspoke, the allennlp demo does not use ts... but its still goo that this component does.
where will this code live?
may i suggest: https://github.com/allenai/varnish
or you can add it directly to the demo

demo/src/components/Interpretation.tsx
jonborchardt: did you decide not to do this?

demo/src/components/Interpretation.tsx
jonborchardt: did you decide not to do this?

demo/src/components/Interpretation.tsx
jonborchardt: did you decide not to do this?

jens321:
@jonborchardt sounds good, I think moving this component to allenai/varnish sounds like a good idea (and it looks like it should be perfectly possible to import from there to the demo). Just want to confirm with @matt-gardner?

matt-gardner:
I don't have a strong opinion on where things live; I would have defaulted to just having it in the demo repository, but if there are typescript issues with that, then just do what makes the most sense.

jonborchardt:
i would propose simply adding it to the allennlp demo site (without types) for now
and ill move to varnish in a bit

1- rename file .js
2- COMMENT out interfaces and DELETE :types
4- add to allennlp
5- add a ticked for jonb to move to varnish

thoughts?

jens321:
Sounds perfect, is there a feature in GitHub I can use for (5)? Or should I just put a comment in the code?

jonborchardt:
in general you can add a new issue on the issues in github
like here for allennlp: https://github.com/allenai/allennlp-demo/issues
or here for varnish: https://github.com/allenai/varnish/issues

in this case, i would add the ticket in varnish and assign it to me

jens321:
Done! I kept the Typescript version in a branch of my fork of allennlp-demo, and I also added the code to the issue in the comments. I wasn't able to specifically assign the issue to you @jonborchardt, maybe I don't have the rights to do that? Or am I missing something? 

Link to issue: https://github.com/allenai/varnish/issues/9

jonborchardt:
Looks good

Eric-Wallace:
I think we should close this PR in favor of #219 ?

jens321:
Yeah closing this would make sense I think, #219 is more up to date

PR #193, opened 2019-04-30
Use the JRE instead of JDK.

schmmd:
This cuts out around 300 MB from the Docker image.


matt-gardner:
@pdasigi is in the middle of removing the need for java right now.  Maybe we should just wait, and remove it entirely from the docker image in a few days?

schmmd:
@matt-gardner great! Although I see no point in waiting--we might as well merge forward and then do so again when we remove the JDK.  I tested this change and the wikitables component still works.  Without the JDK or JRE it fails on startup.

PR #198, opened 2019-05-14
skiffify the demo

joelgrus:


.skiff/cloudbuild-deploy.yaml
```diff
@@ -0,0 +1,92 @@
...
+    'rollout', 'status',
+    '-f', 'deployment.json'
+  ]
+  env: [
+    'CLOUDSDK_COMPUTE_ZONE=us-west1',
+    'CLOUDSDK_CONTAINER_CLUSTER=skiff-prod'
+  ]
+  dir: '.skiff'
+substitutions:
+  _ENV: staging
```
schmmd: I assume this was copied and gets overridden if we specify `prod` in Google Cloud Build.

.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,397 @@
...
+//  - The service definition uses them to find the pods it directs traffic to.
+local labels = {
+    app: config.appName,
+    env: env,
+    contact: config.contact
+};
+
+local model_labels(model_name) = labels + {
+    model: model_name,
+    role: 'model-server'
```
schmmd: @codeviking what is this `role` for?  We'd like to add a comment describing it.

.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,397 @@
...
+                        name: config.appName,
+                        image: image,
+                        args: [ '--no-models' ],
+                        readinessProbe: healthCheck,
+                        livenessProbe: healthCheck,
+                        resources: {
+                            requests: {
+                                // Our machines currently have 2 vCPUs, so this
+                                // will allow 4 apps to run per machine
+                                cpu: '0.5',
```
schmmd: There's a small amount of overhead for Kubernetes services--so this won't actually allow 4 apps per machine but rather 3.

schmmd: Maybe change this to `0.4`?

schmmd: Is this for the root application?  Maybe this can just be less...

app.py
```diff
@@ -293,18 +301,32 @@ def info() -> Response:  # pylint: disable=unused-variable
...
+        logger.info(f"setting up default routes for {model_name}")
         app.add_url_rule(f"/{model_name}", view_func=return_page)
         app.add_url_rule(f"/{model_name}/<permalink>", view_func=return_page)
 
+
+    @app.route('/', defaults={ 'path': '' })
     @app.route('/<path:path>')
     def static_proxy(path: str) -> Response: # pylint: disable=unused-variable
-        return send_from_directory(build_dir, path)
+        if os.path.isfile(os.path.join(build_dir, path)):
```
schmmd: Would you comment what the `else` case means?  I don't understand why this change is necessary.

.skiff/cloudbuild-deploy.yaml
codeviking: Yup!

.skiff/webapp.jsonnet
codeviking: The `role` describe the type of job each part of the system does, so that the relevant Kubernetes service can select things of that type.

Would this comment as described be helpful?:

.skiff/webapp.jsonnet
codeviking: The AllenNLP demo has quite a few different applications. I'd recommend this is be set on a case by case basis for y'alls configuration, since it might be different than the standard use case.

That said for the template, we should likely scale it down. I cut https://github.com/allenai/skiff-template/issues/86 to follow up.

app.py
codeviking: There's a comment below, does that not clarify:

```
            # Send the index.html page back to the client as a catch-all, since
            # we're an SPA and JavaScript acts to handle routes the server
            # doesn't.
```

So, the gist is:

1. Serve the file on disk if one exists
2. If the file on disk doesn't exist, return `index.html`, which references JavaScript that then attempts to handle the route

Let me know if that's a better explanation and I can add it.

PR #199, opened 2019-05-14
Update the skiff config to register the new TLD.

codeviking:
This should not be merged until shortly before the domain is
updated to point at the Skiff cluster.

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -58,16 +58,17 @@ local num_replicas = (
 );
 
 local topLevelDomain = '.apps.allenai.org';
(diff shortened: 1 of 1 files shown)


joelgrus:
so here `config.appName` is "allennlp-demo", so this is saying that hosts will be either

["demo.allennlp.org", "allennlp-demo.apps.allenai.org"]

or

["demo.staging.allennlp.org", "allennlp-demo.staging.apps.allenai.org"]

?

is that what we want?

codeviking:
I think that's what's desired.  We'll keep the old url (that which uses `appName`) active, but add the new.

PR #200, opened 2019-05-15
Simplify container model name.

schmmd:


schmmd:
This annoys me whenever I try to get the logs for the model instead of the cloudsql proxy.

PR #208, opened 2019-05-24
Reduce aggressiveness of readiness probe.

schmmd:


schmmd:
This is what we had before: https://github.com/allenai/allennlp-internal/commit/a919bbd032ec3d47d0dacffcd67c284ab541f167#diff-fe7bb2445792a30e9e2963f7c359fc0eL85

Issue #181, opened 2019-03-21
add language model to demo


joelgrus:
motivation: language models are hot. We already have [a demo for one](https://gpt2.apps.allenai.org/?text=Joel%20is) but it's not part of the AllenNLP demo. It would be make our demo more appealing / interesting if we had language modeling as part of it.

Language models are big, so we probably can't do this until after #180 

success criteria: have a language model demo (likely similar to the gtp2-explorer) which can be configured between at least three language models (such as GPT2, BERT, and ELMo--but maybe we'll have better examples from Irvine).  ELMo is a clear requirement.

matt-gardner:
@rloganiv has some language model implementations that we'd like to get in here, also (an AWD-LSTM and an EntityNLM, to start with).  I think there's some work to do to fix up some APIs to make them better integrated with the library first, though.

dnlbreen:
Not sure if this is an appropriate place to post this question, but I've found that question answering is very slow to evaluate with BERT. For answering about 20 questions across around 50 documents of 1000s of words, for example, it takes several hours to evaluate using a cpu. With a GPU, it still takes around 20-30 minutes. Even the smaller BIDAF model on your demo seems to spend a lot of time, around a minute, to evaluate questions on documents with 1000s of words. Do you know why this is? Are deep learning question answering models practical for real time question answering on documents?

schmmd:
This is done. @dnlbreen I don't know how to answer your question--but you're right this probably isn't the best forum as it's focused on the AllenNLP demo.

PR #210, opened 2019-05-29
add language modeling to the allennlp demo

joelgrus:
This is basically a port of the gpt2-explorer (345M model only) into the allennlp demo.

I did my best to change the visual language so that it "looks like" the rest of the allennlp demo.

This is the first model in the demo that's not actually an allennlp `Model`. However, the demo really just wants a `Predictor`, so I implemented a `Predictor` that just wraps the huggingface gpt-2 code. Then there's some if/then logic around whether to instantiate a predictor from a model archive (in the allennlp case), or some other way (in the gpt2 case).

I don't know that this is the best design for getting non-allennlp models into the demo, but it seems a reasonable first step.

The react code for this is less than ideal because it's a hybrid of the existing gpt2-explorer front-end code and the existing allennlp-demo front-end code. (And also because this demo is not input -> submit button -> results, like the other demos are.)

Anyway, if/when we put more non-allennlp models in the demo, we'll probably learn more about the best way to handle this case.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y
 RUN pip install psycopg2-binary
 RUN pip install sentry-sdk==0.7.1
 RUN pip install python-json-logger
+RUN pip install pytorch-pretrained-bert
 
 # Download spacy model
 RUN spacy download en_core_web_sm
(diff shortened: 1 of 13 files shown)


schmmd:
@joelgrus initially we wanted to add additional language models: https://github.com/allenai/allennlp-demo/issues/181.  You argued the other week that it doesn't make sense to use ELMo in this fashion.  Should we do BERT?

joelgrus:
please hold off on reviewing this, I'm going to make some changes

joelgrus:
ok, please review @jonborchardt (since michael is on PTO)

PR #219, opened 2019-07-04
AllenNLP interpret basic

Eric-Wallace:
The barebones for getting interpretations working on textual entailment.


Eric-Wallace:
@matt-gardner this is ready for review, not sure who to tag.

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
                 "responseData": permadata.response_data
         })
 
+    @app.route('/input-reduction/<model_name>', methods=['POST','OPTIONS'])
+    def attack(model_name: str) -> Response:
+        print("REDUCTION!!!\n\n\n")
```
matt-gardner: Remove print.

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
                 "responseData": permadata.response_data
         })
 
+    @app.route('/input-reduction/<model_name>', methods=['POST','OPTIONS'])
```
matt-gardner: I think this should just be an `/attack/` endpoint, with parameters for whatever else is needed.  There shouldn't be any hotflip- or reduction-specific logic in here; that is unsustainable and bad software design.  If you need more parameters, they should be in the route, like you have for `/interpret/`, and each demo can specify in its own API call which inputs to use, etc.

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
...
+        if model is None:
+            raise ServerError("unknown model: {}".format(model_name), status_code=400)
+        max_request_length = app.max_request_lengths[lowered_model_name]
+        data = request.get_json()
+        serialized_request = json.dumps(data)
+        if len(serialized_request) > max_request_length:
+            raise ServerError(f"Max request length exceeded for model {model_name}! " +
+                              f"Max: {max_request_length} Actual: {len(serialized_request)}")
+
+        inputs_to_attack = {"sentiment-analysis":"tokens","machine-comprehension":"question", "textual-entailment":"hypothesis","naqanet-reading-comprehension":"question","named-entity-recognition":"tokens"}
```
matt-gardner: These lines are too long.  But this whole thing should be removed, anyway, as noted above.

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
...
+        inputs_to_attack_map = {"question":"grad_input_2", "passage":"grad_input_1","hypothesis":"grad_input_1","premise":"grad_input_2","tokens":"grad_input_1"}
+        attack = model.attack_from_json(data,inputs_to_attack[lowered_model_name],inputs_to_attack_map[inputs_to_attack[lowered_model_name]])
+        print(attack)
```
matt-gardner: Remove print.

joelgrus:
yes, I can take a look

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
                 "responseData": permadata.response_data
         })
 
+    @app.route('/input-reduction/<model_name>', methods=['POST','OPTIONS'])
+    def attack(model_name: str) -> Response:
+        print("REDUCTION!!!\n\n\n")
+        if request.method == "OPTIONS":
+            return Response(response="", status=200)
+        lowered_model_name = model_name.lower()
+        model = app.attackers.get(lowered_model_name).get("input-reduction")
```
joelgrus: if your first `.get` fails, you'll then be calling `None.get("input-reduction")` and crash, so I don't think this is the right defense against whatever you're trying to defend against

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
...
+        if model is None:
+            raise ServerError("unknown model: {}".format(model_name), status_code=400)
+        max_request_length = app.max_request_lengths[lowered_model_name]
+        data = request.get_json()
+        serialized_request = json.dumps(data)
+        if len(serialized_request) > max_request_length:
+            raise ServerError(f"Max request length exceeded for model {model_name}! " +
+                              f"Max: {max_request_length} Actual: {len(serialized_request)}")
+
+        inputs_to_attack = {"sentiment-analysis":"tokens","machine-comprehension":"question", "textual-entailment":"hypothesis","naqanet-reading-comprehension":"question","named-entity-recognition":"tokens"}
```
joelgrus: these dicts need proper spacing: one space after colons, and one space after commas

app.py
```diff
@@ -170,6 +184,73 @@ def permadata() -> Response:  # pylint: disable=unused-variable
...
+
+    @app.route('/interpret/<model_name>/<interpreter>', methods=['POST', 'OPTIONS'])
+    def interpret(model_name: str, interpreter: str) -> Response:
+        """
+        Interpret prediction of the model
+        """
+        if request.method == "OPTIONS":
+            return Response(response="", status=200)
+        lowered_model_name = model_name.lower()
+        interpreter = interpreter.replace("_","-")
```
joelgrus: what is this defending against? I can think of examples where it would do the wrong thing

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
+import {  
+  AccordionItem,
+  AccordionItemTitle,
+  AccordionItemBody,
+  } from 'react-accessible-accordion';
+import { ColorizedToken } from './Shared';
+
+function postprocessHotflip(org,data) {
```
joelgrus: the rest of the codebase uses ES6, might as well do that here too

```
const postprocessHotflip = (org, data) => {
```

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+  let original_string_colorized = []
+  let new_string_colorized = []  
+  var obj = ''
+  for (let idx=0; idx<org.length; idx++) {
```
joelgrus: spaces around operators (this comment applies everywhere)

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+function postprocessHotflip(org,data) {
+  let original_string_colorized = []
+  let new_string_colorized = []  
+  var obj = ''
+  for (let idx=0; idx<org.length; idx++) {    
+    if (org[idx] !== data[idx]){
+      console.log(org[idx],data[idx])
+      original_string_colorized.push(
+        <ColorizedToken backgroundColor={"#FF5733"}
+        key={idx}>{org[idx]} </ColorizedToken>)
```
joelgrus: two things here:

1. can you format your JSX to make it more readable; e.g.

```
<ColorizedToken ... key={idx}>
  {org[idx]}
</ColorizedToken>
```

2. that said, the way you're using `ColorizedToken` does not feel right, I have no idea what those hex codes even represent or why they're here. A nicer alternative would be to use semantically meaningful subclasses, like

```
const AttackToken = styled.ColorizedToken`background-color: #26BD19`  // light pink (or whatever)
```
and then use those in your JSX. This (a) makes your JSX a lot more readable, and (b) centralizes these definitions so that if you want to change the colors (because when we varnish up the demo you'll want to change these colors) it's easy to do so

demo/src/components/Hotflip.js
joelgrus: also, I don't like this function name, it gives me no clue what the function actually does

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+          <ColorizedToken backgroundColor={"transparent"}
+          key={idx}>{data[idx]} </ColorizedToken>)
+    }    
+  }
+  
+  return [original_string_colorized,new_string_colorized]
+}
+
+
+export default class HotflipComponent extends React.Component {
```
joelgrus: this component has no state? better to write it stateless

```
const Hotflip = ({ hotflipData, hotflipInput, requestDataObject, task }) => ...
```

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+    if (hotflipData === undefined) {
+      new_string_colorized = " ";      
+    }
+    else{          
+      var [first,second] = postprocessHotflip(hotflipData["original"],hotflipData["final"][0])
+      new_string_colorized = second
+      original_string_colorized = first 
+
+      if (task === "sentiment") {
+        let [pos, neg] = hotflipData["new_prediction"]
```
joelgrus: prefer const to let everywhere you can

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+  constructor(props) {
+    super(props)      
+  }
+  
+  render() {
+    const { hotflipData, hotflipInput, requestDataObject, task } = this.props         
+    var original_string_colorized = '';    
+    var new_string_colorized = '';
+    let new_prediction = ''  
+    if (hotflipData === undefined) {
```
joelgrus: all this logic here seems very convoluted. would a separate code path for this case make more sense?

demo/src/components/Hotflip.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+    else {
+      return (<div> <AccordionItem expanded={true}>
+            <AccordionItemTitle>
+              HotFlip Attack
+              <div className="accordion__arrow" role="presentation"/>
+            </AccordionItemTitle>
+            <AccordionItemBody>            
+              <p> <a href="https://arxiv.org/abs/1712.06751" target="_blank" rel="noopener noreferrer">HotFlip</a> flips words in the Hypothesis to change the model's prediction. We iteratively flip the word in the Hypothesis with the highest gradient until the prediction changes.</p>                                
+              {new_string_colorized !== " " ? <p><strong>Original Premise:</strong> {requestDataObject['premise']}</p> : <p></p>}
```
joelgrus: (constant comment) all of this inline logic makes it very hard to understand what's going on here. better to move that logic all into one place somehow

demo/src/components/InputReduction.js
```diff
@@ -0,0 +1,83 @@
...
+
+export default class InputReductionComponent extends React.Component {
+  constructor(props) {
+    super(props)      
+  }
+  
+  render() {
+    const { inputReductionData, reduceInput, requestDataObject } = this.props     
+
+    console.log(inputReductionData);
```
joelgrus: get rid of console.log

demo/src/components/InputReduction.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
```
joelgrus: all the same comments for this file as for the previous component file

demo/src/components/Model.js
```diff
@@ -50,10 +55,60 @@ class Model extends React.Component {
...
+        method: 'POST',
+        headers: {
+          'Accept': 'application/json',
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(inputs)
+      }).then((response) => {
+        return response.json();
+      }).then((json) => {
+        let stateUpdate = Object.assign({}, this.state)
```
joelgrus: cleaner to use spread operator: `const stateUpdate = { ...this.state }`

demo/src/components/Model.js
```diff
@@ -50,10 +55,60 @@ class Model extends React.Component {
       });
     }
 
-    render() {
+    interpretModel(inputs, interpreter) {
+      const { apiUrlInterpret } = this.props
+
+      fetch(apiUrlInterpret(Object.assign(inputs, {interpreter})), {
```
joelgrus: to me the fact that you have to do this suggests that `apiUrlInterpret` has the wrong API)

demo/src/components/Model.js
```diff
@@ -50,10 +55,60 @@ class Model extends React.Component {
...
+        headers: {
+          'Accept': 'application/json',
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(inputs)
+      }).then((response) => {
+        return response.json();
+      }).then((json) => {
+        let stateUpdate = Object.assign({}, this.state)
+        stateUpdate['interpretData'] = Object.assign({}, { [interpreter]: json }, stateUpdate['interpretData'])
```
joelgrus: 1. use the spread operator
2. are you using an array as a key in an object? is that what you mean to do?

joelgrus: in fact, why not just rewrite this all as

```
const stateUpdate = { interpretData: {}, ...state }
stateUpdate.interpretData[interpreter] = json
```

that seems much cleaner and simpler?

demo/src/components/Model.js
```diff
@@ -50,10 +55,60 @@ class Model extends React.Component {
...
+          'Accept': 'application/json',
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(inputs)
+      }).then((response) => {        
+        return response.json();
+      }).then((json) => {        
+        this.setState({hotflipData: json})
+      });
+    }
```
joelgrus: it feels like there's a lot of repeated code in these three methods

demo/src/components/Saliency.js
```diff
@@ -0,0 +1,217 @@
+import React from 'react'
+import colormap from 'colormap'
+import { Tooltip, ColorizedToken } from './Shared';
+import {  
+  AccordionItem,
+  AccordionItemTitle,
+  AccordionItemBody,
+  } from 'react-accessible-accordion';
+
+const getTokenWeightPairs = (input1Grads, input2Grads, input1_tokens, input2_tokens) => {
```
joelgrus: why camelcase for two of these and snake case for two of them?

demo/src/components/Saliency.js
```diff
@@ -0,0 +1,217 @@
...
+      return { token, weight: 1 - weight }
+    })  
+    return [input1TokensWithWeights]
+  }  
+  else{
+    // We do 1 - weight because the colormap is inverted
+    const input1TokensWithWeights = input1_tokens.map((token, idx) => {
+      let weight = input1Grads[idx]
+      return { token, weight: 1 - weight }
+    })
```
joelgrus: you have this code twice, you should try to have it once

demo/src/components/Saliency.js
```diff
@@ -0,0 +1,217 @@
...
+    const {colormapProps} = this.props
+    // colormap package takes minimum of 6 shades 
+    colormapProps.nshades =  Math.min(Math.max(colormapProps.nshades, 6), 72);
+    let colors = colormap(colormapProps)
+
+    let result_string = [];
+    tokensWithWeights.forEach((obj, idx) => {
+      result_string.push(
+        // Again, 1 -, in this case because low extreme is blue and high extreme is red 
+        <label key={idx} data-tip={(1 - obj.weight).toFixed(3)} style={{ display: "inline-block" }} >
```
joelgrus: the fact that you are doing 1 - weight all over the place (multiple times?) makes the code very hard to reason about, is there a way to not have to do that?

joelgrus:
btw, if you do `npm start` it will run eslint and tell you a lot of these things

demo/src/components/Saliency.js
Eric-Wallace: we could have put this logic inside AllenNLP to make the scores all  1 -. The reason we need this is that the colormap visualization library that we are using makes near 0 very dark red and near 1 very transparent. We want the opposite behavior. if someone changed the color library, then they wouldn't need it.

matt-gardner: Could you localize this `1 - weight` to line 67, where you index into the colormap?  It seems like you should only need it once, either where you create the colormap (just invert it once there), or where you access the color (invert your access).

joelgrus: isn't it the case that the code that generates each `tokensWithWeights` is doing 1-weight and then this code is doing it again (i.e. undoing it)?

demo/src/components/Hotflip.js
Eric-Wallace: sorry could you give more details here? tbh I know very little javascript.

demo/src/components/Model.js
Eric-Wallace: sorry, could you provide more detail here?

schmmd:
Textual entailment doesn't work with this PR when you install allennlp@a1476c0414ad1e1412515c53d9e1f27b69039ead.

```
RuntimeError: Error(s) in loading state_dict for WikiTablesMmlSemanticParser:
	Unexpected key(s) in state_dict: "_decoder_step._start_type_predictor.weight", "_decoder_step._start_type_predictor.bias". 
	size mismatch for _entity_type_encoder_embedding.weight: copying a param with shape torch.Size([4, 200]) from checkpoint, the shape in current model is torch.Size([5, 200]).
	size mismatch for _entity_type_decoder_embedding.weight: copying a param with shape torch.Size([4, 100]) from checkpoint, the shape in current model is torch.Size([5, 100]).
```

Eric-Wallace:
note for future: its probably the embedding matrix.

schmmd:
One more note: when used with master the reading comprehension model takes a very significant amount of additional memory (15 GB+) and so it's not practical to use.

matt-gardner:
Superseded by #262.

PR #220, opened 2019-07-05
varnish up the allennlp demo (step 1)

joelgrus:
This project is not done, although it's sort of reached the end of my abilities. Here's what this PR does:

1. updated a lot of node dependencies to a recent version of React etc (required to use varnish)
2. incorporates the varnish `ThemeProvider` into the AllenNLP app
3. removes a bunch of css from App.css (to allow the varnish css to take over)
4. replaces most of the colors in the demo with varnish theme colors (it's possible I chose poorly here, and it won't hurt my feelings if you think so)
5. adds an AI2 banner to the top of the model pane

Some things it does not do:

1. replace all of the external css files with styled components. I took a stab at this, but e.g. the css in menu.css is very complicated and I was not able to migrate it to styled-components without breaking various functionalities
2. get all of the controls (e.g. buttons) using varnish styles. I also took a stab at this, and things broke pretty spectacularly, so I undid it.
3. use the varnish default disappearing logo bar. It played poorly with the two scrolls (menu + model), so I had to settle for just putting the AI2 banner across the model pane
4. update the colors in the hierplane visualization. but someone should do this, because they look pretty bad with the new theme.


schmmd:
DO NOT MERGE THIS ON MONDAY!  Oren didn't want any big changes today as he's demo-ing to some execs at Vulcan.

schmmd:
@jonborchardt would you be able to take this over the finish line?
@joelgrus thanks for all the work you've done so far.

jonborchardt:
Yes.

jonborchardt:
I added changes to add in varnish.

@joelgrus can you please review, remove anything you dont like, and feel free to extend the pattern into the viz components if you want...

i am offering to help in any way, but i think you should to do the actual push when you are set.

feel free to set up time to do an in person review if it helps.

cheers

jonborchardt:
@codeviking please take a look at general varnish use here and add tickets to varnish if you see something odd...

demo/src/components/DemoInput.js
```diff
@@ -72,9 +80,9 @@ class DemoInput extends React.Component {
         this.state = inputState ? {...inputState} : {}
 
         // What happens when you change the example dropdown
-        this.handleExampleChange = e => {
-            if (e.target.value !== "") {
-                const { groupIndex, exampleIndex } = decodeExampleName(e.target.value)
+        this.handleExampleChange = eVal => {
+            if (eVal !== "-1") {
```
codeviking: This looks like a functional change. Just pinging to make sure it's intentional.

demo/src/components/DemoInput.js
```diff
@@ -102,10 +110,17 @@ class DemoInput extends React.Component {
             this.setState(stateUpdate)
         }
 
-        // for radio input, the second param is simply the value
-        this.handleRadioInputChange = fieldName => value => {
+        // Select input selection
+        this.handleSelectChange = fieldName => eVal => {
             let stateUpdate = {}
```
codeviking: I realize it's existing code, but this could be a `const` since it's not re-assigned.

demo/src/components/DemoInput.js
```diff
@@ -102,10 +110,17 @@ class DemoInput extends React.Component {
...
-            stateUpdate[fieldName] = value;
+            stateUpdate[fieldName] = eVal;
+            this.setState(stateUpdate)
+        }
+
+        // Radio input selection
+        this.handleRadioChange = fieldName => e => {
+            let stateUpdate = {}
```
codeviking: Nit, another spot where a `const` could be used over a mutable value.

demo/src/components/DemoInput.js
```diff
@@ -243,49 +259,74 @@ class DemoInput extends React.Component {
...
     }
 }
 
+const BlockRadio = styled(Radio)`
+  display: block;
+`;
+
+const FormInstructions = styled.div`
+  margin: ${({theme}) => `${theme.spacing.md} 0 ${theme.spacing.md}`};
+  -webkit-transition: margin .2s ease;
```
codeviking: Why are we only adding the `-webkit` prefix? Maybe we remove the prefixes all together, and only restore them if / when necessary.  

The [browser support matrix for transition](https://caniuse.com/#search=transition) looks pretty good, so I bet we can drop this.

demo/src/components/DemoInput.js
```diff
@@ -243,49 +259,74 @@ class DemoInput extends React.Component {
...
+  transition: margin .2s ease;
+
+  span {
```
codeviking: I tend to prefer to make these more specific, as the selector currently applies to *any* `<span />` that's a child of the `<FormInstructions />`.  Maybe we do `> span` to make it only immediate children?

The intuition is that the selector by default might style too many things, and that a developer might be confused in the future when adding a new child component that happens to output a `<span />`

That said if this seems unlikely, or if the effect is desired, then keep it as is.

demo/src/components/DemoInput.js
```diff
@@ -243,49 +259,74 @@ class DemoInput extends React.Component {
...
+    padding-right: ${({theme}) => theme.spacing.xs};
+    max-width: 9.375em;
+    color: ${({theme}) => theme.palette.text.secondary};
+  }
+
+  @media screen and (max-height:800px) {
```
codeviking: Do other things share the `max-height: 800px` declaration? If so maybe we need to capture it in the theme somehow? Can Varnish implementors extend the theme (I think they can, right?)?

demo/src/components/Form.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import { TextArea, Input, Select } from '@allenai/varnish/components';
+
+export const FormField = styled.div`
+  margin-top: ${({theme}) => theme.spacing.md};
+
```
codeviking: Extraneous whitespace.

demo/src/components/Form.js
```diff
@@ -0,0 +1,59 @@
...
+  @media (max-height: ${({theme}) => theme.breakpoints.md}) {
+    margin-top: ${({theme}) => theme.spacing.xs};
+  }
+`;
+
+export const FormLabel = styled.label`
+  display: block;
+  font-weight: ${({theme}) => theme.typography.bodyBold.fontWeight};
+  margin-top: ${({theme}) => theme.spacing.xs};
+
```
codeviking: It looks like we're adding a newline above all `transition` declarations. If that's the preference in this codebase, leave as is (disregard my comment above).

demo/src/components/Form.js
```diff
@@ -0,0 +1,59 @@
...
+`;
+
+const baseInputStyles = css`
+  width: 100%;
+  margin-top: ${({theme}) => theme.spacing.xs};
+  display: block;
+
+  transition: min-height .2s ease, opacity .2s ease;
+
+  &:focus{
```
codeviking: Nit, space after `focus`, `&:focus {`

demo/src/components/Form.js
```diff
@@ -0,0 +1,59 @@
...
+    outline: 0;
```
codeviking: A personal, subjective preference -- but I tend to leave the default focus outline in place.

My rationale is that it's there for accessibility purposes, and something users expect to see as a confirmation of focus.  Removing it means it's on us to make sure our focus style is noticeable enough to the user that there's a clear understanding of where focus is -- we accordingly risk breaking the accessibility of our UIs to keyboard only users. 

Ultimately it's up to design, maybe run the question by Nick and see if he feels strongly one way or another.  You can safely ignore my two-cents if it's contrary to REVIZ opinion :).

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
+      </NarrowMenuItem>
+    );
 
-const MenuAccordion = styled(Accordion)`
-  && {
-    background: transparent;
+    return (
+      <OuterGrid>
+        <Logo>
+          <a href="http://www.allennlp.org/" target="_blank" rel="noopener noreferrer">
```
codeviking: I'd drop the `noreferrer` value.  We're linking to our own site, so it's ok (and in fact, desired) to pass referrer information.

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
+          )}
+        </MenuContent>
+      </OuterGrid>
+    );
   }
+}
+
+export const Menu = styled(MenuBase)`
+  height: 100%;
+  position: fixed;
```
codeviking: Did you try `position: sticky` over `fixed`? I find it's my go-to over `fixed` these days. That said it might be the wrong option in this case (I'm not sure how it's being used).

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
+  z-index: 2;
```
codeviking: I like putting the various `z-index` values in a theme or some other single source of truth, as it makes it easy to shift all of them if/when that becomes necessary (which I find it often does), i.e.:

```
ZIndex: {
   /* Values 0-100 reserved for the pane(s) */
   LEFT_PANE: 2,
   /* Values 100-150 reserved for modals */
   MODAL: 100,
}
```

The proposed syntax / values are just a general idea :)

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
-const MenuAccordionArrow = styled.div`
-  && {
-    top: 3px;
-    left: 3px;
-    color: #63a7d4;
-    width: 22px;
-    height: 10px;
+const Logo = styled.div`
+  z-index: 3;
+  padding: ${({theme}) => `1.3125rem ${theme.spacing.xl}`};
```
codeviking: The `rem` value is very specfiic. Maybe comment what this evaluates to / how it was selected?

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
-      }
-    }
+const NarrowMenuItem = styled(AntMenu.Item)`
+  &&& {
+    line-height: 1.4em;
+    height: initial;
+    padding-bottom: 0.375em;
+    padding-top: 0.375em;
+    margin-bottom: -0.125em;
+    margin-top: -0.125em;
```
codeviking: Can we try to avoid negative margin? It's one of my "avoid at all costs" CSS features, since when I use it it tends to be a hack that I'm using to avoid a larger refactor / tweak that's necessary (that ultimately would make the layout more flexible).

demo/src/components/Menu.js
```diff
@@ -1,134 +1,117 @@
...
 
-const MenuAccordionItemBody = styled(AccordionItemBody)`
-  && {
-    animation: none;
-    padding-left: 0.6em;
+const NarrowSubMenu = styled(AntMenu.SubMenu)`
+  &&& {
+    margin-bottom: ${({theme}) => theme.spacing.md};
+
+    .ant-menu-submenu-title {
```
codeviking: How often do we end up with ant class names mixed into our CSS? Seems like it could make upgrading later down the road (i.e. if ANT changes this in a future release, we'll have to hunt for references like these across all the demos that use that version).

demo/src/components/DemoInput.js
jonborchardt: the affect the the same....
elsewhere in the code, we look for -1 instead of ""
just makes it more clear... IMO

codeviking: Cool, that makes sense. Was just making sure it was intentional. 👍

demo/src/components/DemoInput.js
jonborchardt: will fix

demo/src/components/DemoInput.js
jonborchardt: will fix

demo/src/components/DemoInput.js
jonborchardt: yes, i removed some elsewhere, missed this one

demo/src/components/DemoInput.js
jonborchardt: in this case, i was matching exiting behavior... i dont know the code well enough to say it should only be the first span... i leave this to the team

demo/src/components/DemoInput.js
jonborchardt: made consistent using theme

demo/src/components/Form.js
jonborchardt: removed

demo/src/components/Form.js
jonborchardt: removed

demo/src/components/DemoInput.js
codeviking: 👍

demo/src/components/Form.js
jonborchardt: added

demo/src/components/Form.js
jonborchardt: this is from existing style... and nick has not specified a varnish theme yet

demo/src/components/Menu.js
jonborchardt: removed

demo/src/components/Menu.js
jonborchardt: this is a left menu... i dont think sticky makes sense

demo/src/components/Menu.js
jonborchardt: yes, need to add to varnish

jonborchardt: im going to call out of scope for this pr

demo/src/components/Menu.js
jonborchardt: ill leave a comment...
it matches up nicely with the initial layout of the right pane

demo/src/components/Menu.js
jonborchardt: i would love to pick your brain on how to do this without...

demo/src/components/Menu.js
jonborchardt: not that often... but a few times...
i think most of them can be pull into and managed via varnish encapsulation of ant components... ie, we likely want all of our left menus (we only have 1 at this point) to have similar augmentations...

nick needs to design first though

joelgrus:
overall I think it looks very nice, here are my comments:

1. when I do `npm run start`, I get a bunch of eslint errors, we should fix those
2. I don't like that there's no visual distinction between the menu headers e.g. "Answer a Question" and the tasks e.g. "Reading Comprehension". In the current demo they have different sizes and indentations. I don't care if it's that, but I think they should look different.
3. I don't get why the "Other" group is initially collapsed whereas the rest of the groups are initially expanded, I feel like they should be the same
4. with the scroll bar and my default resolution, the menu headers are too long to fit on the screen, "Annotate a sent...". Obviously we can't accommodate arbitrarily long headers, but I feel like we should be wide enough to handle the default ones.
5. I don't like the delay on the select controls, it makes me angry (because I know that the choices are already there and there's no reason to make me wait for them)

jonborchardt:
@joelgrus 
1- most of those warnings are in code that was not changed.  ill try to get to soon, but feel free to clean up if you have time

2- ill speak with nick... but the distinction is a) icon, b) indention, c) dropdown arrow on right.  varnish has not added full menu support yes, so you are getting menus without the full nick style... some day we will swap it out for a varnish menu

3- you can set other to always be open if you like...  its up to you (i just took a guess)

4- i agree, but it looks fine on my machine... you can make it wider if need be... can you send a screen shot?

5- you can kill the delay via css if you like.  varnish has not yet added full select support... so there has been no discussion of staying with our dropping the default ant delay. (it seems fast enough for me... maybe show me what you are seeing)

joelgrus:
ok, I think everything is dealt with (other than the pause in the select control, which I couldn't figure out)

PR #224, opened 2019-07-16
Allow minor / revision updates for Varnish.

codeviking:
This modifies the dependency fix I just made so that minor / revision
updates are automatically installed from Varnish if the appropriate
update / upgrade command is used (I can't remember the exact command,
and I think it's different between yarn and npm).

Sorry for the noise @joelgrus.

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@allenai/varnish": "0.3.16",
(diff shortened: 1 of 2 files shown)


jonborchardt:
you really should be updating the css file reference also... but it wont matter for this fix

codeviking:
Ah, shucks. I failed to catch that.  I'll post an update for each repo shortly.

jonborchardt:
It will not matter in this case... But would be nice to keep in sync

PR #225, opened 2019-07-16
No yarn

codeviking:
Remove all `yarn` references.


jonborchardt:
Why?

codeviking:
@jonborchardt It's bad practice to have both.  In fact, every time one executes `yarn` when there's a lockfile triggers a warning.

The build script uses `npm` and not `yarn`, so this way users will use the same package manager as what's deployed. This helps ensure things are consistent, since `yarn` and `npm` have slightly different semantics around dependency resolution.  Given the desire is to achieve reproducible builds, varying package management solutions is counter-intuitive to that :).

jonborchardt:
I would recommend removing npm then 😀

codeviking:
@jonborchardt Haha, yea, I prefer `yarn` too.  That said since the build mechanism and docs default to `npm`, this was the path of least friction.  If we decide to update to `yarn` down the road I'm all about that and will help with the update :).

jonborchardt:
Ok then 🤢

PR #230, opened 2019-07-16
Jonb/add icons

jonborchardt:
image


demo/src/components/Menu.js
```diff
@@ -116,3 +116,16 @@ const WrappingLink = styled(InternalLink)`
     white-space: pre-wrap;
     word-wrap: break-word;
 `;
+
+// TODO: this will eventually be exported from varnish... but needed here now
+class SvgIcon extends React.PureComponent {
+  IconImg = styled.img`
+    vertical-align: initial;
+  `;
```
codeviking: This only works in select scenarios (i.e. the parent has to have the right `display` type).

If you're trying to center this, I don't think it's the concern of the icon, but rather the parent layout. I'd propose something like:

```
<FlexRow vAlign={true}>
   <Icon />
   <OtherStuff />
</FlexRow>
```

In other words, I don't know that we want to capture this behavior in the child -- rather, we want to throw it in some sort of parent for greater flexibility, etc.

jonborchardt: the issue is, that the ant parent does a funny thing and adds a negative em to vertical-align...

jonborchardt:
i dont have access to this team city.... @joelgrus can you help me get access or see the logs?

joelgrus:
there should be a "log in as guest" button

but in this case your problem is

```
[19:30:50][Step 1/2] ./src/components/Menu.js
[19:30:50][Step 1/2]   Line 128:  'IconImg' is not defined  react/jsx-no-undef
```

demo/src/components/Menu.js
```diff
@@ -116,3 +116,16 @@ const WrappingLink = styled(InternalLink)`
     white-space: pre-wrap;
     word-wrap: break-word;
 `;
+
+// TODO: this will eventually be exported from varnish... but needed here now
+class SvgIcon extends React.PureComponent {
+  IconImg = styled.img`
```
joelgrus: const IconImg

jonborchardt: fixed

PR #236, opened 2019-07-23
Install from Github, so it works.

codeviking:


codeviking:
We may want to hold off, as the `WikiTablesMmlSemanticParser` model appears to fail with the latest n' greatest.

PR #237, opened 2019-07-24
Fix #235; Display something nice if there's no output.

### LINKED ISSUE #235: Page goes blank when posting an incomplete sentence
https://demo.allennlp.org/semantic-role-labeling/OTUwMjc2

1.  Go to https://demo.allennlp.org
2.  Select "Semantic Role Labeling"
3.  Enter "Jhon store data into the file." as the sentence

The page goes blank instead of showing an error.

image

codeviking:
Hierplane throws up a nasty stack trace when there's no output
returned from the backend. This handles that scenario cleanly by
rendering a message asking the user to alter their input.

### DIFF
--- a/demo/src/components/demos/SemanticRoleLabeling.js
+++ b/demo/src/components/demos/SemanticRoleLabeling.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { ExternalLink } from  '@allenai/varnish/components';
 import { withRouter } from 'react-router-dom';
+import styled from 'styled-components';
 
 import { API_ROOT } from '../../api-config';
 import Model from '../Model'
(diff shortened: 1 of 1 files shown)


codeviking:
Here's how things look:

image

schmmd:
Looks good. So this wasn't a regression from the recent changes? Sorry--perhaps I shouldn't have bothered you about this.

I hope you got to learn something about this demo at least.

codeviking:
> Looks good. So this wasn't a regression from the recent changes? Sorry--perhaps I shouldn't have bothered you about this.

I don't think so.  That said it's always possible. We did update Hierplane to a new version of React, which in theory could've introduced the bug (though the associated code in Hierplane didn't change w/ the update, so it seems unlikely).

> I hope you got to learn something about this demo at least.

Definitely. I got my local environment setup and in triaging setup issues learned a lot about the mechanics :).

codeviking:
Also, I'm not sure what protocol is for PRs.  If the fix seems a-ok and I should merge, release and verify let me know and I'll get that rolling.

schmmd:
@codeviking I will merge / release / verify.  You 🚀

schmmd:
Just curious--why not try / catch?

codeviking:
> Just curious--why not try / catch?

We'd have to use an `<ErrorBoundary />` since the error occurs when it's rendered (we just talked about this in person).  We'll add this to the template.

Issue #241, opened 2019-07-24
Demo needs a cache breaker


schmmd:
The demo currently caches static assets, forcing the user to do a hard refresh to see new content.  Instead we should have a cache breaker similar to the one in https://github.com/allenai/skiff-template.

Peace.  Michael

PR #244, opened 2019-07-26
Set the cache headers so we don't cache HTML.

schmmd:
Fixes https://github.com/allenai/allennlp-demo/issues/241


codeviking:
LGTM!

PR #248, opened 2019-07-29
Resolve correctly with trailing slash.

schmmd:


app.py
```diff
@@ -343,6 +332,14 @@ def static_css_proxy(path: str) -> Response: # pylint: disable=unused-variable
     def static_media_proxy(path: str) -> Response: # pylint: disable=unused-variable
         return send_from_directory(os.path.join(build_dir, 'static/media'), path)
 
+    @app.route('/<path:path>')
+    def static_proxy(path: str) -> Response: # pylint: disable=unused-variable
+        # Send the index.html page back to the client as a catch-all, since
+        # we're an SPA and JavaScript acts to handle routes the server
+        # doesn't.
+        return send_file(os.path.join(build_dir, 'index.html'))
```
codeviking: This this still handle `/` correctly? I think the `@app.route('/', defaults={ 'path': '' })` bit was required to do so. If you've verified that it works fine then I'm misremembering.

The removal of this bit:

```
        if os.path.isfile(os.path.join(build_dir, path)):	
            return send_from_directory(build_dir, path)
```

Worries me slightly. That was meant to serve resources that exist on disk if they're requested. Have you made sure serving static assets (i.e. images, like the logo n' such) still work dorrectly?

schmmd: Yes--the existing routes for static assets cover them all.  I ran in an incognito window with printlns and we never hit that other path.

codeviking: Great!

schmmd:
I don't know either.  If you look earlier though--index.html is served using the syntax I add here.  Before index.html was being served with different code with non-slash routes.

app.py
schmmd: `/` was handled correctly earlier in the file.  That line was redundant.

schmmd:
```
    @app.route('/')
    def index() -> Response: # pylint: disable=unused-variable
        return send_file(os.path.join(build_dir, 'index.html'))
```

codeviking: Ah, nice!

I fear I may have introduced the madness, as I tried to help resolve some issues with static asset serving awhile back.

Thanks for cleaning this up!

codeviking:
🤘

PR #250, opened 2019-07-29
Release several fixups

schmmd:
* Don't let browsers cache HTML pages.
* Allow trailing slashes when directly navigating to URLs.
* Remove sentry.
* Improve stackdriver logs by organizing pods by model.


schmmd:
Not sure if we should CR releases since we already CR changes to master (staging).

brendan-ai2:
Yeah, I don't think we should code review twice. Are there any release specific checks you'd like me to perform? Otherwise, happy to stamp this LGTM.

PR #251, opened 2019-07-30
updating varnish to get new footer

jonborchardt:
also fixed a few lingering issues with last varnish install

### DIFF
--- a/demo/public/index.html
+++ b/demo/public/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/hierplane@0.0.9/dist/static/hierplane.min.css">
-    <link rel=stylesheet href="https://cdn.jsdelivr.net/npm/@allenai/varnish@0.3.16/theme.css" />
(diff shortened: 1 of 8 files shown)


jonborchardt:
image

demo/public/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/hierplane@0.0.9/dist/static/hierplane.min.css">
-    <link rel=stylesheet href="https://cdn.jsdelivr.net/npm/@allenai/varnish@0.3.16/theme.css" />
+    <link rel=stylesheet href="https://cdn.jsdelivr.net/npm/@allenai/varnish@0.3.19/theme.css" />
```
codeviking: What if we wrote JS in Varnish that automatically adds this to the `document.head`?  This way users wouldn't have to add this manually, and it'd stay in sync w/ the version without any user effort.

We'd want to make sure there's not an existing one already in the head, of course, and provide a `boolean` that lets the user disable this functionality.  Nothing to do now, but maybe queue this up to Varnish if you like the idea?

PR #268, opened 2019-09-24
UI Fixes for the Demo

Eric-Wallace:
- [x] Hotflip for LM still has G' (replaced it)
- [x] Hotflip for LM isn't doing the equality check correctly, so it changes every single word even though the prediction has changed (seems to work fine, was a one-off problem )
- [x] NER is out of order still (reverse fixed)
- [x] NER input reduction still not reducing all the way (its fine, the beam was is just too small sometimes)
- [x] Switch BiDAF back to GloVe version for speed? (added a third option)
- [x] When the input is modified for GPT-2 and BERT, the interpretation either shows the old one, or it says loading forever. When the input is modified, there needs to be some sort reset. (set loading to false)
- [x] Add an "Adversarial Attacks" header similar to "Saliency Maps" (changed Saliency maps to interpretations)

### DIFF
--- a/demo/src/components/Hotflip.js
+++ b/demo/src/components/Hotflip.js
@@ -142,7 +142,7 @@ export default class HotflipComponent extends React.Component {
     const target = targeted === undefined ?
       " "
     :
-      <p> Change prediction to (leave blank to allow any change): <FormInput type="text" onChange={ this.updateTargetWord }/> </p>
+      <p> Leave blank to allow any change (untargeted). For targeted attacks, enter a single token that you want to flip the mask to: <FormInput type="text" onChange={ this.updateTargetWord }/> </p>
 
     const buttonDisplay = (flippedString !== " " && targeted === undefined) ?
       " "
--- a/demo/src/components/Hotflip.js
+++ b/demo/src/components/Hotflip.js
@@ -122,6 +122,9 @@ export default class HotflipComponent extends React.Component {
     }
     // data is available, display the results of Hotflip
     else {
+        if (this.state.loading) { // loading is done
+            this.setState({ loading: false });
+        }
         [originalString, flippedString] = colorizeTokensForHotflipUI(hotflipData["original"],
                                                                      hotflipData["final"][0])
         newPrediction = hotflipData["new_prediction"]
--- a/demo/src/components/demos/LanguageModel.js
+++ b/demo/src/components/demos/LanguageModel.js
@@ -38,7 +38,6 @@ const Wrapper = styled.div`
 
 const ModelArea = styled.div`
   background: ${({theme}) => theme.palette.common.white};
-  min-height: 30em;
 `
 
 const Loading = styled.div`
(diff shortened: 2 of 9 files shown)


matt-gardner:
On the second point, I just ran hotflip for LM locally, and it flipped 3 words.  It looks like it's working to me.

Eric-Wallace:
Last thing for @matt-gardner for a future PR is the disabling of input reduction for naqanet.

demo/src/components/Hotflip.js
```diff
@@ -122,6 +122,9 @@ export default class HotflipComponent extends React.Component {
     }
     // data is available, display the results of Hotflip
     else {
+        if (this.state.loading) { // loading is done
+            this.setState({ loading: false });
+        }
```
joelgrus: this whole construct feels really strange to me, I feel like whatever code is responsible for fetching the hotflip data should also be responsible for setting the loading the flag (analogous to how the predictions themselves work)

demo/src/components/demos/LanguageModel.js
```diff
@@ -178,6 +178,20 @@ const getGradData = ({ grad_input_1 }) => {
   return [grad_input_1];
 }
 
+const cleanTokensForDisplay = (tokens) => {
+  console.log(tokens)
```
joelgrus: remove console.log

joelgrus: also it would be good to have some comments in this code, in a month no one will remember why it's doing what it does

demo/src/components/Hotflip.js
joelgrus: but I'm not familiar enough with the interpret stuff to know where that code is

matt-gardner: So, I wanted to do that, but I didn't know how to pass that state back down to where it needed to be.  The function that does the fetching lives in `Model.js`, while the piece that needs the loading state in way down here, in `Hotflip.js`.  How do I have `Model.js` set state on `Hotflip.js`?  If you can tell me how to do that, I will fix this.

demo/src/components/demos/LanguageModel.js
matt-gardner: I'll fix this now.

demo/src/components/Hotflip.js
joelgrus: I believe the standard way to do that would be to have that state live in `Model.js` and get passed down as a prop to `Hotflip.js`

matt-gardner: And then, because of how we've designed our models, I have to modify every single demo to pass this through, instead of just modifying one file to add a somewhat hacky but functional state update.  That was why I did it this way.

matt-gardner: Maybe it would have been better to have a function on `Model.js` that adds all of the interpret components, and the concrete models just call that function with whatever specific state they need.  Then I can pass the state through just once, inside of `Model.js`, where it makes sense.  That probably would have been cleaner.  I'll open an issue to fix this.

PR #269, opened 2019-09-25
undo reverse, moving to allennlp

Eric-Wallace:


Eric-Wallace:
@matt-gardner Once we merged the latest allennlp PR, we need to add the new commit SHA here

PR #271, opened 2019-09-26
Add more code highlighting in README

bryant1410:


README.md
```diff
@@ -8,9 +8,9 @@ To run the demo locally for development, you will need to:
 
 1. Create a fresh environment:
 
-    ```
-    conda create -n allennlp-demo python=3.6
-    source activate allennlp-demo
+    ```bash
+    conda create -n allennlp-demo python=3.7
+    conda activate allennlp-demo
```
matt-gardner: @schmmd, I think you said something recently about using `source` vs. `conda` here - do you have an opinion on this change?

bryant1410: (just as a reference: https://stackoverflow.com/a/49600711/1165181)

schmmd: What I said only applied to Slurm jobs.  `conda` is preferred in general, but it has some weird integrations with your environment so it doesn't work when running a job.

This change seems great to me!

schmmd:
Thanks @bryant1410 !

README.md
bryant1410: Oh, yeah, I think you're referring to conda/conda#7980. [You can also do the following](https://github.com/conda/conda/issues/7980#issuecomment-524154596):

```bash
eval "$(conda shell.bash hook)"
conda activate my_env
```

PR #274, opened 2019-10-02
Add usage information to Sentiment Analysis

schmmd:
@Eric-Wallace could you additionally add instructions for evaluating and training?  It would look like this: https://github.com/allenai/allennlp-demo/blob/master/demo/src/components/demos/ReadingComprehension.js#L580

Feel free to push commits directly to this PR.

### DIFF
--- a/demo/package-lock.json
+++ b/demo/package-lock.json
@@ -1340,7 +1340,7 @@
     "address": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/address/-/address-1.0.3.tgz",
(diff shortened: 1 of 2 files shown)


Eric-Wallace:
i don't have write access, so I can't push here =(

schmmd:
@Eric-Wallace just sent you an invite to the AllenNLP team in GitHub, which should fix the problem.

Eric-Wallace:
great, I also added this https://github.com/allenai/allennlp/pull/3320 for the training config.

Eric-Wallace:
@schmmd think this PR is good to go.

PR #276, opened 2019-10-02
add more documentation

joelgrus:


README.md
```diff
@@ -102,20 +98,20 @@ Here is a [pull request](https://github.com/allenai/allennlp-demo/commit/149d068
...
     ```
 
 Make sure `text_classifier` matches the name from your AllenNLP predictor. In our case, the predictor class should have `@Predictor.register('text_classifier')` at the top.
 
 3. In `app.py` consider adding a log of your model's outputs. Search for `log_blob` in the `predict` function for an example of how to do this.
 
-4. The backend is now set up. Now let's create the front end for your model. Add your model under its associated category in the `modelGroups` object in `demo/src/models.js`. 
+4. The backend is now set up. Now let's create the front end for your model. Add your model under its associated category in the `modelGroups` object in `demo/src/models.js`.
 ```js
 {model: "sentiment-analysis", name: "Sentiment Analysis", component: SentimentAnalysis}
```
schmmd: It took me a bit to realize that `SentimentAnalysis` is that task specific component that we're re-using here.

README.md
```diff
@@ -125,6 +121,32 @@ Also make sure to import your component at the top of the file.
 
 You can find more information about the front end creation at the bottom of the README.
 
+## Contributing a non-AllenNLP Model
+
+After some recent changes, it should be possible to add non-AllenNLP components to the demo
```
schmmd: It should be possible or is it possible?  I would change "non-AllenNLP components" to emphasize that you might use this for AllenNLP components from forks or past versions of AllenNLP, which seems more likely than a non-AllenNLP component.

README.md
```diff
@@ -125,6 +121,32 @@ Also make sure to import your component at the top of the file.
...
+    }
+```
+
+and finally modify `demo/src/models.js` to add its entry to the menu:
+
+```
+    {"model": "my-new-model", "name": "Descriptive Name"}
+```
+
+(you don't have to give a `component` for it).
```
schmmd: Why don't you need to give a `component` for it?  I think because it serves itself rather than XYZ, but I would spell this out.

schmmd:
Also, please have @brendan-ai2 review before merging.

brendan-ai2:
There's a few things I think we need (apologies if I missed them or they're documented elsewhere):

1. Operational information
  - What does the demo run on? (Skiff?)
  - How do we release it?
  - How do we check its status?
  - Are there any alerts/logs?
2. Is the information for how to run via Docker up to date? I thought there were now multiple images that would run for each individual model.

Thanks!

joelgrus:
added info along these lines, please take a look.

the demo actually runs fine on a single machine (other than the memory pressure), the skiff version does a few things to "distribute" it:

(1) there is a dedicated "demo front-end" container that doesn't load any models
(2) the ingress controller routes e.g. `/predict/machine-comprehension` to a dedicated "machine comprehension" container
(3) that container loads only the machine comprehension model

so that you end up with a bunch of containers, each running the same (full) app but with only one model loaded, and with the k8s ingress routing requests accordingly. without skiff / k8s, this all just happens on a single machine.

README.md
```diff
@@ -68,21 +87,37 @@ docker run -p 8000:8000 -v $HOME/.allennlp:/root/.allennlp --rm allennlp/demo:$G
...
+
+To deploy the production demo, merge this branch into the `release` branch.
+
+## Monitoring
 
+You can access the demo logs through [Marina](https://marina.apps.allenai.org/a/allennlp-demo),
+which links to GCP, or by visiting GCP directly. The two that are most frequently useful are
 
+* Cloud Build - History (where you can see if / why your builds failed)
+* Kubernetes Engine - Workloads (where you can find the various containers making up the service)
```
brendan-ai2: Links would make this even more useful. Presumably https://console.cloud.google.com/cloud-build/builds?project=ai2-allennlp? But I don't see the demo at https://console.cloud.google.com/kubernetes/workload?project=ai2-allennlp&workload_list_tablesize=50...

brendan-ai2:
Thanks for the updates! :) If you could add those links, I'd appreciate it. LGTM

PR #277, opened 2019-10-03
Fix gpt2 link

brendan-ai2:
- Frontend link only


brendan-ai2:
This works for me locally, but my frontend experience is very limited, so I don't know if it's the "right way". Took me a bit too as I started by modifying `app.py` which I now think isn't necessary. @Eric-Wallace, do you rely on linking to the backend or just the frontend? I.e., are you hitting us programmatically or just linking the site itself?

joelgrus:
I would prefer a less one-off solution. something along the lines of creating `redirects.js` that looks like

```
{"gpt2": "next_model_lm"}
```

and then you could load it in in `App.js`, and in the top level `Demo` component check it and remap the model if necessary. or something similar like that, does that make sense?

brendan-ai2:
Yep, makes sense. I've done something similar by including a `redirects` field in `demo/src/models.js`. Given that external users have noticed I don't think we need a perfect solution here, but let me know if I've done something very stupid.

demo/src/App.js
```diff
@@ -47,12 +47,14 @@ const App = () => (
 // and it renders the specific task in an iframe.
 const Demo = (props) => {
   const { model } = props.match.params
+  const redirectedModel = modelRedirects[model] ? modelRedirects[model] : model
```
joelgrus: simpler: `modelRedirects[model] || model`

demo/src/App.js
```diff
@@ -47,12 +47,14 @@ const App = () => (
...
+  props.match.params.model = redirectedModel
```
joelgrus: modifying `props` like this seems bad, instead I'd do (see next comment)

demo/src/App.js
```diff
@@ -47,12 +47,14 @@ const App = () => (
...
 
   return (
     <React.Fragment>
       <Header alwaysVisible={true} />
       <div className="pane-container">
-        <Menu selectedModel={model} clearData={() => {}}/>
+        <Menu selectedModel={redirectedModel} clearData={() => {}}/>
         <SingleTaskFrame {...props}/>
```
joelgrus: `<SingleTaskFrame {...props, model: redirectedModel}/>

brendan-ai2: This doesn't seem to work.

```
./src/App.js
  Line 57:  Parsing error: Unexpected token, expected "}"

  55 |       <div className="pane-container">
  56 |         <Menu selectedModel={redirectedModel} clearData={() => {}}/>
> 57 |         <SingleTaskFrame { ...props, model: redirectedModel }/>
     |                                    ^
  58 |       </div>
  59 |     </React.Fragment>
  60 |   )
```

brendan-ai2: `model` isn't a parameter to `SingleTaskFrame`. I tried adding it but that didn't seem to help. Pushed the broken code so you can see what I tried.

joelgrus: oh, I see, then we'll have to tweak SingleTaskFrame to do

```
  const { slug } = props.match.params
  const { model } = props
```

this is not ideal, but I believe having a component modify its props is a react antipattern, cf https://stackoverflow.com/questions/24939623/can-i-update-a-components-props-in-react-js

brendan-ai2:
Thanks for the help and review!

Issue #281, opened 2019-11-03
Misaligned micro-header logo


aaronsarnat:
![image](https://user-images.githubusercontent.com/8367927/68081616-1b9ee800-fdce-11e9-82be-13b6c0589614.png)

aaronsarnat:
Assigning to @jonborchardt. FYI @Pelagic-sky.

jonborchardt:
Yes, allennlp does not have up to date varnish.  Should be fixed.

PR #283, opened 2019-11-12
Updating to Varnish 0.6.0

aaronsarnat:
**This PR:**

- Updates to Varnish 0.6.0 (latest version as of this writing)
- Adds link to models list in README (this was a pain point for me when I was trying to run locally)
- Fixes minor header logo layout issues (and fixes https://github.com/allenai/allennlp-demo/issues/281)
- Fixes [favicon resource size error](https://user-images.githubusercontent.com/8367927/68702010-d3ee2e00-053c-11ea-997c-025d2ed302e2.png) in `manifest.json`

---

Note that it appears this repo uses `npm` and not `yarn`, so I updated dependencies and committed a new `package-lock.json` file in lieu of a yarn.lock.

---

**In-Browser Screenshot:**

image


aaronsarnat:
@schmmd pinging you for FYI purposes

demo/src/App.js
```diff
@@ -57,6 +59,7 @@ const App = () => (
             <Route path="/:model/:slug?" component={Demo}/>
           </Switch>
         </BlockOverflow>
+      </DefaultLayoutProvider>
```
codeviking: AllenNLP has it's own code for accomplishing the multi-pane layout that it uses.  It pre-dates both PRIOR and the addition of those code to Varnish.

We could choose to remove a bunch of code from AllenNLP and refactor it to use the Varnish version. This would:

* Reduce the bundle size, as we won't include source code for the same thing (but implemented differently) twice
* Canonicalize the implementation of the pane based layout

That said, the multi-pane layout in Varnish has a fairly egregious bug:
https://github.com/allenai/varnish/issues/200

So, for the time being I'd leave this as is, but I'd put a `TODO` here:

```
TODO: Use Varnish's multi-pane layout, rather than our home rolled one.
```

I'd also cut a ticket in Varnish to record taking care of this.  Please reference https://github.com/allenai/varnish/issues/200 in it's description, i.e:

> Update AllenNLP demo to use Varnish's layout mechanism for multi-pane layouts. Blocked until #200 is fixed.

aaronsarnat: Issue for refactoring allennlp-demo w/ multi-pane: https://github.com/allenai/allennlp-demo/issues/284

matt-gardner: One thing to be careful of: Joel recently added frames to our layout, so that we can have the UI itself served by separate underlying demos.  I'm not really familiar with how any of this works, but it seemed relevant to this issue.  I haven't followed this completely, so if you're already aware of this, or if it's not actually a problem, then great, you can ignore this comment.

aaronsarnat: Thanks for the context @matt-gardner. Looking at [Joel's iframe structural change](https://github.com/allenai/allennlp-demo/commit/105556318c3967f9ff08ff18438035b8c205815d#diff-0631aa0ae3c5d498f46de044a4bfd16b), that part is happening at a deeper nesting level of the component tree. The layout change I made is higher up in the wrapper part of the tree that shouldn't affect any of the routing.

I also tested the UI by clicking around to different demos and all the routing and per-model iframe injection appears to work correctly.

aaronsarnat:
Merging on green

demo/src/App.js
matt-gardner: Great, as long as you're aware of it.  I thought it might affect the refactoring that you mentioned in #284, but I don't know enough to know whether it would or not.

aaronsarnat: That's a good point. I'm not sure if it will or not, but I'll add a note to that issue to reference the iframe stuff so we can make sure we don't miss that detail.

codeviking: @matt-gardner Great callout.  It might.

We should definitely be careful to *not* remove that mechanism.  Our layout controls should wrap the `<iframe />`, and the `<iframe />` should derive it's dimensions from our layout mechanism.  If this isn't the case, then AllenNLP should stick with it's current implementation and *not* use the Varnish based mechanism.

PR #293, opened 2020-01-21
Add tooltip IDs where relevant so they aren't duplicated.

meslater1030:


jonborchardt:
this seems fine
and i applaud the initiative...
but a stonger fix would be to move to antd tooltip: https://ant.design/components/tooltip/

PR #294, opened 2020-01-21
Upgrade to Varnish 0.8.6

meslater1030:
This PR:

* Upgrades to Varnish 0.8.6
* Changes the usage of impacted components

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@allenai/varnish": "^0.6.0",
+    "@allenai/varnish": "0.8.6",
     "ajv": "^6.10.0",
     "antd": "^3.23.2",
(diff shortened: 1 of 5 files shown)


demo/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@allenai/varnish": "^0.6.0",
+    "@allenai/varnish": "0.8.6",
```
meslater1030: I tend to pin versions so that we can do a quick visual check when upgrading and also in this case so we can remember to upgrade the CSS when upgrading Varnish. Happy to use the `^` though if that's more consistent with how things are done in ReViz.

jonborchardt: agreed

meslater1030:
Thanks @jonborchardt yes - I looked at all the pages/loaded all the models and nothing sticks out to me as obviously wrong.

PR #295, opened 2020-01-23
Cleanup various React issues

meslater1030:
This PR:

* Updates deprecated React Lifecycle Hooks
* Removes any state updates within render methods
* Changes anchor tags to buttons where we utilize `onClick` rather than href as recommended by the a11y linter
* Adds keys to rendered lists of React elements
* Removes unused imports
* Addresses other console warnings and errors as appropriate

### DIFF
--- a/demo/src/App.js
+++ b/demo/src/App.js
@@ -113,7 +113,7 @@ class SingleTaskDemo extends React.Component {
   // We also need to update the state whenever we receive new props from React router.
   componentDidUpdate({ match }) {
     const { model, slug } = match.params;
-    if (model !== this.state.selectedModel && slug !== this.state.slug) {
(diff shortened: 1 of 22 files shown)


demo/src/App.js
```diff
@@ -111,9 +111,11 @@ class SingleTaskDemo extends React.Component {
   }
 
   // We also need to update the state whenever we receive new props from React router.
-  componentWillReceiveProps({ match }) {
+  componentDidUpdate({ match }) {
     const { model, slug } = match.params;
-    this.setState({selectedModel: model, slug: slug});
+    if (model !== this.state.selectedModel && slug !== this.state.slug) {
```
jonborchardt: should this be an ||

PR #310, opened 2020-02-06
Give labels to all the sets of accordion menus

meslater1030:
This PR:

Adds labels to all the sets of accordion menus so that it doesn't look weird to have random spaces between categories.

Screen Shot 2020-02-06 at 3 58 24 PM


demo/src/components/demos/LanguageModel.js
```diff
@@ -219,7 +219,7 @@ const Attacks = ({attackData, attackModel, requestData}) => {
     hotflipData["final"][0] = cleanTokensForDisplay(hotflipData["final"][0]);
   }
   return (
-    <OutputField>
+    <OutputField label="Model Attacks">
```
meslater1030: If there is a better name for this, please enlighten me.

PR #311, opened 2020-02-12
Crude mechanism for using different `allennlp` versions.

brendan-ai2:
- For @ZhaofengWu's new coref model.
- Separate Dockerfiles, tests, and `requirements.txt`.
- Updates cloudbuild config to create coref images.
- Despite this all the server code (frontend, backend and models) is packaged in each Docker image. In addition to being wasteful in terms of space, this seems somewhat brittle. In short, if we accidentally loaded models from incompatible versions, we would crash. I don't think this happens now, but I could see that bug developing. Some kind of disentangling here is advisable.


matt-gardner:
cc @nitishgupta

brendan-ai2:
@matt-gardner, @nitishgupta this isn't ready for review yet. I hit some issues with deploying this to an ad-hoc environment, but Sam helped get that resolved. I'll ping you later today assuming things go smoothly from here.

brendan-ai2:
This is working now. One of our suggested docs for coreference is producing some strange results, but @ZhaofengWu is going to look into that a bit more. See https://allennlp-demo.dv-5.apps.allenai.org/coreference-resolution/MTQxNjA2MQ==. I'm unsure of whether we should wait to release until after that's addressed. We could leave it in as a failure case.

schmmd:
@brendan-ai2 my suggestion is to remove the example from the drop down and ship it.  I'm glad @ZhaofengWu is investigating separately.

brendan-ai2:
Thanks for the review, @schmmd. @ZhaofengWu identified that the issue appears to be related to tokenization. When sentences are separated by multiple spaces we can reproduce this bug. I suspect a Spacy update since our last deploy could have triggered this. For now I'm just editing the sentence and Zhaofeng will cut an issue to fix this.

PR #312, opened 2020-02-12
Set the canonical host only for prod environments.

codeviking:
This fixes an issue where Skiff ad-hoc environments weren't working because `demo.$env.allennlp.org` wouldn't resolve to the Skiff cluster. When this occurred this prevented a TLS certificate from being issued.

I confirmed this by live `kubectl edit`ing @brendan-ai2's `dv-4` ad-hoc.  Removing the `allennlp.org` entries resolved the issue.

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -58,18 +58,16 @@ local num_replicas = (
 );
 
 local topLevelDomain = '.apps.allenai.org';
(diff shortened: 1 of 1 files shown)


brendan-ai2:
The tests are currently failing because I've reconfigured them to build and test an additional Dockerfile that only exists in my other PR. I'm going to bypass them and merge.

PR #313, opened 2020-02-14
Update coref model description

ZhaofengWu:


demo/src/components/demos/Coref.js
```diff
@@ -31,7 +31,7 @@ const description = (
     </span>
     <ExternalLink href = "http://cemantix.org/data/ontonotes.html" target="_blank" rel="noopener">{' '} the Ontonotes 5.0 dataset {' '}</ExternalLink>
     <span>
-    in early 2017.
+    in early 2017. We substitute the GloVe embeddings that it uses with BERT embeddings.
```
brendan-ai2: Just to make things a bit more explicit and flow better I'd modify your addition to:

```
The model here is based on that paper, but we have substituted the GloVe embeddings that it uses with BERT embeddings. On Ontonotes that gives an F1 score of $SCORE on test.
```

Naturally replacing $SCORE.

PR #322, opened 2020-02-25
Jon/header work

jonborchardt:
small tweaks to
1- get header to hide when scrolling
2- fix initial scroll position when clicking task to task
3- upgraded varnish (while here)
4- remove unused code
5- made logo a bit larger to match other demos


demo/src/components/ScrollToTop.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const ScrollToTop = () => {
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, [pathname]);
+
+  return null;
+}
```
codeviking: Why not use https://github.com/allenai/varnish/blob/master/components/ScrollToTopOnPageChange.tsx?

jonborchardt: because i forgot it was there!

jonborchardt: fixed

PR #323, opened 2020-02-25
Add RoBERTa SST model

ZhaofengWu:
Sorry for the many changes that remove trialing whitespaces -- my editor does that automatically.

### DIFF
--- a/demo/src/components/demos/SentimentAnalysis.js
+++ b/demo/src/components/demos/SentimentAnalysis.js
@@ -31,7 +31,16 @@ const NAME_OF_GRAD_INPUT = "grad_input_1"
 
 // Text shown in the UI
 const description = (
-  <span> Sentiment Analysis predicts whether an input is positive or negative. The model is a simple LSTM using GloVe embeddings that is trained on the binary classification setting of the <a href="https://nlp.stanford.edu/sentiment/treebank.html">Stanford Sentiment Treebank</a>. It achieves about 87% accuracy on the test set.</span>
+  <span>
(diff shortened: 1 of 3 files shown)


schmmd:
I added Jon to look at the react.

ZhaofengWu:
Sounds good. I was basing this off of QA which does similar things for NAQANET.

ZhaofengWu:
Hi @jonborchardt , mind taking a quick look?

demo/src/components/demos/SentimentAnalysis.js
```diff
@@ -86,12 +127,20 @@ const Attacks = ({attackData, attackModel, requestData}) => {
...
+
   const [positiveClassProbability, negativeClassProbability] = responseData['probs']
   const prediction = negativeClassProbability < positiveClassProbability ? "Positive" : "Negative"
 
   let t = requestData;
   const tokens = t['sentence'].split(' '); // this model expects space-separated inputs
 
+  // The RoBERTa-large model is very slow to be attacked
+  const attacks = model && model.includes('RoBERTa') ?
+    " "
```
jonborchardt: this can likely be null instead of " "

ZhaofengWu: I copied this from what we are doing for QA. https://github.com/allenai/allennlp-demo/blob/7ab9258d611efb78cc15a995aac5984ab810cf68/demo/src/components/demos/ReadingComprehension.js#L197-L200 I'm not familiar with nulls in JS/React (undefined?) and I think it's nice to have the repo to use the same style. Since it doesn't look like a big problem I will leave this as is for now.

PR #324, opened 2020-02-26
Update model descriptions.

schmmd:


schmmd:
@matt-gardner do you have a description for the masked-lm model I can use?  The description there is pretty bare presently--it'd be good to describe the model that is used.

demo/src/components/demos/NamedEntityRecognition.js
```diff
@@ -34,8 +34,8 @@ const description = (
   <span>
     <span>
         The named entity recognition model identifies named entities
-        (people, locations, organizations, and miscellaneous)
-        in the input text. This model is the "baseline" model described in
+        (people, locations, organizations, etc.)
```
matt-gardner: "miscellaneous" is actually a label in the data (see https://demo.allennlp.org/named-entity-recognition/MTQ2NTgzMw==).  There are four labels.  So what was here originally is actually better.

matt-gardner: Well, it describes the `elmo-ner` model accurately.  There's a good argument that the above description should describe a task, leaving the model description for somewhere else.  Then I would change the whole sentence to "Named entity recognition is the task of identifying different kinds of named entities in a given text, such as people, locations, or organizations."

schmmd: Yes--I modified it because we have two NER models and the instructions were specific to one.  Your suggested text is an improvement.

schmmd: I'm sticking with my original text.  To do something better we need a larger refactor.

PR #331, opened 2020-03-16
Fix remaining character when user cleared input of language model

yoquankara:
Previous code did not update component state when user cleared all
text input, probably to avoid backend request.
However, it resulted in a UI bug where the last character could
not be deleted as user might expect.


yoquankara:
Reproduction for the issue this PR is fixing:

1. In "Language Model" or "Masked Language Model", use backspace key to delete all input text in "Sentence" form.
2. The last character can be deleted.

Expected behavior:
User can delete all previous text and input a new sentence.

Issue #339, opened 2020-04-05
Saliency Maps for the RoBERTa model in Sentiment Analysis Are not using WordPiece Tokenization


IsThatYou:
The gradients returned by [interpret method](https://github.com/allenai/allennlp-demo/blob/master/app.py#L403) contains special tokens like [CLS] and [SEP]. Moreover, it splits words into subwords. However, in the current [code](https://github.com/allenai/allennlp-demo/blob/d4929b14422d0db17b192becab5235c6b0be6e97/demo/src/components/demos/SentimentAnalysis.js#L136), It splits words by space. This is definitely creating wrong/misaligned saliency maps. 

This may become a recurring issue going forward as more Transformer models are added. 

@ZhaofengWu

matt-gardner:
Thanks for pointing this out @IsThatYou!  A PR to fix this is definitely welcome; we're a bit swamped with other things right now, so it'll take us a while to get to this.

IsThatYou:
Currently, my temporary fix involves both changes in allennlp and allennlp-demo. Specifically, the dictionary returned by predict_json also contains the sentenced tokenized into wordpiece. Then I set `this.state.tokens =  json['tokens']`. This way on https://github.com/allenai/allennlp-demo/blob/d4929b14422d0db17b192becab5235c6b0be6e97/demo/src/components/demos/SentimentAnalysis.js#L136, we can comment out this line, and just use the tokens passed instead. Would this be helpful? I think one thing that I can think of to make this generalizable is to have the tokens displayed in interpretations to be returned from predict_json only. Instead of borrowing from requestData.

matt-gardner:
Yes, if the predictor doesn't currently return the tokenized sentence, that's a problem, it needs to be added.  And then the UI should just use the tokenization returned by the predictor.  I agree with this solution.  Care to make whatever PRs are appropriate to fix this?

IsThatYou:
Sure, I can do that. Should I also change the respective code in allennlp for sentiment analysis? Currently, I add the tokens in the `make_output_human_readable` method. This is kind of specific for each model type.

matt-gardner:
Yeah, can you make a PR to allennlp also, to get the tokens passed through?  I'm not sure whether the best way is to do it in the model or the dataset reader, but we can figure that out once there's a PR.  Definitely start the PR with what you're doing now, and I'll look at it and see what we think.

IsThatYou:
sounds good, I will @ you once I create a PR there.

[...]

PR #342, opened 2020-04-07
now the sentiment analysis interpret uses tokens returned by predict_json, rather than requested data

IsThatYou:
PR created in reference to #339. This is just a very simple fix so that the interpret code uses tokens returned by `predict_json` rather than the requested data. This way special tokenization (like wordpiece) can work properly for saliency maps. 

As mentioned, this would require changes in allennlp too. Do you want me to do that as well? If so, should I change the `make_output_human_readable` method or something else? @matt-gardner

### DIFF
--- a/demo/src/components/Model.js
+++ b/demo/src/components/Model.js
@@ -54,7 +54,7 @@ class Model extends React.Component {
         this.props.updateData(inputs, json)
         //
         // requestData, responseData
-
+        this.setState({tokens: json['tokens']});
         if (window.frameElement) {
           // Based on http://www.awongcm.io/blog/2018/11/25/using-iframes-api-to-toggle-client-side-routing-of-react-router-for-legacy-web-apps/
           window.frameElement.ownerDocument.defaultView.history.pushState({}, '', newPath)
--- a/demo/src/components/Model.js
+++ b/demo/src/components/Model.js
@@ -54,7 +54,7 @@ class Model extends React.Component {
         this.props.updateData(inputs, json)
         //
         // requestData, responseData
-        this.setState({tokens: json['tokens']});
+
         if (window.frameElement) {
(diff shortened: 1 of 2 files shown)


demo/src/components/Model.js
```diff
@@ -54,7 +54,7 @@ class Model extends React.Component {
         this.props.updateData(inputs, json)
         //
         // requestData, responseData
-
+        this.setState({tokens: json['tokens']});
```
matt-gardner, line 57: Can't you just grab the 'tokens' key from the response data below?  I don't think we should need to modify `Model.js`. Similar to what's done here: https://github.com/allenai/allennlp-demo/blob/abb6e30695cedb9209cd4d262b0db7b24a5d4100/demo/src/components/demos/TextualEntailment.js#L225-L226

And given the current state of the backend, we should make it so this fallbacks gracefully to the current tokenization if the proper key isn't present.

IsThatYou: That's a good point. I have changed the code to what you suggested.

demo/src/components/demos/SentimentAnalysis.js
```diff
@@ -131,10 +131,7 @@ const Output = ({ responseData, requestData, interpretData, interpretModel, atta
 
   const [positiveClassProbability, negativeClassProbability] = responseData['probs']
   const prediction = negativeClassProbability < positiveClassProbability ? "Positive" : "Negative"
-
-  let t = requestData;
-  const tokens = t['sentence'].split(' '); // this model expects space-separated inputs
-
+  const tokens = responseData['tokens'];
```
matt-gardner, line 134:
```suggestion
  const tokens = responseData['tokens'] || requestData['sentence'].split(' ');
```

matt-gardner: Can you try this change and see if it works?  We want the code to not crash in the case that `responseData['tokens']` is not present.  Giving bad output is better than crashing.

IsThatYou: yeah, this works. I just committed your code.

PR #343, opened 2020-04-07
Update module descriptions

nitishgupta:


demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
```diff
@@ -41,7 +41,7 @@ const allModules = [
   new ModuleInfo(
     'find',
     'find(Q) → P',
-    'For text spans in the question, find similar text spans in the passage.',
+    'For text span in the question, find similar text spans in the passage.',
```
schmmd, line 44: You need an article here.  Do you mean "the text span"?

demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
```diff
@@ -52,17 +52,17 @@ const allModules = [
   new ModuleInfo(
     'relocate',
     'relocate(Q, P) → P',
-    'Find spans in the passage related to an argument in the question.',
+    'Based on the question, select the relevant argument for the input passage spans.',
```
schmmd, line 55: This is not clearer than the original to me.  I still don't understand.

demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
```diff
@@ -81,23 +81,23 @@ const allModules = [
...
     'find-max-num',
     'find-max-num(P) → P',
-    'Select the text span in the passage with the largest number.',
+    'Output the span associated to the largest number amongst the input spans.',
   ),
   new ModuleInfo(
     'find-min-num',
     'find-min-num(P) → P',
-    'Select the text span in the passage with the smallest number.',
+    'Output the span associated to the smallest number amongst the input spans.',
```
schmmd, line 100: This wording is a bit awkward.

"Output the span associated with the smallest number from the input spans."

What are the input spans?

nitishgupta: How about "Amongst the input spans, select the span associated to the largest number"

schmmd: That seems good, but to read more naturally in American grammar I would prefer "Among the input spans, select the span associated with the largest number."

Here are two references from a quick Google that confirmed my intuitions:

https://english.stackexchange.com/questions/78403/acceptable-uses-for-associated-with-or-associated-to
https://www.grammarly.com/blog/amongst-among/

nitishgupta: Sure, I'll update. There are a few other places where we've used "associated to" which I'll change to "associated with"

demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
nitishgupta: "Based on the question, select the relevant argument spans associated with the input passage spans." -- is this better? Trying a phrasing similar to `filter` module.

demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
nitishgupta: "For the selected text span in the question, find similar text spans in the passage." -- is this better?

schmmd: Perfect!

demo/src/components/demos/reading-comprehension/nmn/ModuleInfo.js
nitishgupta: Given a question "Who scored the first goal?", the relocate gets as input 
Q = "Who scored"
P = all goal spans
The output is the spans that highlights the respective players who scored the input goals.
We're calling the players as the "Who scored" argument of the "goals"
Does this example help?

schmmd: That's super helpful.  Does any of the following work?

"Select relevant passage spans which may answer the question."
"Select passage spans which may answer the question."
"Select passage spans which are relevant to answering the question."

nitishgupta: I'm afraid no, since this module is not always the high level model that answers the final questions. 
Example -- "How old is LA Lakers' coach?" -- in this case `relocate` would be used as
`relocate("coach", find("LA Lakers"))` -- where the relocate outputs the "coach" argument for the input passage span. Is that clearer?

schmmd: I'm not sure what verbiage is best ;-(

nitishgupta: I'll check with Matt and update

nitishgupta: Matt says we can go with `Based on the question, select the relevant argument for the input passage spans.` for now. We will most likely be updating all very soon with more detail anywway. I'll send a PR with the changes.

PR #348, opened 2020-04-08
Shorten one of the examples

matt-gardner:
The example was too long to fit inside the DROP NMN, so this shortens it.


matt-gardner:
Just going to merge this, as it's a one-line change to an example string.

PR #349, opened 2020-04-09
Move to a new SNLI model

dirkgr:
This should fix the demo.


dirkgr:
Works at https://allennlp-demo.fix-elmo-snli.apps.allenai.org.

PR #352, opened 2020-04-20
Update usage for the AllenNLP demos to use 1.0.0rc1

schmmd:


schmmd:
This was a bit tedious, and clearly this is a rough approach long-term as it includes a significant amount of redundant code.  Also--I'm starting to think I should have a separate section on how to install AllenNLP, because over time the usage examples could diverge on the AllenNLP version.  @dirkgr what do you think on this point?

demo/src/components/demos/ConstituencyParser.js
```diff
@@ -61,25 +61,35 @@ const examples = [
 
 const apiUrl = () => `${API_ROOT}/predict/constituency-parsing`
 
+const modelUrl = 'https://storage.googleapis.com/allennlp-public-models/elmo-constituency-parser-2020.02.10.tar.gz'
+
+const bashCommand =
+    `pip install --pre allennlp allennlp-models
+echo '{"sentence": "If I bring 10 dollars tomorrow, can you buy me lunch?"}' | \\
+allennlp predict --include-package allennlp_models ${modelUrl} -`
```
dirkgr, line 69: You should not need `--include-package allennlp_models` anymore.

dirkgr: Actually, in rc1, you still need that. So maybe we keep it until rc2.

demo/src/components/demos/ConstituencyParser.js
```diff
@@ -61,25 +61,35 @@ const examples = [
 
 const apiUrl = () => `${API_ROOT}/predict/constituency-parsing`
 
+const modelUrl = 'https://storage.googleapis.com/allennlp-public-models/elmo-constituency-parser-2020.02.10.tar.gz'
+
+const bashCommand =
+    `pip install --pre allennlp allennlp-models
```
schmmd, line 67: I should probably factor this out into a section about installing AllenNLP for all the examples.  What do you think?

dirkgr: If it gets its own section, it needs to be more involved than `pip install --pre allennlp allennlp-models`. We could have a little explanation that includes how to clone and `pip install -e .`, so users can hack on the code themselves.

If it's just that one `pip` command, let's leave it here.

schmmd: I might split it out in a separate PR because it applies to both sections (the bash section and the python section).

demo/src/components/demos/ConstituencyParser.js
schmmd: I hate having these URLs copied and pasted all over too.  We need to address it, but I'm not sure it makes sense to address before we support usage for all our models.  Presently when we have a task with multiple models we have usage for one of them.

Ideally this data would be connected to `models.json` and could pull them out from there.  *maybe* we can do this now and perhaps I'll look at it briefly today.

schmmd:
@dirkgr merging since you made two comments which I read to be optional.

PR #368, opened 2020-04-23
Cache images during PR build

schmmd:
Uncached the build takes around 7 minutes.  Cached the build takes around 3 minutes.

The savings is modest, but this doesn't cost anything since storage on DockerHub is free, so I figured we'd merge it in for now.


epwalsh:
This is interesting: https://pythonspeed.com/articles/faster-multi-stage-builds/

We may want to consider doing a multi-stage build. But if we stick with normal builds, we probably don't need to use this third-party action. Sounds like we could just `docker pull` before building.

schmmd:
I think the last failure was spurious (npm error).

I don't think multi-stage builds are worth the complexity for `allennlp-demo`.  Doing a pull before the build is exactly what Skiff does to get some caching for Docker builds.  It sometimes works, but doesn't work if the previous build was dissimilar to the present build--which can often happen on pull requests that diverge from master.

I'm not entirely sure what this plugin is doing, but it's pushing intermediate stages to our DockerHub and using them to cache the next build.  I think the proof is in the pudding in that the builds to speed up by 2x.  If we switched to doing a pull before the build, we would probably need to add an action for master builds to push artifacts and pull those.  I would also bet that our caching wouldn't be quite as effective.

I think that (assuming the tests pass) we should merge this in, and maybe experiment with something else down the line.

PR #371, opened 2020-04-25
Don't serve the UI with Python.

codeviking:
This change sets up a dedicated, NGINX container for serving the UI.
NGINX is a great candidate for this, as the files are static and can
be served from disk without executing code on the server.

This also allows us to remove all code from the Python application
related to serving UI files, which makes things simpler.

I tested this locally via the revized instructions for running things
via `docker`, and will deploy an ad-hoc shortly to confirm the changes
in full.

I changed the Python application's root URL to return some information
about the AllenNLP version and the models it's configured to serve.
This information isn't leaking anything, as the repository is already
public, so it felt like a safe change. I can remove this if we feel
otherwise.

@schmmd I'm assigning you, but please feel free to reassign
to whomever is best.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,8 @@
 # This Dockerfile is used to serve the AllenNLP demo.
-
 FROM allennlp/commit:adeb1b1278619ff2d74d4fd82825e50a36f95ff4
 LABEL maintainer="allennlp-contact@allenai.org"
 
 WORKDIR /stage/allennlp
-
-# Install npm early so layer is cached when mucking with the demo
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y nodejs
 
 # Install python dependencies
 COPY requirements.txt requirements.txt
@@ -17,10 +13,6 @@ RUN spacy download en_core_web_sm
 
 COPY scripts/ scripts/
 COPY server/models.py server/models.py
-
-# Now install and build the demo
-COPY demo/ demo/
-RUN ./scripts/build_demo.py
 
 COPY tests/ tests/
 COPY app.py app.py
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ docker run -p 8000:8000 \
 ```
 
 You can change `reading-comprehension` to the name of the model [here](https://github.com/allenai/allennlp-demo/blob/master/models.json) that you'd like to run. You probably shouldn't try to
-run all of them, as it's quite slow.
+run all of them, as they will take a long time to download and require a significant
+amount of memory.
 
 Note that the `run` process may get killed prematurely if there is insufficient memory allocated to Docker. As of September 14, 2018, setting a memory limit of 10GB was sufficient to run the demo. See [Docker Docs](https://docs.docker.com/docker-for-mac/#advanced) for more on setting memory allocation preferences.
 
--- a/README.md
+++ b/README.md
@@ -70,9 +70,11 @@ To run the demo locally for development, you will need to:
 
 ## Running with Docker
 
-Here is an example for how to manually build the Docker image and run the demo on port 8000.
-As above, you probably don't want to load every model,
-so you should specify the specific models you want in your `docker run` command.
+We use `docker` to package the application and serve it using Kubernetes. Below are instructions
+for running things locally in a fashion that's close to how we do things in production, which
+can be useful for troubleshooting weird issues.
+
+**First, start up the backend / API for a model like so:**
 
 ```bash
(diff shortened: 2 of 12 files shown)


Dockerfile
```diff
@@ -1,13 +1,9 @@
 # This Dockerfile is used to serve the AllenNLP demo.
-
 FROM allennlp/commit:adeb1b1278619ff2d74d4fd82825e50a36f95ff4
 LABEL maintainer="allennlp-contact@allenai.org"
 
 WORKDIR /stage/allennlp
 
-# Install npm early so layer is cached when mucking with the demo
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y nodejs
```
codeviking, line 9: This has a nice side-effect of getting us off of Node 8, which was EOL'd a long time ago and likely has some nice security issues :).

dirkgr:
That's a lot of reviewers.

schmmd:
@dirkgr it'd be good if you or @epwalsh reviewed these PRs--but you don't both need to.  Looks like I'm assigned so I'll take a look too--on Monday.

codeviking:
@dirkgr @schmmd Sorry for the wide distribution, I just didn't know who to refer this to. Please feel free to prune, and I'll follow your lead on future changes.

schmmd:
@epwalsh it'd be great to have you clued in on these.  I added you as an assignee.

README.md
```diff
@@ -82,13 +84,25 @@ docker run -p 8000:8000 \
            -v $HOME/.allennlp:/root/.allennlp \
            --rm \
            allennlp/demo:$GIT_HASH \
-           --model model1 --model model2
+           --model reading-comprehension
 ```
 
-Where you see "model1" and "model2" examples above, you would use actual model names which are listed as JSON object key names [here](https://github.com/allenai/allennlp-demo/blob/master/models.json).
+Youc an change `reading-comprehension` to the name of the model [here](https://github.com/allenai/allennlp-demo/blob/master/models.json) that you'd like to run. You probably shouldn't try to
```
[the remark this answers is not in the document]
codeviking: Good catch, thanks! The S2 team used to remark on how they think I was the source of 99% of the typos in their documentation :).

codeviking:
Cool, I got it working in a deployed environment via an ad-hoc. It'll only be around for the next hour, but I can spin it back up after the weekend for people to play with if they'd like. That can be found [here](https://allennlp-demo.ohplz.apps.allenai.org/reading-comprehension/MTc2MzU5Mg==).

README.md
```diff
@@ -82,13 +84,25 @@ docker run -p 8000:8000 \
            -v $HOME/.allennlp:/root/.allennlp \
            --rm \
            allennlp/demo:$GIT_HASH \
-           --model model1 --model model2
+           --model reading-comprehension
```
schmmd, line 87: Nice improvement

README.md
```diff
@@ -82,13 +84,25 @@ docker run -p 8000:8000 \
...
 ```
 
-Where you see "model1" and "model2" examples above, you would use actual model names which are listed as JSON object key names [here](https://github.com/allenai/allennlp-demo/blob/master/models.json).
+You can change `reading-comprehension` to the name of the model [here](https://github.com/allenai/allennlp-demo/blob/master/models.json) that you'd like to run. You probably shouldn't try to
+run all of them, as it's quite slow.
```
schmmd, line 91: They will take a long time to download and a significant amount of memory.

README.md
```diff
@@ -82,13 +84,25 @@ docker run -p 8000:8000 \
...
 
 Note that the `run` process may get killed prematurely if there is insufficient memory allocated to Docker. As of September 14, 2018, setting a memory limit of 10GB was sufficient to run the demo. See [Docker Docs](https://docs.docker.com/docker-for-mac/#advanced) for more on setting memory allocation preferences.
 
+**Next, in a separate terminal, start up the UI:**
+
+```bash
+export GIT_HASH=`git log -1 --pretty=format:"%H"`
+docker build -t allennlp/demo-ui:$GIT_HASH .
+docker run -p 8080:80 --rm allennlp/demo-ui:$GIT_HASH
```
schmmd, line 100: The `GIT_HASH` was meant for https://demo.allennlp.org/info but it's been broken.  It's less needed now that we have Skiff so we could remove it if it would simplify things.  It's been broken so long we probably should.

README.md
codeviking: Updated, thanks.

README.md
codeviking: Gotcha, will do.

codeviking: Done.

PR #376, opened 2020-04-28
Update nmn-drop docker image

nitishgupta:
Fixed some issues with nmn-drop demo. @matt-gardner @codeviking


matt-gardner:
@epwalsh, this looks like it's failing because of the secrets on PRs problem.  Am I diagnosing that right?  And if so, what's the fix?

epwalsh:
@matt-gardner I think you're right. I just tried a fix here: https://github.com/allenai/allennlp-demo/pull/377

matt-gardner:
Thanks @epwalsh!

PR #380, opened 2020-04-29
Remove DB config from the UI server.

codeviking:
The UI is now served by NGINX and doesn't access the database, this
PR removes the database related config from the deployment, making
it a bit simpler and following the principle of least priviledge.

I also removed the `GIT_SHA` environment variable as that's been
unused for some time.

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -287,13 +287,6 @@ local db_env_variables = [
     }
 ];
 
-local env_variables = db_env_variables + [
-    {
-        name: 'GIT_SHA',
-        value: sha
-    }
-];
-
 local deployment = {
     apiVersion: 'extensions/v1beta1',
     kind: 'Deployment',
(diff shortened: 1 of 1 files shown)


codeviking:
I'm deploying an ad-hoc now to verify that this works.

codeviking:
Looks like it's working fine:
https://allennlp-demo.nodb.apps.allenai.org/reading-comprehension/MTc4NDY4Mw==

.skiff/webapp.jsonnet
```diff
@@ -328,12 +321,9 @@ local deployment = {
                                 cpu: '50m',
                                 memory: '100Mi'
                             }
-                        },
-                        env: env_variables
-                    },
-                    cloudsql_proxy_container
```
schmmd, line 334: Can you also delete where `cloudsql_proxy_container` is defined?

codeviking: Nope. We need that elsewhere.  The API servers still need (and use) that.

schmmd: Gotcha, thanks!  For some reason I was thinking the UI managed this.

PR #381, opened 2020-04-29
Kick off the new paradigm for serving models.

codeviking:
This includes the code for a small Flask application that serves
predictions using the bidaf model. The application uses some common code
for request logs and configuration that individual applications
shouldn't have to worry about.

I use `docker` to package the resulting application and a bit of 
`PYTHONPATH` munging so that code can be imported from the 
`allennlp_demo` module. This feels a little clunky, and raises some 
interesting questions around what happens if an individual model 
relies on something that doesn't play nicely with `common`/ -- 
say, for instance, an old version of Python.  That said I think in 
this scenario things will be fine, as the user would simply modify 
their `Dockerfile` to not include the `common` module and instead 
replace that code with their own, modified version.

Nothing in the demo package uses this code yet, this is meant
to just get everyone aligned with respect to how I plan on building
out individual endpoints.

An outstanding question is around testing -- I'd like to write
integration tests that verify the entire ensemble, but these would likely
take a long time to run given that each endpoint has to download an
archive at startup.


allennlp_demo/common/logs.py
```diff
@@ -0,0 +1,78 @@
...
+    # Output a request log our own with information we're interested in.
+    @app.after_request
+    def log_request(r: Response) -> Response:
+        t = time.perf_counter() - g.start
+        rl = RequestLogEntry(r.status_code, request.method, request.path, request.args,
+                             request.remote_addr, request.headers.get('X-Forwarded-For'), t)
+        extra = { "request": asdict(rl) }
+        logging.getLogger("request").info("", extra=extra)
+        return r
+
```
codeviking, line 78: I chose to fork from the approach used in `skiff-template` a bit here, and instead wrote code for outputting a request log that we have full control over.

The template's approach makes sure the NGINX logs are in JSON, but doesn't modify those for Flask. That's because Flask uses Werkzeug under the hood, and there's no documented way to change the request log format. Here we implement our own so that the logs from our API server can be easily searched against, parsed, etc.  This is fine for the template, but less so for this effort since I don't anticipate running a reverse proxy per endpoint (though we may end up doing so, we'll see).

allennlp_demo/common/config.py
```diff
@@ -0,0 +1,19 @@
+import json
+
+from dataclasses import dataclass
```
codeviking, line 3: I find myself confused, as I just ran into a bug in another project where `dataclasses` wasn't working, and it's because the project used `python:3.6` and dataclasses weren't added until `3.7` (I think).

I'm going to dig into this a bit more, as I'm stumped.

codeviking: Ah, I see:

``
[I] ~/h/a/a/bidaf ❯❯❯ pip show dataclasses                                                                 (bidaf)
Name: dataclasses
Version: 0.7
Summary: A backport of the dataclasses module for Python 3.6
Home-page: https://github.com/ericvsmith/dataclasses
Author: Eric V. Smith
Author-email: eric@python.org
License: Apache
Location: /Users/sams/local/opt/miniconda3/envs/bidaf/lib/python3.6/site-packages
Requires:
Required-by: transformers, allennlp
```

So `allennlp` provides it, and I just happened to have used it.  That said I noticed AllenNLP's docs encourage users to use Python `3.7`, so I'm going to upgrade this to use that too.

File: allennlp_demo/bidaf/model.json [hunk 3]
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "bidaf",
+    "archive_file": "https://storage.googleapis.com/allennlp-public-models/bidaf-model-2020.03.19.tar.gz",
+    "predictor_name": "reading-comprehension",
+    "max_request_length": 311108
```
schmmd, line 5: I don't remember what this is for btw...

allennlp_demo/bidaf/requirements.txt
```diff
@@ -0,0 +1,66 @@
+allennlp==1.0.0rc3
```
schmmd, line 1: this is a frighteningly long list.  Surely we don't need all these?  Are you listing all transitive dependencies?  I recommend *not* doing this as it will make it more likely things get messed up.  For example, whata if we upgrade allennlp to the next version of pytorch but someone doesn't update pytorch here?

allennlp_demo/common/config.py
```diff
@@ -0,0 +1,19 @@
...
+
+@dataclass(frozen=True)
+class Model:
+    """
+    Class capturing the options we support per model.
+    """
+    id: str
+    archive_file: str
+    predictor_name: str
+    max_request_length: int
```
schmmd, line 13: maybe add a comment about what this is, as it's not immediately obvious

allennlp_demo/bidaf/requirements.txt
codeviking: I got this like so:

```
pip install allennlp_models==1.0.0rc3 flask
pip freeze -l > requirements.txt
```

Maybe I should just put `allennlp_models` and `flask` in this, and let the rest of implicit. It seems less deterministic, but if that's preferable for some reason let me know.

allennlp_demo/bidaf/model.json
codeviking: There's [code in the original demo](https://github.com/allenai/allennlp-demo/blob/master/app.py#L258) which rejects a request above a certain max length.

Thanks for reminding me about this, as I should figure out how I'm going to honor this.

That said I'm not sure how useful it is. The original code buffers the entire request into memory and serializes it before enforcing the limit, which is expensive. I think it'd be preferable to enforce this in a reverse proxy, which can do it a much cheaper fashion.

It looks like the original change was made while investigating some OOM issues:
https://github.com/allenai/allennlp-demo/pull/168.  That said the PR also suggests that it's only a possible fix -- maybe we drop this and see if issues come back?

allennlp_demo/bidaf/requirements.txt
schmmd: You're right that it's less deterministic, but I think it's preferable in that it clearly communicates what's needed and what can be upgraded.  But let's let @epwalsh weigh in--I'm happy to go with his recommendation.

allennlp_demo/bidaf/model.json
codeviking: We chatted, sounds like it did solve the OOM issue, and we should have a resonable limit.

I'm going to use [the Ingress Controller](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size) to do it.

allennlp_demo/common/config.py
codeviking: It's gone now, per the our conversation about using NGINX to do it.

allennlp_demo/bidaf/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM python:3.7.7-stretch
+LABEL maintainer="allennlp-contact@allenai.org"
+
+WORKDIR /allennlp_demo
+COPY bidaf/requirements.txt bidaf/requirements.txt
+RUN pip install -r bidaf/requirements.txt
+
+RUN spacy download en_core_web_sm
+
+COPY . .
```
matt-gardner, line 13: When there are more models in here, this will copy all of the different demo code, instead of just the current model and the common components.  Is that really what you want?

allennlp_demo/bidaf/api.py
```diff
@@ -0,0 +1,39 @@
...
+    # For simplicity, we use Flask's built in server. This isn't recommended, per:
+    # https://flask.palletsprojects.com/en/1.1.x/tutorial/deploy/#run-with-a-production-server
+    #
+    # That said we think this is preferable because:
+    #   - It's simple. No need to install another WSGI server and add logic for enabling it in
+    #     the right context.
+    #   - Our workload is CPU bound, so event loop based WSGI servers don't get us much.
+    #   - We use Kubernetes to scale horizontally, and run an NGINX proxy at the front-door, which
+    #     adds the resiliency and other things we need for production.
+    app.run(host="0.0.0.0", port=8000)
```
matt-gardner, line 39: Pretty sure the [attack and interpret endpoints] need to be moved here too.

allennlp_demo/bidaf/Dockerfile
codeviking: Nope, good catch. Thanks.

allennlp_demo/bidaf/api.py
codeviking: I don' think [bidaf](https://github.com/allenai/allennlp-demo/blob/master/app.py#L50) has those endpoints. I will need to do so for other models.

matt-gardner: It does: https://github.com/allenai/allennlp-demo/blob/6b42a00a07c6dd0bfcded05186db074cd66d743e/app.py#L57-L58.

(And sorry, looks like I forgot to actually put in the link that I meant to in my earlier comment.)

codeviking: Yikes. I was looking for `bidaf` but should've been looking for `reading-comprehension`.

Thanks.

allennlp_demo/bidaf/requirements.txt
epwalsh: I think this is okay. We could also use something like [pipenv](https://pipenv.pypa.io/en/latest/) which gives you the benefits of both approaches: you specify only direct dependencies in the `Pipfile` and then a  `Pipfile.lock` is generated which ensures builds are deterministic unless you explicitly update.

In my experience though it's annoyingly slow and doesn't always work how you'd expect. So it's probably not worth it right now.

allennlp_demo/bidaf/api.py
codeviking: I'm going to port these over in a separate PR coming shortly, after getting a single "end to end" endpoint in place.

PR #383, opened 2020-04-30
Add K8s config for the new bidaf endpoint.

codeviking:
This wires up the K8s config required for the bidaf endpoint and
demonstrates how I plan to share code across endpoints.

The `.skiff/bidaf.jsonnet` file is pretty minimal -- to the extent
one might ask if we could omit it all together. I'd like to keep it
as to encourage a clear entry point for configuring each endpoint, so
that if / when customizations are necessary it's clear how to do so.


codeviking:
I verified that this works by setting up a trigger and deploying things:

```
~ ❯❯❯ curl -s https://demo.allennlp.org/api/bidaf | jq
{
  "allennlp": "1.0.0rc3",
  "archive_file": "https://storage.googleapis.com/allennlp-public-models/bidaf-model-2020.03.19.tar.gz",
  "id": "bidaf",
  "predictor_name": "reading-comprehension"
}
```

codeviking:
Here's what an endpoint will look like in Marina:
https://marina.apps.allenai.org/a/allennlp-demo-mapi-bidaf

PR #384, opened 2020-04-30
Cache predictions in the new endpoint.

codeviking:
This adds code for serving cached predictions to the new endpoint.


allennlp_demo/bidaf/api.py
```diff
@@ -23,9 +27,29 @@
     def info() -> str:
         return jsonify({ **asdict(model), "allennlp": VERSION })
 
+    @lru_cache(maxsize=128)
```
epwalsh, line 30: I think we could make this an order of magnitude higher at least, right?

codeviking: Maybe. I was trying to change as little as possible -- so I maintained the existing value.

@schmmd @epwalsh If either of you tell me a magic # to put here instead I'm happy to add it.

schmmd: 1024 is my suggestion

codeviking, line 30:
```suggestion
    @lru_cache(maxsize=1024)
```

codeviking: Done.

PR #388, opened 2020-05-01
Add tests for the new bidaf endpoint.

codeviking:
Before porting over a new model I thought it'd be nice to get
some tests in place for verying these endpoints.

This tests for the new bidaf endpoint and does so in
a fashion that should make them repurposable across endpoints
where possible.

~~The tests are integration tests and depend on a running server and not
yet wired up to CI -- that'll come later.~~

The tests can now be executed without a server.

### DIFF
--- a/allennlp_demo/bidaf/test_api.py
+++ b/allennlp_demo/bidaf/test_api.py
@@ -1,0 +1,8 @@
+import os
+import requests
+
+from allennlp_demo.common.testing import ModelEndpointTests
+
+# For this endpoint we can simply use the standard tests.
+class TestBidafEndpoint(ModelEndpointTests):
+    pass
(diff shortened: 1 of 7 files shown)


allennlp_demo/bidaf/test_api.py
```diff
@@ -0,0 +1,14 @@
...
+import requests
+import pytest
+
+from allennlp_demo.common.testing import ModelEndpointTests
+from allennlp_demo.bidaf.api import BidafModelEndpoint
+
+@pytest.fixture
+def client():
+    endpoint = BidafModelEndpoint()
+    return endpoint.app.test_client()
```
codeviking, line 11: @epwalsh Every endpoint will now need to provide a test client like so. Let me know what you think and if there's another way we might do this.

epwalsh: That seems reasonable

PR #391, opened 2020-05-01
Test model loading

dirkgr:
This is not a fast test, especially without caching.


codeviking:
FYI, in the model-serving effort I'm adding tests that download the model archive and test the HTTP endpoints each model is responsible for.  Here's the base test case that every model implements (as to provide more, customize behavior, or skip tests): https://github.com/allenai/allennlp-demo/blob/master/allennlp_demo/common/testing.py.

Given this maybe we just wait until that lands as I think it might test what we're doing here too?

epwalsh:
This will be covered by https://github.com/allenai/allennlp-demo/pull/445

PR #393, opened 2020-05-01
Port NAQANet to the new serving mechanism.

codeviking:
This adds code for serving the NAQANet reading comprehension model
via the new mechanism.

The hotflip attacker is again disabled due to an outstanding bug,
and prediction test needed to be modified as the output is different.


matt-gardner:
Hotflip bug is fixed in master.

codeviking:
> Hotflip bug is fixed in master.

Great! Does this auto-release, or is there a manual process for crafting a new release client?

matt-gardner:
I think you need to update the base docker image (which gets auto released), and the models commit in requirements.txt.

PR #394, opened 2020-05-02
Port the NMN endpoint to the new serving mechanism.

codeviking:
This migrates the NMN endpoint to the new serving mechanism.

This one is tricky, as currently it's served via a custom image that was
built once from a fork of the demo. This makes a great candidate to
validate the new serving mechanism.

I had to modify the way we load the model archive to allow overrides
to be passed. I also had to migrate the `allennlp_plugins.allennlp_models`
import to the entrypoint for each endpoint, as in this case that code isn't
needed nor available, since the NMN endpoint relies on an older version
of AllenNLP.

Because of the custom, non `pip` managed dependencies the best way to run
this endpoint locally is via `docker`. I've added a `README` that provides
instructions for doing so.

Currently the server boots up just fine,~~but the tests don't quite
pass yet. I'm still working that out.~~ Tests pass now. They just use a lot of RAM.


allennlp_demo/common/http.py
```diff
@@ -59,20 +59,28 @@ class ModelEndpoint:
...
         self.model = model
         self.app = Flask(model.id)
         self.configure_logging()
 
-        self.predictor = Predictor.from_path(model.archive_file, model.predictor_name)
+        # TODO: I'm no Python expert, but this seems a bit risky. There's all sorts of side-effect
+        # that importing at runtime could entail. That said we know we only import this once
+        # while instantiating an endpoint, so maybe it's ok. I'll ask about this before merging.
+        if load_plugins:
+            import allennlp_plugins.allennlp_models
```
codeviking, line 71: Please pay particular attention to this line. I have no idea if it's a bad idea to do this, and would love input from someone who knows more about Python than I :).

codeviking:
> Currently the server boots up just fine, but the tests don't quite
pass yet. I'm still working that out.

The tests fail due to an OOM exception. I have Docker configured
to allow 8 GB, which seems like it should be enough.

I'll try increasing this next week to see if things pass. That said
I think we'll want to noodle on this a bit too.

codeviking:
Increasing the limit to 11.7 GB allowed the tests to pass. They took ~10.8 GB or so at peak utilization.

codeviking:
By disabling that test memory usage hovers around 5 GB (which is what we see in prod too).

codeviking:
Disabling the test allows things to pass, though memory usage is still high.

allennlp_demo/common/http.py
codeviking: Disregard, I factored this out.

PR #397, opened 2020-05-04
add black, flake8, mypy to CI

epwalsh:


.flake8
```diff
@@ -0,0 +1,25 @@
+[flake8]
+max-line-length = 115
```
codeviking, line 2: I think we've been using 100 characters here, but it might be different for AllenNLP. @schmmd would know best.

epwalsh: Yea I just copied this from AllenNLP. I'm indifferent to the actual number though. 100 is fine with me if that's what you guys want

codeviking: Ah, stick with `115` if that's what AllenNLP is using.

PR #398, opened 2020-05-04
Fix linting errors

epwalsh:


.dockerignore
```diff
@@ -1,3 +1,5 @@
 __pycache__
 
 *.pyc
+
```
codeviking, line 4: There's a `.dockerignore` file in `allennlp_demo` as well. We'll probably need this to that file as well (maybe you already did that).

allennlp_demo/bidaf/api.py
```diff
@@ -1,16 +1,18 @@
-import sys
 import os
-# This adds ../ to the PYTHONPATH, so that allennlp_demo imports work.
-sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))))
```
codeviking, line 4: Removing this breaks the ability for someone to run it without docker, I think, without modifying their `PYTHONPATH`, right?

Is there a reason we don't want to do this? I ported it from AllenNLP: https://github.com/allenai/allennlp/blob/master/allennlp/__main__.py#L12

allennlp_demo/bidaf/api.py
```diff
@@ -1,16 +1,18 @@
...
-import allennlp_plugins.allennlp_models
+from allennlp.common.util import import_module_and_submodules
+
 from allennlp_demo.common import config, http
 
+
 class BidafModelEndpoint(http.ModelEndpoint):
     def __init__(self):
+        # TODO: should just import the exact submodule we need.
+        import_module_and_submodules("allennlp_models")
```
codeviking, line 11: Why is this preferable to `import allennlp_plugins.allennlp_models`?  That method was per @schmmd.

epwalsh: We just got rid of `allennlp_plugins.allennlp_models` so that won't work anymore

allennlp_demo/bidaf/api.py
epwalsh: I was mainly concerned with having to add this in every single `api.py` file. But for your question about running it directly, it would work fine as long as they're running it from the root of the repo since Python always checks the current directory for packages.

The other option - which is usually how we handle this issue - is creating a `setup.py` file so we can `pip install -e .` the `allennlp_demo` package, so it will always be importable no matter what directory you're in

.dockerignore
epwalsh: Good catch

allennlp_demo/nmn_drop/api.py
```diff
@@ -1,15 +1,16 @@
...
-from allennlp.common.util import import_submodules
-from typing import Mapping
+
 
 class NMNDropModelEndpoint(http.ModelEndpoint):
     def __init__(self):
-        import_submodules("semqa")
+        # TODO: only import the exact submodules we need.
+        import_module_and_submodules("semqa")
+        import_module_and_submodules("allennlp_models")
```
codeviking, line 13: We don't need to import `allennlp_models` for this model, as it doesn't use any of the code there (it's a custom model that's provided via the `nmn-drop` package defined above.

I don't think we can use `import_module_and_submodules` as I don't think that existed in the AllenNLP version this uses (`0.9.0-unreleased`).

Just make sure to try to build and run this like so to verify it's working:

```
cd allennlp_demo
docker build -f nmn_drop/Dockerfile -t nmn:latest
docker run --rm -it --entrypoint pytest nmn:latest -x -ra nmn_drop/
```

allennlp_demo/bidaf/api.py
codeviking: I agree we should keep boilerplate minimal.

I think it's fine to just recommend that people run things from the root. That said it seems like we should then advise that they build from the root as well.

So along with this change could you:

1. Update the docs in the `allennlp_demo/README.md` file (that'll eventually probably merge into the parent).

2. Update the paths in each `Dockerfile` in `allennlp_demo/*` to include the full path, i.e. change `bidaf/Dockerfile` to `allennlp_demo/bidaf/Dockerfile`.

3. Remove [this line](https://github.com/allenai/allennlp-demo/blob/master/allennlp_demo/common/.skiff/cloudbuild-deploy.yaml#L16).

Then everything is from the root and I think it's ok to remove this code. Let me know what you think.

epwalsh: Makes sense

allennlp_demo/bidaf/api.py
codeviking: Ah, that's a good reason :).

It should be easy to import only what's needed. I think in this case you could at least do `allennlp_models.rc`, or even `allennlp_models.rc.bidaf` (I'm not sure if this exists). Might be worth trying now since you're doing the work?

allennlp_demo/README.md
```diff
@@ -60,6 +55,6 @@ that's restarted whenever you change the code.
 Each endpoint has tests. To run them run `pytest` in the model endpoint's working directory:
 
 ```bash
-cd bidaf/
+cd allennlp_demo/bidaf/
 pytest
```
codeviking, line 59: With the change to be from root I think this will need to change to:

```
pytest allennlp_demo/bidaf
```

Then you can remove the `cd`. The tests import `allennlp_demo` so they'll need to be executed from root (I think).

PR #400, opened 2020-05-04
Add a few more models, improve porting process

schmmd:


epwalsh:
Adds a couple more models and simplifies the process of adding new demos and maintaining the existing (soon to be deployed) demos by providing a default `Dockerfile` and `requirements.txt`. So now if a demo doesn't provide its own `Dockerfile` in its sub-directory, it will be built using the default.

This has a couple benefits. For one, it's flexible because we can still override the image build easily by just adding a custom Dockerfile in the demo's subdirectory. But it also makes maintaining all of the demos easier because most of them will be using the same default Dockerfile and requirements, which provides motivation to try to keep the demos up-to-date and consistent with the AllenNLP version they depend on.

Makefile
```diff
@@ -19,16 +18,39 @@ typecheck :
...
+			-f allennlp_demo/$*/Dockerfile \
+			-t allennlp-demo-$*:$(DOCKER_LABEL) \
+			- < $*-context.tar.gz; \
+	else \
+		echo "Using default build from allennlp_demo/Dockerfile"; \
+		docker build \
+			-f allennlp_demo/Dockerfile \
+			-t allennlp-demo-$*:$(DOCKER_LABEL) \
+			--build-arg DEMO=$* - < $*-context.tar.gz; \
+	fi
```
codeviking, line 41: We may not need the whole custom context solution post re-org, right?

epwalsh: Yea we could probably remove it

allennlp_demo/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+#> ENTRYPOINT [ "python" ]
+#> CMD [ "allennlp_demo/${DEMO}/api.py" ]
+#
+# The reason is that Python will interpret the string in CMD literally as just
+# "allennlp_demo/${DEMO}/api.py" without substituting environment variables.
+# Environment variable evaluation is a something that a shell handles, so we need
+# to wrap the `python` command in a shell (`/bin/bash` in this case).
+ENV DEMO ${DEMO}
+ENTRYPOINT [ "/bin/bash" ]
+CMD [ "-c", "python allennlp_demo/${DEMO}/api.py" ]
```
codeviking, line 37: Interesting! I think there's a gotcha here. The Python program will be a child of `bash`, and accordingly might not get signals as we'd expect. I think this can cause some weird behavior on Kubernetes, like getting in the way of graceful shutdowns (as the program won't receive the kill signal, and may instead need to be hard stopped).

Here's [Docker's documentation](https://docs.docker.com/engine/reference/builder/#entrypoint) on this:

> The shell form prevents any CMD or run command line arguments from being used, but has the disadvantage that your ENTRYPOINT will be started as a subcommand of /bin/sh -c, which does not pass signals.

I think things might work properly if we use `exec. Maybe the trick is to do `-c`, `exec`, `python ...`?  I'm not totally sure.  

To test I'd shell into the container, run `ps` and make sure the `python` executable isn't a child of `bash`.

allennlp_demo/constituency_parser/.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,17 @@
...
+
+local common = import '../../common/.skiff/common.libsonnet';
+local model = import '../model.json';
+
+function(image, cause, sha, env, branch, repo, buildId)
+    // This tells Kubernetes what resources we need to run.
+    // For more information see:
+    // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes
+    local cpu = '50m';
+    local memory = '1Gi';
```
codeviking, line 16: @schmmd @epwalsh Did we get these values from what we see in production?

For this model I see it peak at 3 GB:

image

allennlp_demo/constituency_parser/requirements.txt
```diff
@@ -1,3 +1,4 @@
+# See the Docker image for the version of AllenNLP
```
codeviking, line 1: I don't see the allennlp version in the `Dockerfile`, am I missing it?

allennlp_demo/coref/.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,17 @@
...
+
+local common = import '../../common/.skiff/common.libsonnet';
+local model = import '../model.json';
+
+function(image, cause, sha, env, branch, repo, buildId)
+    // This tells Kubernetes what resources we need to run.
+    // For more information see:
+    // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes
+    local cpu = '50m';
+    local memory = '1Gi';
```
codeviking, line 16: Looks like we're just setting the default here. This is fine for CPU (we know we use more CPU when things are active), but might not work so well for RAM.

To set this value I usually go to the [workloads](https://console.cloud.google.com/kubernetes/workload?project=ai2-reviz&pageState=(%22workload_list_table%22:(%22f%22:%22%255B%257B_22k_22_3A_22Is%2520system%2520object_22_2C_22t_22_3A11_2C_22v_22_3A_22_5C_22False_~*false_5C_22_22_2C_22i_22_3A_22is_system_22%257D%255D%22))) page in K8s and look at usage statistics for the model in question, then try to set it to what appears to be the model's normal usage (per replica, the deployment screen shows usage for all replicas, I think).

allennlp_demo/constituency_parser/requirements.txt
schmmd: The comment should be removed or we should use a base image in the Dockerfile.  My bad, I think a lost a commit a long time ago...

allennlp_demo/dependency_parser/.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,17 @@
...
+
+local common = import '../../common/.skiff/common.libsonnet';
+local model = import '../model.json';
+
+function(image, cause, sha, env, branch, repo, buildId)
+    // This tells Kubernetes what resources we need to run.
+    // For more information see:
+    // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes
+    local cpu = '50m';
+    local memory = '1Gi';
```
codeviking, line 16: Another spot where we probably need to set specific values.

Also, did we create a GCB trigger yet for these and verify that it deploys?

doc/AddingModelEndpoint.md
```diff
@@ -23,11 +23,14 @@ selected a model, follow these steps to port it to the new solution:
     ```bash
     cd allennlp_demo/oscar_drop
     rsync -rv --filter="- __pycache__" --filter="- .pytest_cache" \
-        ../oscar_drop/ ./
+        ../bidaf/ ./
```
codeviking, line 26: Oops! Nice catch.

doc/AddingModelEndpoint.md
```diff
@@ -23,11 +23,14 @@ selected a model, follow these steps to port it to the new solution:
...
     ```
 
-3. Next, open up the `Dockerfile` and replace all instances of `bidaf/` with
-   `oscar_drop/`.
+3. If the default Dockerfile (`allennlp_demo/Dockerfile`) works for `oscar_drop`, you don't need
+   to create a new one - the default will automatically be used. If, for some reason, you
+   do need a custom Dockerfile to build the demo (for example, if `oscar_drop` requires
+   a different version of `allennlp` than the default) then you can just copy the default over
+   to `oscar_drop/Dockerfile` and modify it as you need.
```
codeviking, line 33: 👍

doc/AddingModelEndpoint.md
```diff
@@ -55,9 +58,13 @@ selected a model, follow these steps to port it to the new solution:
...
-    cd allennlp_demo
-    docker build . -f allennlp_demo/oscar_drop/Dockerfile -t oscar_drop:latest && \
-        docker run  -v $HOME/.allennlp:/root/.allennlp --rm -it --entrypoint pytest oscar_drop:latest -x -ra
+    make oscar_drop-test
+    ```
+
+    And then
+
+    ```bash
+    make oscar_drop-run
```
codeviking, line 67: These are great. I've been meaning to try using `make` for building `Docker`. Thanks for paving the way.

Makefile
codeviking: Cool, I'll let you make that change if / when -- just seems like a nice simplification if it works out.

allennlp_demo/constituency_parser/.skiff/webapp.jsonnet
schmmd: I didn't modify them and I must have missed this.  Ideally we could re-use what we had before.

https://github.com/allenai/allennlp-demo/blob/master/.skiff/webapp.jsonnet#L326
https://github.com/allenai/allennlp-demo/blob/master/models.json#L66

But maybe we didn't set limits before and so we need to inspect production?  The memory will change with the requests potentially, as in large request can spike the memory usage.

allennlp_demo/Dockerfile
schmmd: Can't we just have an empty `ENTRYPOINT`?

schmmd: https://stackoverflow.com/questions/37634483/default-docker-entrypoint

epwalsh: So we're not sure why this works without exec but somehow `python` actually does become PID 1.

codeviking: @epwalsh Confirmed that with this setup Python still gets pid 1, and appears to be receiving signals properly -- so I think we're good to go with this, we can always change it later if it proves to be a problem.

allennlp_demo/constituency_parser/requirements.txt
epwalsh: I forgot to update this one

allennlp_demo/constituency_parser/.skiff/webapp.jsonnet
codeviking: We're not setting limits, just requests -- so they'll be allowed to burst.

Using prior values seems fine.  We can probably leave CPU at `50m`, as our demos use very little CPU and this lets us fit as many demos on a single machine as is allowed by RAM.

PR #403, opened 2020-05-04
Testing for each model

epwalsh:
- Adds easy Docker testing for each model by wrapping the Docker commands in the Makefile so you can now build and run any demo with `make allennlp_demo/bidaf/Dockerfile`
- Adds a testing job to GitHub Actions for each demo we have so far
- Refactors the Dockerfiles for each demo so that they inherit from the corresponding allennlp base image


Makefile
```diff
@@ -12,3 +14,12 @@ typecheck :
 		--no-strict-optional \
 		--no-site-packages \
 		--cache-dir=/dev/null
+
+allennlp_demo/%/Dockerfile : context.tar.gz FORCE
+	docker build -f $@ -t allennlp-demo-$*:$(DOCKER_LABEL) - < context.tar.gz
+	docker run --rm -p 8000:8000 -v $$HOME/.allennlp:/root/.allennlp allennlp-demo-$*:$(DOCKER_LABEL) $(ARGS)
+
+context.tar.gz : FORCE
+	tar -czvf $@ $(DEMO_SRCS)
```
codeviking, line 23: TIL that you could prepare a tar file to replace the context. Nice!

Makefile
```diff
@@ -12,3 +15,12 @@ typecheck :
 		--no-strict-optional \
 		--no-site-packages \
 		--cache-dir=/dev/null
+
+allennlp_demo/%/Dockerfile : context.tar.gz FORCE
+	docker build -f $@ -t allennlp-demo-$*:$(DOCKER_LABEL) - < context.tar.gz
+	docker run --rm -p $(DOCKER_PORT):8000 -v $$HOME/.allennlp:/root/.allennlp allennlp-demo-$*:$(DOCKER_LABEL) $(ARGS)
```
codeviking, line 21: Do we need to expose the port at all?  The `-p` flag just exposes `8000` on the local, host machine and maps it to the container port.  If we're not making any requests from the host to the container (I don't think we are), then we can omit the `-p` flag all together and the `DOCKER_PORT` environment variable.

allennlp_demo/common/testing/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .model_endpoint_test_case import ModelEndpointTestCase
+from .rc import make_rc_endpoint_test_case
```
codeviking, line 2: I'm not familiar with this import syntax. Is this just "relative to the current directory"? Is this conventional / preferred to fully qualified imports, and if so are we converting all of them to this syntax?

allennlp_demo/common/testing/rc.py
```diff
@@ -0,0 +1,108 @@
...
+        as necessary.
+        """
+
+        endpoint: ModelEndpoint
+
+        def setUp(self):
+            super().setUp()
+            with open(
+                self.PROJECT_ROOT / "allennlp_demo" / "common" / "testing" / "fixtures" / "rc.json"
+            ) as fh:
```
codeviking, line 30: Does this need to be absolute? Couldn't this be loaded relative to the `__file__`? That way we don't need `PROJECT_ROOT`?

allennlp_demo/common/testing/__init__.py
epwalsh: Yeup, relative to current directory. It's mostly a matter of preference and because I was being lazy. But I should probably change them to absolute imports since we're doing that everywhere else.

allennlp_demo/naqanet/test_api.py
```diff
@@ -1,24 +1,18 @@
...
-from flask.testing import FlaskClient
-from allennlp_demo.common.testing import ModelEndpointTests
+from allennlp_demo.common.testing import make_rc_endpoint_test_case
 from allennlp_demo.naqanet.api import NAQANetModelEndpoint
-from typing import List
 
 
-@pytest.fixture
-def client():
+class TestNAQANetModelEndpoint(make_rc_endpoint_test_case()):  # type: ignore
```
codeviking, line 7: Why the factory? It's ultimately something I can follow somewhat easily, but I don't see any immediate reason for it.  Couldn't this just be `TestNAQANetModelEndpoint(RcModelEndpointTestCase)`?

allennlp_demo/nmn_drop/test_api.py
```diff
@@ -1,18 +1,14 @@
+from typing import List
+
 import pytest
 
-from flask.testing import FlaskClient
 from allennlp_demo.nmn_drop.api import NMNDropModelEndpoint
-from allennlp_demo.common.testing import ModelEndpointTests
-from typing import List
+from allennlp_demo.common.testing import make_rc_endpoint_test_case
```
codeviking, line 6: These imports are different than the ones we have in the test file. We should choose one way imo and stick with it.

allennlp_demo/common/testing/__init__.py
codeviking: I also realized this is in a `__init__.py` file and I don't know what that does. What's this for / do?

allennlp_demo/common/testing/model_endpoint_test_case.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ModelEndpointTestCase(TestCase):
+    """
+    A base class to run model endpoint tests.
+
+    Each endpoint should have a corresponding test class that derives from this.
+    """
+
+    PROJECT_ROOT: Path = (Path(__file__).parent / os.pardir / os.pardir / os.pardir).resolve()
```
codeviking, line 18: Does this `/` magic work on non linux machines? Not that it's an explicit requirement, just curious.

allennlp_demo/naqanet/test_api.py
epwalsh: Yea so I tried to make that work and in the process lost a lot of time of my life that I will never get back. The problem is that pytest will discover the base class `RcModelEndpointTestCase` (since it inherits from `unittest.TestCase`) and try to run all of the test methods on it, even though we only want to run tests on the subclasses. We can tell pytest to skip anything from the base class, but we see a whole bunch of skipped tests in the output. So I solved that by creating this base class dynamically so that pytest never knows about it.

Now that I think about, we might be able to just use pytest test classes (https://docs.pytest.org/en/latest/xunit_setup.html?highlight=setup_class#class-level-setup-teardown) and not have this issue. I'll give that try.

allennlp_demo/common/testing/model_endpoint_test_case.py
epwalsh: I'm not sure. But allennlp doesn't support windows anyway so none of this probably works on windows

Makefile
epwalsh: Yea not for running the tests. I think I'll split off the make command for testing so we I'll get rid of that

allennlp_demo/naqanet/test_api.py
matt-gardner: Jumping in here; I just happened to notice this point.  If the methods on the base class don't have `test` in their name, then pytest won't use them for tests.  We did this in our [data iterator tests](https://github.com/allenai/allennlp/blob/8532dc5d906e29e055f36b7d76ec0a84453181f6/allennlp/tests/data/iterators/basic_iterator_test.py#L22), back before we removed them.  I haven't looked at all at this particular issue, just leaving this here in case it's useful.  If not, feel free to ignore.

codeviking: Gotcha. I was simply overriding tests I wanted to skip like so:

```
class SomeEndpointTests(RcModelEndpointTestCase)
   @pytest.mark.skip(reason="Skip it, skip it good")
   def test_predict():
        pass
```

Here's an example: https://github.com/allenai/allennlp-demo/blob/master/allennlp_demo/nmn_drop/test_api.py#L33

epwalsh: @matt-gardner I think the main issue though is that we *want* to test those methods, just not in the base class

allennlp_demo/common/testing/model_endpoint_test_case.py
codeviking: Gotcha. Sounds good.

The [docs](https://docs.python.org/3/library/pathlib.html#operators) aren't totally clear, but it seems like it might work by converting a `Path` to a `WindowsPath`. Anyway, to your point it's something we can probably ignore.

allennlp_demo/common/testing/__init__.py
epwalsh: Putting imports in the `testing/__init__.py` file means you can do

```
from allennlp_demo.common.testing import ModelEndpointTestCase
```

instead of

```
from allennlp_demo.common.testing.model_endpoint_test_case import ModeEndpointTestCase
```

allennlp_demo/naqanet/test_api.py
matt-gardner: One option (again, take it or leave it, I don't have a strong opinion here): rename the test method to have the implementation of the test, but have subclasses call the tests directly with `super()`.  So, e.g., `assert_predict_works`.  Then you're sharing basically all of the implementation, and you have just a tiny bit of (customizable) boilerplate in the subclasses.

epwalsh: Ok phew, dumping unittest.TestCase and just using pytest-style classes avoids this whole issue

epwalsh:
I think we could make `ModelEndpointTestCase` more useful but I'll hold off on trying to guess what to add until we start porting more demos over.

allennlp_demo/common/testing/__init__.py
codeviking: Gotcha. That's nice. Thanks.

codeviking:
> I think we could make ModelEndpointTestCase more useful but I'll hold off on trying to guess what to add until we start porting more demos over.

I think every endpoint has a predict endpoint. So I think the code for that, and the code that tests that it's cached and such could likely be in the base class.

The input and expected output might change, so maybe we define a function for providing these values that individual test classes can override.

```python
class BaseModelEndpointTest
   def predict_input():
       # ...
   def verify_predict_output(output: JsonDict):
       # ...
   def test_predict()
       output = self.client.post("/predict", self.predict_input())
       self.verify_predict_output(output)
```

This way the other models don't need to separately implement the nuts and bolts of the predict cases, nor the caching logic, etc.

If that sounds good to you, maybe you can port that stuff over in a follow up PR?

Issue #401, opened 2020-05-04
Add namespace to sentiment analysis model config and upload new model


gagb:
Hi,

The [usage instructions](https://demo.allennlp.org/sentiment-analysis) result in an error. I am not sure what the fix is. You can reproduce it using the following steps:

1. Create a new conda environment and installed AllenNLP:
`pip install allennlp==1.0.0rc3 allennlp-models==1.0.0rc3`
 
2. Run the example code:
```
from allennlp.predictors.predictor import Predictor
import allennlp_models.sentiment
predictor = Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/sst-roberta-large-2020.02.17.tar.gz")
predictor.predict(
  sentence="a very well-made, funny and entertaining picture."
)
```

Here's the error message I get. `run.py` contains code in step #2:
```
File "run.py", line 4, in <module>
    print(predictor.predict(
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/predictors/text_classifier.py", line 25, in predict
    return self.predict_json({"sentence": sentence})
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/predictors/predictor.py", line 66, in predict_json
    return self.predict_instance(instance)
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/predictors/predictor.py", line 189, in predict_instance
    outputs = self._model.forward_on_instance(instance)
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/models/model.py", line 141, in forward_on_instance
    return self.forward_on_instances([instance])[0]
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/models/model.py", line 167, in forward_on_instances
    outputs = self.make_output_human_readable(self(**model_input))
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/models/basic_classifier.py", line 171, in make_output_human_readable
    [
  File "/home/bansalg/miniconda2/envs/bert3/lib/python3.8/site-packages/allennlp/models/basic_classifier.py", line 172, in <listcomp>
    self.vocab.get_token_from_index(token_id.item(), namespace=self._namespace)
[...]

PR #404, opened 2020-05-05
Working model for SST

dirkgr:
Fixes https://github.com/allenai/allennlp-demo/issues/401.


dirkgr:
This won't work until the upload is complete. At 700kb/s from my house, this will take a while.

ZhaofengWu:
If this is a newly trained model, the model description should be updated with its performance. If you're just changing the config then nvm.

On a related note, maybe we can have a centralized table that lists all our models with their performance numbers?

dirkgr:
We'll have to work on an overall way to organize our models better, but it won't make it into 1.0.

PR #407, opened 2020-05-06
Refactor permalinks to not store the full response.

codeviking:
This PR adjusts the current permalink mechanism so that the
full response body isn't stored in the database. This means the
UI code is now responsible for reissueing the request to the server.

This has a few benefits:

1. The response data is large. Removing this might significantly reduce
    the database size.
2. It ensures the UI can render the output. If we attempt to rerender
    cached output with a new version of the UI, we can't be sure things
    will work.
3. This gets us closer to how the new permalink system will work. In fact,
    with this approach I can re-use the current database instead of provisioning
    a new one. This avoids any work associated with migrating data and such.

As part of this I also removed the `headers` column, as it appears to
have been for diagnostic purposes only and no-one on the team appears to
have used it. This should save some additional space, though I'd expect
it to be negligible.

I tested this with permalinks from prod for two models, coref resolution
and the bidaf-elmo reading comprehension model. Everything worked fine.
I also verified that new permalinks were generated for these two models
as well.

After deploying this I plan to execute the following SQL statements
manually after I've verified that everything is working:

```sql
ALTER TABLE queries DROP COLUMN headers;
ALTER TABLE queries DROP COLUMN response_data;
```

Lastly as part of this I made the waiting indicator look a little nicer
(it's a loading spinner instead of text) and fixed the issue where on
larger screens there's a light-grey region on the bottom of the
left-hand and main panes.  I also removed some stray code that
was related to the demo's prior life as an `<iframe />` -- the code
should be safe to remove as that functionality was deprecated
last quarter.

@schmmd and @dirkgr are likely the best reviewers here, as I think
they might have the most context.


codeviking:
Deleting `headers` won't save much:

```
demo=> select pg_size_pretty(sum(pg_column_size(headers))) FROM queries;
 pg_size_pretty
----------------
 1139 MB
(1 row)
```

Deleting `response_data` will save an immense amount of data:

```
demo=> select pg_size_pretty(sum(pg_column_size(response_data))) FROM queries;
 pg_size_pretty
----------------
 76 GB
(1 row)
```

Wow. That's ~75% of the space we use right now...!

Naturally, we then wonder how much size the `request_data` takes up:

```
demo=> select pg_size_pretty(sum(pg_column_size(request_data))) FROM queries; pg_size_pretty
----------------
 674 MB
(1 row)
```

Less than `headers!`. We're going to have a nice n' tidy, small little database after this lands.

PR #408, opened 2020-05-06
Separate linting from formatting, add instructions to development docs

epwalsh:


.github/workflows/ci.yml
```diff
@@ -30,11 +30,19 @@ jobs:
     - name: Install requirements
       run: |
         pip install -r requirements.txt
+
+    - name: Format
+      if: always()
+      run: |
+        make lint
```
codeviking, line 37: I assumed this would be something like `make format`, or something like that? Is it really the same as the `Lint` target?

epwalsh: Good catch

allennlp_demo/README.md
```diff
@@ -55,3 +55,7 @@ Each endpoint has tests. To run them run `pytest` in the model endpoint's workin
 ```bash
 pytest allennlp_demo/bidaf
 ```
+
+We also use [`flake8`](http://flake8.pycqa.org/) to lint our code base, [`black`](https://black.readthedocs.io) to check formatting, and [`mypy`](http://mypy-lang.org/) to run typechecks.
+
+You can run each of these checks, respectively, with `make lint`, `make format`, and `make typecheck`.
```
codeviking, line 61: 👍 Thanks for writing this up.

PR #409, opened 2020-05-06
Fix permalinks for a few models.

codeviking:
This fixes permalinks for a few models by:

- Modifying `/permadata` to not require the model name, since it's
  not used.
- Adding the `--no-models` flag which starts up a Flask server
  without any models. We use this to serve the `/permadata` from
  a single deployment.
- Deploy a single version of the Python app and route all `/permadata`
  requests to it.
- Remove rules routing requests to `/permadata/:model-name` to each
  model endpoint.
- Set the default backend, rather than using a path that captures
  all traffic.
- Remove the `/task` endpoints which aren't used anymore.

Permalinks for 3 tasks (sentiment analysis, dependency parsing and
textual entailment) broke when I factored the UI out to be served by
NGINX.

The issue is that for these three tasks there's no model which matches
the URL path. This causes the `/permadata` request to be issued to the
NGINX container instead of the Python application, which doesn't work.

For instance, for the reading comprehension task the bidaf model has
the name `reading-comprehension`. The request for
`/permadata/reading-comprehension` is therefore sent to the Python
endpoint for that model. The permadata requests for the other
reading comprehension models are *also* sent to that endpoint, but 
this works fine as the model name parameter isn't used.

For textual entailment and the tasks where this doesn't work there's
no generic model whose name matches the URL path for the task exactly.
The default route therefore forwards the request to the NGINX server
that serves the UI, which obviously isn't well suited to handle the request.

This all felt too brittle to maintain, so I factored this out into
something that makes a little more sense and is closer to where we're
headed. Luckily this has only been broken for a week or so, not months.

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -155,7 +155,7 @@ local cloudsql_volumes = [
 ];
 
 // Each model typically has its own service running that handles several different endpoints
-// (/predict, /permadata, /task, /attack, etc.).  This is a convenience function that will route
+// (/predict, /task, /attack, etc.).  This is a convenience function that will route
 // all of those endpoints to the model service, instead of the main frontend.
 // TODO(mattg): we might want to change this some day so that all model backend services start with
 // /model/[model-name], so that we only have to have one route per backend, instead of this mess of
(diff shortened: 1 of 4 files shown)


schmmd:
I only had a brief moment to look, but looks good to me!

PR #410, opened 2020-05-06
Rearrange things a bit.

codeviking:
Part 1 of the reorg effort. If this works out @epwalsh will take the second
half, which may or may not involve lifting the remaining files in the root
into an `api/` or `models/` directory or something like that.

This moves the old Python application into `old/`, and renames `demo/` to
`ui/`. I split the root level `.skiff/` directory into two pieces, one
chunk now lives in `old/.skiff` and is responsible for deploying the old
application. The other lives in `ui/.skiff` and is responsible for
deploying just the UI bits.

For good measure I also did some cleanup in the `webapp.jsonnet` files
to consolidate to `underscore_based_names` as the use of `camelCase`
and underscores bugged me.


codeviking:
> LGTM! I think the "Checks" workflow will pass if you go into the Makefile and set SRC = allennlp_demo/ instead of whatever it is now.

Sweet. Thanks for the catch. I made the change and just reduced it to *only* the `allennlp_demo` module for now. Let me know if I should do something differently there.

PR #411, opened 2020-05-07
Docker image spring cleaning.

codeviking:
In the reorganization effort I forgot to port add a `.dockerignore` file
to old. This does that and cleans up a few things that didn't need to
end up in the resulting images.


old/Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM allennlp/commit:26e313b7d11eacf3a78c09c34608ad1bfd7db120
 LABEL maintainer="allennlp-contact@allenai.org"
 
-WORKDIR /stage/allennlp
+WORKDIR /app
```
codeviking, line 5: In shelling into the container I noticed that we're mixing the source for AllenNLP with the demo application. This seems like it could cause weird issues, so I moved it to it's own root level directory and made sure things still work.

ui/.dockerignore
```diff
@@ -1,3 +1,5 @@
+.skiff/
```
codeviking, line 1: Thanks @epwalsh for noticing that we didn't need these files in our images!

PR #412, opened 2020-05-07
Part 1 of the Great Permalinks Migration

codeviking:
This is part 1 of the *Great Permalinks Migration*, an epic saga
that tells the tale of a lone traveler, seeking simple, persistent
AllenNLP demo links.

In this PR I move the `/permadata` endpoint into a new, self-standing
service that has no relation to the old code. I tried to change as
little as possible but couldn't help but take a few liberties:

1. I switched the `POST` to a `GET`, which seems more canonical.
2. I replaced the `Permadata` nomenclature with `PermaLink`. This makes
    more sense to me, but let me know if it seems worse, names are hard.
3. I removed some logic that "health checked" the database before trying to
    query it. I'm pretty sure this was only required because the long-lived db
    connection was being closed after being open for too long. Instead we now
    open a database connection as needed. This is potentially a little slower, but
    a lot less complex, and we've had no issues with the latency associated with
    this in other high traffic demos.

There's also some nomenclature changes to reflect that all endpoints in `api/` are no
longer model specific.

Also, as part of this I got `docker-compose` wired up as a way to set up an end-to-end
local development environment with a single model. I added documentation for this,
please make sure to give that a read. I'm sure there's a few typos in there, as that's
my ™️ after all.

### Before Merging

1. [x] Update the existing Cloud Build triggers, changing the `_MODEL_PATH` parameter to `SUBMODULE_PATH`.
2. [x] Delete the existing endpoints, since the names changed. They'll be recreated after
         this merges and #1 is done.
3. [x] Copy the database credentials into the permalinks api namespace.

## Part 2 Coming to a Theatre Near You Soon

Part 2, after this lands, involves adding an endpoint for creating
permalinks, at which point I'll change the UI to use and remove all
permalink related code from the old application alltogether.


allennlp_demo/permalinks/db.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_result(self, perma_id: int) -> Optional[PermaLink]:
+        try:
+            return self.data[perma_id]
+        except IndexError:
+            return None
+
+    @classmethod
+    def from_environment(cls) -> Optional["InMemoryDemoDatabase"]:
+        return InMemoryDemoDatabase()
```
codeviking, line 200: I copied this from `server/db.py`. It's unmodified for the most part.

allennlp_demo/permalinks/models.py
```diff
@@ -0,0 +1,30 @@
...
+    Convert the permalink slug back to the integer id.
+    Returns ``None`` if slug is not well-formed.
+    """
+    byt = slug.encode("utf-8")
+    try:
+        int_bytes = base64.urlsafe_b64decode(byt)
+        return int(int_bytes)
+    except (binascii.Error, ValueError):
+        logging.getLogger(__name__).error("Unable to interpret slug: %s", slug)
+        return None
```
codeviking, line 30: This is also a mostly unmodified port of `server/permalinks.py`.

codeviking:
@epwalsh @schmmd This is ready for a look. Sorry for the scope-creep -- I wanted to get the end-to-end environment up so that I could test the full permalinks flow before putting this out. It now works (but took a bit more effort than I first thought). That said I think it'll be super useful for future efforts.

.github/workflows/ci.yml
```diff
@@ -86,3 +86,13 @@ jobs:
     - name: Build and test image
       run: |
         cd api && make nmn_drop-test DOCKER_TAG=$GITHUB_SHA
+
+  permalinks:
+    name: Permalinks API
+    runs-on: [self-hosted]
```
codeviking, line 92: @epwalsh I'm not sure if I need this for this test target. This API doesn't need AllenNLP or anything fancy, so maybe I can do with a standard runner (though I'm not sure if this better or not, let me know).

codeviking:
I need to merge this to merge forward so I'm going to go ahead. That said I'm happy to tend to feedback after the fact.

.github/workflows/ci.yml
epwalsh: Standard runner will work fine for this. Docker is already available on the standard runners

epwalsh, line 92:
```suggestion
    runs-on: ubuntu-latest
```

codeviking: 🤘

PR #415, opened 2020-05-07
generalize ModelEndpointTestCase

epwalsh:


api/allennlp_demo/common/testing/model_endpoint_test_case.py
```diff
@@ -15,16 +16,28 @@ class ModelEndpointTestCase:
...
 
     endpoint: ModelEndpoint
     """
     Should be defined by each subclass.
     """
 
+    predict_payload: Dict[str, Any]
+    """
+    Payload to send to the /predict route.
+    """
```
codeviking, line 32: Is it standard Python practice to put comments for members after them too? (I have no idea, just asking for informational purposes).

api/allennlp_demo/common/testing/model_endpoint_test_case.py
```diff
@@ -15,16 +16,28 @@ class ModelEndpointTestCase:
...
+    This points to the absolute path of `api/`. So, for example, this file is located
+    in `PROJECT_ROOT / allennlp_demo / common / testing`.
+    """
 
     endpoint: ModelEndpoint
     """
     Should be defined by each subclass.
     """
 
+    predict_payload: Dict[str, Any]
```
codeviking, line 29: Naming is always tough. What if we called this `predict_inputs` so it's the same as the `predict_json` parameter name that is corresponds to? Here's the docs for the API I'm referencing: http://docs.allennlp.org/master/api/predictors/predictor/#predict_json

api/allennlp_demo/common/testing/model_endpoint_test_case.py
```diff
@@ -42,3 +55,55 @@ def setup_method(self):
 
     def teardown_method(self):
         pass
+
+    def test_predict(self):
+        """
+        Test the /predict route.
+        """
+        response = self.client.post("/predict", json=self.predict_payload)
```
codeviking, line 63: Do pytest classes share the same client, and thereby the same cache, or will a new client always be instantiated for each test? If the client is shared we may need the `no_cache` argument here to be sure we're not serving a cached response.

api/allennlp_demo/constituency_parser/test_api.py
```diff
@@ -4,3 +4,4 @@
 
 class TestConstituencyParserModelEndpoint(ModelEndpointTestCase):
     endpoint = ConstituencyParserModelEndpoint()
+    predict_payload = {"sentence": "If I bring 10 dollars tomorrow, can you buy me lunch?"}
```
codeviking, line 7: 🥪 💰

api/allennlp_demo/common/testing/model_endpoint_test_case.py
epwalsh: I like it!

api/allennlp_demo/common/testing/model_endpoint_test_case.py
epwalsh: Yea, it's weird but it is actually the official way to document members defined like this

> String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called "attribute docstrings". - https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

schmmd: Python 🤦

codeviking: Haha, thanks. TIL!

api/allennlp_demo/common/testing/model_endpoint_test_case.py
epwalsh: So, yes - they do all share the same client because otherwise the model would have to be re-loaded before each test method is run. That's a good point though. I just refactored a bit so that the caches will be cleared before each test method is run. Do you think that provides enough isolation for test methods?

epwalsh: https://github.com/allenai/allennlp-demo/pull/415/commits/ed82be7ebe87a1ea9eabb22b323a7e7cf22d7bc0

epwalsh:
> Just curious, why did the imports shift? Is it convention to put stdlib imports first (I can start doing that)? Can we get black to do this for us?

Yea, the convention is

```
stdlib imports
*blank line*
third party imports
*blank line*
local imports
```

api/allennlp_demo/common/testing/model_endpoint_test_case.py
codeviking: Sure, that looks good. 

I prefer just using the `no_cache` argument as we did before as it seems a little simpler, but it's not a big deal, this works too.

epwalsh:
> Maybe we change `setup_method` to `clear_caches`, or is this a pytest convention? (If so that's ok, and leave it as is).

This is a pytest thing https://docs.pytest.org/en/latest/xunit_setup.html#method-and-function-level-setup-teardown

PR #418, opened 2020-05-09
Use the new API to create permalinks as well.

codeviking:
This is part two of the great permalinks refactor. In this PR I
remove the code from the old Python application that writes permalinks
to the database and use the new service instead.

The API contract didn't change as to prevent the number of things
that are changing at once. Future PRs will start reshaping things into
the format we need for the new endpoints.


codeviking:
I'm going to go ahead and merge now, but am happy to tend to feedback after the fact.

PR #421, opened 2020-05-11
adds remaining models

epwalsh:
- atis parser
- wikitables parser
- nlvr parser
- semantic role labeling
- glove sentiment analysis
- named entity recognition
- next token lm
- open information extraction
- roberta snli
- roberta mnli **(this is broken, but is also broken in the old / currently deployed demo)**
- masked lm

I verified that all of these run and pass their tests locally, except for Roberta MultiNLI, which fails with vocab `KeyError` exception. It appears to be broken on the demo site as well.

### DIFF
--- a/api/allennlp_demo/atis_parser/.skiff/webapp.jsonnet
+++ b/api/allennlp_demo/atis_parser/.skiff/webapp.jsonnet
@@ -1,0 +1,17 @@
+/**
+ * This file defines the infrastructure we need to run the API endpoint on Kubernetes.
+ *
+ * For more information on the JSONNET language, see:
+ * https://jsonnet.org/learning/getting_started.html
+ */
+
+local common = import '../../common/.skiff/common.libsonnet';
+local model = import '../model.json';
+
+function(image, cause, sha, env, branch, repo, buildId)
+    // This tells Kubernetes what resources we need to run.
+    // For more information see:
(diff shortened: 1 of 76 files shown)


api/allennlp_demo/atis_parser/.skiff/webapp.jsonnet
```diff
@@ -0,0 +1,17 @@
...
+
+local common = import '../../common/.skiff/common.libsonnet';
+local model = import '../model.json';
+
+function(image, cause, sha, env, branch, repo, buildId)
+    // This tells Kubernetes what resources we need to run.
+    // For more information see:
+    // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes
+    local cpu = '50m';
+    local memory = '1Gi';
```
epwalsh, line 16: Haven't set up the trigger yet. Will go through all of them later and update the memory if needed

api/allennlp_demo/atis_parser/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM allennlp/commit:93024e53c1445cb4630ee5c07926abff8943715f
+
+WORKDIR /app/
+COPY allennlp_demo/atis_parser/requirements.txt allennlp_demo/atis_parser/requirements.txt
+ENV EXCLUDE_ALLENNLP_IN_SETUP true
```
codeviking, line 5: Is this something we added to this PR, or something that's specific to AllenNLP?

PR #423, opened 2020-05-12
Add new permalinks fields to support the new serving mechanism.

codeviking:
This adds two new fields to the `queries` database table to support
permalinks for models served by the new mechanism. These fields
are `task_name` and `model_id`.  The `model_id` will tell the UI
which endpoint to invoke, while the `task_name` is purely to be used
by administrators to reverse engineer a full URL from the database. 
Right now that's possible by introspecting the `request_data`, but
this isn't terribly easy to do.

The new fields are optional for backwards compatibility. This way the
old models can continue creating permalinks as they did before, and 
we can transition new models to the new mechanism on a case by case
basis.

The new fields aren't yet used, but I added a test to make sure they
work. In the next PR I'll transition a the Reading Comprehension demo
to use the new endpoints and make use of this functionality.  Part of this
will involve migrating the old permalinks to the new format -- I haven't quite
thought through how to do that yet.

I'll need to execute the `ALTER TABLE` command that I added to `schema.sql`
manually as there's no mechanism for executing migrations. This will 
accordingly involve a small window of downtime (the site will still work, 
permalinks will just be broken which the UI and API handle gracefully).

### DIFF
--- a/api/allennlp_demo/permalinks/db.py
+++ b/api/allennlp_demo/permalinks/db.py
@@ -21,7 +21,14 @@ class DemoDatabase:
     In the future it could also be used to store user-submitted feedback about predictions.
     """
 
-    def insert_request(self, requester: str, model_name: str, request_data: Dict) -> Optional[int]:
+    def insert_request(
+        self,
+        requester: str,
+        request_data: Dict,
+        model_name: Optional[str] = None,
+        model_id: Optional[str] = None,
+        task_name: Optional[str] = None,
+    ) -> Optional[int]:
         """
         Add the request to the database so that it can later
         be retrieved via permalink.
@@ -45,14 +52,14 @@ class DemoDatabase:
 
 # SQL for inserting requests into the database.
 INSERT_SQL = """
-        INSERT INTO queries (model_name, requester, request_data, timestamp)
-        VALUES (%(model_name)s, %(requester)s, %(request_data)s, %(timestamp)s)
+        INSERT INTO queries (model_name, requester, request_data, timestamp, model_id, task_name)
+        VALUES (%(model_name)s, %(requester)s, %(request_data)s, %(timestamp)s, %(model_id)s, %(task_name)s)
         RETURNING id
         """
 
 # SQL for retrieving a prediction from the database.
 RETRIEVE_SQL = """
-        SELECT model_name, request_data
+        SELECT model_name, request_data, model_id, task_name
         FROM queries
         WHERE id = (%s)
         """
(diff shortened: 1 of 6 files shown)


api/allennlp_demo/permalinks/db.py
```diff
@@ -108,7 +115,14 @@ def from_environment(cls) -> Optional["PostgresDemoDatabase"]:
...
+    def insert_request(
+        self,
+        requester: str,
+        request_data: Dict,
+        model_name: Optional[str] = None,
+        model_id: Optional[str] = None,
+        task_name: Optional[str] = None,
+    ) -> Optional[int]:
         try:
             conn = self.connect()
```
epwalsh, line 127: This is unrelated to the PR, but I'm just curious why we create a new connection on each `insert` and `get`? I'm not too familiar with `pyscopg2`. Is it not recommended to re-use the same connection across requests?

codeviking: Yup, that's the recommendation -- the motivation being to reduce latency, as opening a connection is expensive.

That said there's also a cost using a connection pool.  We'd need to create a pool, decide if we need a thread-safe one or not, and make sure we're always releasing a connection when we're done with them.  This isn't terribly complicated, but never the less adds complexity. 

Ultimately in practice we've seen very little additional latency associated with opening a connection for every request -- so we opted to keep things simple. If / when we find there's unacceptable latency we can easily wire in a pool (we could also do so now if you think we should).

Of note, I did [change this mechanism when porting the code over from the old application](https://github.com/allenai/allennlp-demo/commit/10a3031168520c21681e785918fa91e1f6df2b60#diff-384834467c99fd7e3b7f587f9bcd8cb5L100). Previously we were opening a single connection per replica and re-using it, but we also had this bizarre health check mechanism to ensure the connection was active. This was probably necessary because connections were long-lived and often idle -- so the database closed them to save resources.  Using the build in Pool abstractions and setting `maxconn=1` would've likely achieved the same effect, and the `psycopg` library would've taken care of handling idle / closed connections and all that jazz.

Let me know what you think -- happy to put a connection pool in place if we think we should.

epwalsh: Gotcha. Eh, I don't know, it might be worth it. I can look into it since I want to get familiar with pysocopg2 anyway. I'll make an issue to follow up on this so I don't forget.

codeviking: Sounds good.

PR #426, opened 2020-05-12
fix new linting errors from new version of flake8

epwalsh:
New version of flake8 was just released which introduces some new lints.

### DIFF
--- a/api/allennlp_demo/permalinks/api.py
+++ b/api/allennlp_demo/permalinks/api.py
@@ -102,7 +102,7 @@ class PermaLinkService(Flask):
                 )
                 return jsonify(int_to_slug(id))
             except psycopg2.Error as err:
-                self.logger.exception(f"Error saving permalink: ", err)
+                self.logger.exception("Error saving permalink: %s", err)
                 raise InternalServerError("Unable to create permalink")
 
 


api/allennlp_demo/permalinks/api.py
```diff
@@ -102,7 +102,7 @@ def create_permalink() -> Response:
                 )
                 return jsonify(int_to_slug(id))
             except psycopg2.Error as err:
-                self.logger.exception(f"Error saving permalink: ", err)
+                self.logger.exception("Error saving permalink: %s", err)
```
codeviking, line 105: Any rationale for not using the Python 3's `f` strings? I tend to use them quite liberally -- maybe it's just an argument that "less is more"?

epwalsh: I think f-strings are great, but the usage here isn't quite right. To actually make use of the f-string it would have to be `self.logger.exception(f"Error saving permalink: {err}")`. But this isn't recommended for logging messages since the f-string `f"Error saving permalink: {err}"` would always be evaluated, even if the logger doesn't end up emitting the message.

codeviking: >  To actually make use of the f-string it would have to be self.logger.exception(f"Error saving permalink: {err}")

Haha! Good catch. This was a bug / unintentional mistake. Glad the linter caught it :).

> But this isn't recommended for logging messages since the f-string f"Error saving permalink: {err}" would always be evaluated, even if the logger doesn't end up emitting the message.

Awesome. Really helpful to know. I am but a Python padowan, and appreciate the training in the force :).

epwalsh: *Unfortunately* I have many years of Python under my belt. But I'm glad I can occasionally impart some of this knowledge that I've reluctantly acquired on my peers to save them from the traps of Python.

PR #427, opened 2020-05-12
Migrate the RC demo to use the new endpoint.

codeviking:
This change modifies the RC demo to use the new endpoints. It should
be representative of the work we do demo-by-demo.

I redacted a few identifiers that I changed in the interest of
cleanliness to reduce the amount of work. I also fixed an issue
with routing in the local environment.

After this I plan on adding a local config that proxies all
API requests to production for working on the UI only. This
should make it easier to test changes that don't require working
on the API.

### DIFF
--- a/old/models.json
+++ b/old/models.json
@@ -1,27 +1,4 @@
 {
-    "reading-comprehension": {
-        "archive_file": "https://storage.googleapis.com/allennlp-public-models/bidaf-model-2020.03.19.tar.gz",
-        "predictor_name": "reading-comprehension",
-        "max_request_length": 311108
-    },
-    "elmo-reading-comprehension": {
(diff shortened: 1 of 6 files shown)


old/models.json
```diff
@@ -1,27 +1,4 @@
...
-        "max_request_length": 25819
-    },
-    "nmn-drop": {
-        "archive_file": "https://storage.googleapis.com/allennlp-public-models/drop-nmn-2020.04.04.tar.gz",
-        "predictor_name": "drop_demo_predictor",
-        "max_request_length": 25819,
-        "overrides": "{\"model\": {\"debug\":true, \"beam_size\":1}}",
-        "memory": "4Gi",
-        "image": "gcr.io/ai2-reviz/allennlp-demo-nmn-drop:34cfc7e1122c6ca4a8cd0d7f23cd2fb12c64a057"
-    },
```
codeviking, line 24: 🎉

PR #428, opened 2020-05-12
Make it easy to point a local UI at production.

codeviking:
This adds a way to run a local verison of the UI that uses the
production services. This makes it easier to test changes to the
UI that don't require backend updates.

I chose to still run a permalinks service as to keep permalinks
from developer environments from cluttering the production database.


codeviking:
I'm going to merge this so I can use it to do some additional testing of #427.  Happy to tend to feedback after as warranted.

PR #429, opened 2020-05-12
Port NER to the new endpoints.

codeviking:
This modifies the NER demo to use the new endpoints.


codeviking:
> Ahhh I need to learn JS!

We'll get you ramped up on TypeScript instead :). Hopefully @jonborchardt will do some work this quarter (or next) to port the demo to TypeScript -- the compiler goes a long way to make these types of large applications easy to work with!

PR #445, opened 2020-05-13
test in cloud build

epwalsh:
This adds a testing step to the Cloud Builds and removes the model tests from GitHub Actions.

### DIFF
--- a/api/allennlp_demo/common/.skiff/cloudbuild-deploy.yaml
+++ b/api/allennlp_demo/common/.skiff/cloudbuild-deploy.yaml
@@ -10,6 +10,13 @@ steps:
   args: [
     $_SUBMODULE_PATH-build, DOCKER_BASE_NAME=gcr.io/$PROJECT_ID/$REPO_NAME, DOCKER_TAG=$COMMIT_SHA
   ]
+- id: test
+  dir: api
+  name: gcr.io/cloud-builders/docker
+  entrypoint: make
+  args: [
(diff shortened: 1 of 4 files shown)


epwalsh:
It works! https://console.cloud.google.com/cloud-build/builds/69842a45-65e7-412d-a813-3eff902ddc84?project=ai2-reviz

api/allennlp_demo/common/.skiff/cloudbuild-deploy.yaml
```diff
@@ -10,6 +10,13 @@ steps:
   args: [
     $_SUBMODULE_PATH-build, DOCKER_BASE_NAME=gcr.io/$PROJECT_ID/$REPO_NAME, DOCKER_TAG=$COMMIT_SHA
   ]
+- id: test
+  dir: api
+  name: gcr.io/cloud-builders/docker
+  entrypoint: make
+  args: [
+    $_SUBMODULE_PATH-test, DOCKER_BASE_NAME=gcr.io/$PROJECT_ID/$REPO_NAME, DOCKER_TAG=$COMMIT_SHA
+  ]
```
codeviking, line 19: This will run as part of the deploy, after a merge to master. So this will block a deploy if the tests fail -- is that the intent.

We can wire them up to run for PRs too -- we'll just need to put this step (and any other required ones) in a separate `cloudbuild.yaml` file and configure a trigger for it.  We'd need a trigger per model (which would end up being a lot), but maybe it's worth it?

epwalsh: So, I think its definitely good to run the tests before deploying. And I was going to ask you about hooking this up to pull requests. I'll give it a try.

PR #447, opened 2020-05-13
update build timeout

epwalsh:
I think it's fairly rare for a cloud build to last longer than 15 minutes, but one did on the last master commit. So this just ups the timeout to 20 minutes to be safe.


epwalsh:
> Also, when a timeout fails the deployment times out the deployment will still occur as long as the rudder deploy step has been executed. The job timing out doesn't cancel the change.

I was wondering about this. Thanks!

PR #449, opened 2020-05-14
Add the info service.

codeviking:
This adds a simple service that lists all endpoints and some
information we know about each.

The endpoint is a little slow, it takes about 3 or 4 seconds to
collect all of the information. I imagine this is because of the
blocking requests we make to each endpoint. We could either
accept the latency or use concurrency to make it faster. I imagine
making the requests concurrently would greatly speed things up
since they're all I/O bound.

Here's what the output looks like:

```
  {
    "id": "semantic-role-labeling",
    "info": {
      "allennlp": "1.0.0rc3",
      "archive_file": "https://storage.googleapis.com/allennlp-public-models/bert-base-srl-2020.03.24.tar.gz",
      "id": "semantic-role-labeling",
      "overrides": null,
      "predictor_name": "semantic-role-labeling"
    },
    "url": "https://demo.allennlp.org/api/semantic-role-labeling"
  },
  {
    "id": "wikitables-parser",
    "info": {
      "allennlp": "0.9.1-unreleased",
      "archive_file": "https://storage.googleapis.com/allennlp-public-models/wikitables-model-2020.02.10.tar.gz",
      "id": "wikitables-parser",
      "overrides": null,
      "predictor_name": "wikitables-parser"
    },
    "url": "https://demo.allennlp.org/api/wikitables-parser"
  }
```

This can provide some nifty things, like:

```
[I] ~/h/a/api ❯❯❯ curl -s http://localhost:8000 | jq -r '.[].info.allennlp' | sort | uniq
0.9.0-unreleased
0.9.1-unreleased
1.0.0rc3
null
```

The `null` is from the permalinks service. It'd be easy to filter out with a `grep -v`, or something.

If this looks good I'll need to create a trigger after deploying and get this out the door.

### DIFF
--- a/api/allennlp_demo/info/prod_api.py
+++ b/api/allennlp_demo/info/prod_api.py
@@ -1,0 +1,29 @@
+"""
+Production entry point that launches the gevent WSGI server in place of Flask's
+default one.
+
+The workload for this service is entirely I/O bound, so should improve the
+service's ability to handle more concurrent connections without using a
+multi-process WSGI server like `gunicorn`.
+"""
+from gevent import monkey, pywsgi
+
+# This ensures third party libs use the non-blocking socket, so that we actually
+# benefit from using `gevent`. They recommend that it be one of the first
+# lines executed in your program, which is why we import it and immediately
+# execute the command.
+#
+# See: http://www.gevent.org/intro.html#monkey-patching
+monkey.patch_all()
+
+import kubernetes  # noqa: E402
+
+from allennlp_demo.info.api import InfoService  # noqa: E402
+
+if __name__ == "__main__":
+    kubernetes.config.load_incluster_config()
+    app = InfoService()
+    # Note, we set `log=None` to disable the `gevent` request log in favor
+    # of ours which includes the info we want.
+    server = pywsgi.WSGIServer(("0.0.0.0", 8000), app, log=None, error_log=app.logger)
+    server.serve_forever()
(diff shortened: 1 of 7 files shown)


api/allennlp_demo/info/prod_api.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+from gevent import monkey, pywsgi
+
+# This ensures third party libs use the non-blocking socket, so that we actually
+# benefit from using `gevent`. They recommend that it be one of the first
+# lines executed in your program, which is why we import it and immediately
+# execute the command.
+#
+# See: http://www.gevent.org/intro.html#monkey-patching
+monkey.patch_all()
```
schmmd, line 17: Don't you feel sheer terror when calling this function?  I do from the name alone.

codeviking: Totally. The confusing part is it doesn't seem like it makes any difference. I timed both the prod and local API responses and they're both 3s. 

I figured since the service is almost entirely I/O bound (we're just fetching stuff from different places) that `gevent` would pay dividends, but it didn't.

codeviking:
>  If we identify an obvious and easy way to speed this up, I'm all for it, otherwise, let's just live with it for now.

The biggest time sink are the individual calls to each individual API endpoint. If we parallelized that with a `Thread` this would likely get a lot faster.  I could probably do that without too much work, but the code would get more complex -- lmk what you think?

schmmd:
Based on what you just said--I'd leave it for now.  Let's address if we have a clear idea that doesn't make the code much more complex and we're sure it will work ;-)

codeviking:
> Based on what you just said--I'd leave it for now. Let's address if we have a clear idea that doesn't make the code much more complex and we're sure it will work ;-)

The information doesn't change unless a deploy occurs, so another way would be to write the data to a single configmap that this service reads.  That said this would involve some build time acrobatics that isn't terribly straightforward either.

epwalsh:
I made some improvements with https://github.com/allenai/allennlp-demo/pull/450

PR #450, opened 2020-05-14
make endpoint info requests concurrently

epwalsh:
This does a couple things.

1. It uses a helper module `grequests` to do the endpoint info requests concurrently.
2. It uses a single `requests.Session` for all of the HTTP requests to reduce overhead.

On my computer this reduces the time to serve a single request from ~1.6 seconds to ~150 milliseconds.

### DIFF
--- a/api/allennlp_demo/info/api.py
+++ b/api/allennlp_demo/info/api.py
@@ -3,18 +3,29 @@ The info service is an endpoint that lists all API endpoints and some additional
 about each. It's purely for discovery, and intended for use primarily by administrators (but is
 ok to be public.
 """
-from typing import Optional, Mapping, Any
-from dataclasses import dataclass
 
-import flask
-import kubernetes
-import requests
-import logging
+# This needs to be imported first, since it calls gevent.monkey.patch_all(),
+# which ensures third party libs (like requests) use the non-blocking socket, so that we actually
+# benefit from using `gevent`. They recommend that it be one of the first
+# lines executed in your program, which is why we import it first.
+#
+# See: http://www.gevent.org/intro.html#monkey-patching
+import grequests
 
-from allennlp_demo.common.logs import configure_logging
+from dataclasses import dataclass  # noqa: E402
+import logging  # noqa: E402
+from typing import Optional, Mapping, Any, List  # noqa: E402
+
+import flask  # noqa: E402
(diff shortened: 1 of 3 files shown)


epwalsh:
We could also throw in caching with a short-ish cache timeout if we wanted to get really crazy

api/allennlp_demo/info/api.py
```diff
@@ -49,33 +64,52 @@ def from_ingress(ingress: kubernetes.client.ExtensionsV1beta1Ingress) -> Optiona
...
 
+        cache = Cache(config={"CACHE_TYPE": "simple"})
+        cache.init_app(self)
+
+        self.session = Session()
+        adapter = adapters.HTTPAdapter(pool_maxsize=100)
+        self.session.mount("https://", adapter)
+
         @self.route("/", methods=["GET"])
+        @cache.cached(timeout=10)
```
epwalsh, line 83: Short cache timeout ensures our responses stay fresh, but still limits the amount of requests we have to make to our model endpoints and greatly improves response time during high traffic (not that I'd expect a lot of traffic to this).

codeviking: Is this 10 seconds, or 10 minutes?

epwalsh: 10 seconds

api/allennlp_demo/info/api.py
```diff
@@ -3,18 +3,35 @@
...
+# These together ensure that third party libs (like requests) use the non-blocking socket,
+# so that we actually benefit from using `gevent`. They recommend that it be one of the first
+# lines executed in your program, which is why we import these first.
+#
+# See: http://www.gevent.org/intro.html#monkey-patching
+import grequests
+from gevent import monkey
 
-from allennlp_demo.common.logs import configure_logging
+monkey.patch_all()
```
codeviking, line 17: Does this play nice with the Flask WSGI server?

api/allennlp_demo/info/api.py
codeviking: Gotcha, sounds good.

api/allennlp_demo/info/api.py
epwalsh: I just tested it to make sure. Looks all good

PR #452, opened 2020-05-14
Add the `sha` and `github_url` to the info output.

codeviking:
@epwalsh Noted that the SHA would be helpful. Skiff apps have this by default so I just had to wire it in real quick.

### DIFF
--- a/api/allennlp_demo/info/api.py
+++ b/api/allennlp_demo/info/api.py
@@ -38,8 +38,8 @@ class Endpoint:
     id: str
     url: str
     info: Optional[Mapping[str, Any]]
-    sha: Optional[str]
-    github_url: Optional[str]
+    commit_sha: Optional[str]
+    commit_url: Optional[str]
 
     @staticmethod
     def from_ingress(ingress: kubernetes.client.ExtensionsV1beta1Ingress) -> Optional["Endpoint"]:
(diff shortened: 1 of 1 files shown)


api/allennlp_demo/info/api.py
```diff
@@ -62,7 +64,14 @@ def from_ingress(ingress: kubernetes.client.ExtensionsV1beta1Ingress) -> Optiona
         path = paths[0].path.rstrip("(/.*))?$")
         url = f"https://{rule.host}{path}"
 
-        return Endpoint(id, url, None)
+        sha = ingress.metadata.annotations.get("apps.allenai.org/sha")
+        repo = ingress.metadata.annotations.get("apps.allenai.org/repo")
+        if sha is not None and repo is not None:
+            github_url = f"https://github.com/{repo}/commit/{sha}"
```
epwalsh, line 70:
```suggestion
            github_url = f"https://github.com/allenai/{repo}/commit/{sha}"
```

codeviking: Ah, the `repo` includes the `allenai/` designation, so we don't need to add that.  In theory we could support repositories from different orgs n' such.

api/allennlp_demo/info/api.py
```diff
@@ -38,6 +38,8 @@ class Endpoint:
     id: str
     url: str
     info: Optional[Mapping[str, Any]]
+    sha: Optional[str]
+    github_url: Optional[str]
```
epwalsh, line 42: I think it might be more descriptive to name these `commit_sha` and `commit_url`. What do you think?

api/allennlp_demo/info/api.py
codeviking: Err, you're right, and I'm wrong, good catch!

api/allennlp_demo/info/api.py
codeviking: Sure.

api/allennlp_demo/info/api.py
epwalsh: When I ran locally it didn't include the org

image

PR #464, opened 2020-06-04
fix margins on a task that does not use mormal panes

### LINKED ISSUE #303: Spacing is off on model internals
image

https://demo.allennlp.org/masked-lm

Note that the model internals but up against the vertical bar instead of having the appropriate amount of padding.

jonborchardt:
fixes #303


jonborchardt:
before: 
image

after:
image

Issue #483, opened 2020-06-12
NMN always returns "No results" despite results in /predict


schmmd:
![image](https://user-images.githubusercontent.com/954798/84521752-52a17a00-ac8a-11ea-9365-aa4c987de3c6.png)

![image](https://user-images.githubusercontent.com/954798/84521740-4f0df300-ac8a-11ea-8996-50f8eba252ae.png)

aimichal:
Okay, I can repro in prod by going to https://demo.allennlp.org/reading-comprehension/, choosing the model NMN and the example "Who stars in The Matrix?". The server responds with an answer, but "No answer returned" is shown.

aimichal:
Same thing locally. Debugging now!

PR #484, opened 2020-06-12
Reading Comprehension NMN fix

aimichal:
This fixes https://github.com/allenai/allennlp-demo/issues/483.

I can repro by going to /reading-comprehension, choosing the model NMN and the example "Who stars in The Matrix?". The server responds with an answer, but "No answer returned" is shown.

Looks like the "No answer returned." is shown due to the `NoAnswer()` call at the bottom of the default case in `AnswerByType`:

https://github.com/allenai/allennlp-demo/blob/ae62fc3774bf98488745267e8e8aa3ccd29c34f0/ui/src/components/demos/ReadingComprehension.js#L397-L426

In this default case, `responseData` is destructed to extract the field `best_span_str`, which doesn't exist in the response to this example. The response does, however, have a `predicted_ans` field. Here's what `responseData` looks like in this default case:

> image

When I use `predicted_ans` instead of `best_span_str` then the NMN model works for this example:

> image

I don't know of this is the right fix; looking for advice from someone closer to the systems involved.

### DIFF
--- a/ui/src/components/demos/ReadingComprehension.js
+++ b/ui/src/components/demos/ReadingComprehension.js
@@ -395,7 +395,7 @@ const AnswerByType = ({ responseData, requestData, interpretData, interpretModel
       }
 
       default: { // old best_span_str path used by BiDAF model
-          const { best_span_str: bestSpanStr } = responseData;
+        const { best_span_str: bestSpanStr } = responseData;
         if(question && passage && bestSpanStr) {
           const start = passage.indexOf(bestSpanStr);
           const head = passage.slice(0, start);
@@ -434,15 +434,15 @@ const Output = (props) => {
     case NMNModel.name:
     case NMNModel.modelId:
       return (
-          <div className="model__content answer">
-            <nmn.Output response={props.responseData}/>
-          </div>
+        <div className="model__content answer">
+          <nmn.Output response={props.responseData}/>
+        </div>
       );
     default:
       return (
-          <div className="model__content answer">
-            <AnswerByType {...props} />
-          </div>
+        <div className="model__content answer">
+          <AnswerByType {...props} />
+        </div>
       )
   }
 }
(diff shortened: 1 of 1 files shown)


schmmd:
@matt-gardner can you help @aimichal out here?  Somehow this regressed recently and I'm not sure why.

aimichal:
There's also another use of `best_span_str` in ReadingComprehension.js, which I didn't touch:

https://github.com/allenai/allennlp-demo/blob/ae62fc3774bf98488745267e8e8aa3ccd29c34f0/ui/src/components/demos/ReadingComprehension.js#L212-L221

(...and also in api/allennlp_demo/common/testing/rc.py, FYI.)

ui/src/components/demos/ReadingComprehension.js
```diff
@@ -394,8 +394,8 @@ const AnswerByType = ({ responseData, requestData, interpretData, interpretModel
         return NoAnswer();
       }
 
-      default: { // old best_span_str path used by BiDAF model
-        const { best_span_str: bestSpanStr } = responseData;
```
matt-gardner, line 398: Does this break bidaf?  Looking at the code, I think it probably would.  I would recommend a system that prefers one of these, but falls back to the other if the first isn't present.  I don't have a strong preference on which should be preferred.

matt-gardner:
I'm confused because the `answer_type` doesn't appear to be present in the response data, and that's the real problem.  What you're looking at is a passage span, and it should have been labeled as such.  Why isn't `answer_type` present?

matt-gardner:
Oh, hang on, this is the wrong UI entirely.  The components that are supposed to be used are here: https://github.com/allenai/allennlp-demo/tree/master/ui/src/components/demos/reading-comprehension/nmn.  This model looks like it got ported using the wrong demo component.

ui/src/components/demos/ReadingComprehension.js
aimichal: Yeah, it did. Good call. Thanks!

aimichal:
Okay, new fix is to the branching in the Output component. It was only branching if request data's model was `NMN (trained on DROP)`, but in practice this was `nmn`. Now it branches on either.

The permalinks database has both of these model values, so this should allow backwards compatibility with those old values.

aimichal:
I've confirmed that BiDAF and NMN models work for me now, locally:

> image

PR #503, opened 2020-06-29
Fgn

dirkgr:
Using increasingly stupid names for this branch so that Marina works with it


dirkgr:
That test fails with error 137 (which means OOM). @epwalsh, how can we get this thing more memory?

epwalsh:
@dirkgr I'm not sure how to allocate more memory to these builds. We might have to change the machine type.

dirkgr:
We solved this problem in a different way.

PR #507, opened 2020-06-30
fixes simple gradient interpret for next_token_lm

epwalsh:
From https://github.com/allenai/allennlp-models/pull/85


api/allennlp_demo/next_token_lm/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM allennlp/commit:637dbb159082999c546ac2fc64746b88e5c9d1b5
```
schmmd, line 1: Can you add a comment about why this version is needed?  Please also specify whether it's, e.g. before/after 1.0.

PR #525, opened 2020-08-06
update npm

aimichal:
This is meant to fix the vuln:

> image

I did it by deleting package-lock.json, updating package.json to have the newer version of npm, and running npm install.

Demo site runs locally after this change.


ui/package.json
```diff
@@ -9,7 +9,7 @@
     "colormap": "^2.3.1",
     "hierplane": "^0.2.1",
     "lodash": "^4.17.11",
-    "npm": "6.14.4",
+    "npm": "6.14.6",
```
codeviking, line 12: It's odd that `npm` shows up here. Given `npm` is bundled with node, the only reason for it to exist here is if it's somehow invoked programmatically, within a JavaScript program.

I bet that's not happening, as I can't find any usages via some simple `grep` / `ag` commands.

Maybe we can remove it?

jonborchardt: +1

aimichal:
```
~/code/allennlp-demo/ui ‹michalg/deps› $ npm install depcheck -g
~/code/allennlp-demo/ui ‹michalg/deps› $ depcheck
Unused dependencies
* ajv
* npm
* react-proptypes
* react-vis
Missing dependencies
* react-d3-tree: ./src/components/BeamTree.js
```

# 🤔

aimichal:
I removed ajv, npm, react-proptypes, and react-vis and the UI still seems to run.

I left src/components/BeamTree.js in place, because it's not referenced anywhere. @epwalsh if you haven't already, you should consider deleting this dead code to avoid confusion.

PR #528, opened 2020-09-11
Add Skiff Stats.

codeviking:
This adds support for user metrics provided by Skiff stats.

If this ends up providing the metrics your group needs we can at some point turn off Google Analytics. For now I've left them in place, but let me know if no one is using that (in which case I'd propose we remove them, as it's better for our user's privacy and will make the page a little faster).

### DIFF
--- a/ui/public/index.html
+++ b/ui/public/index.html
@@ -115,5 +115,6 @@
         <path d="M30.2,13.7c0.1,0.3,0.1,0.5,0.2,0.8c0,0.1,0,0.1,0.1,0.2c1.2-0.7,2.4-1.3,3.7-1.9c0-0.1,0-0.1,0.1-0.2 c0-0.1,0-0.1,0.1-0.2H30C30,12.8,30.1,13.3,30.2,13.7z" class="st0"></path>
       </symbol>
     </svg>
+    <script src="https://stats.allenai.org/init.min.js" data-app-name="allennlp-demo" async></script>
   </body>
 </html>


codeviking:
Weird, here's the error:

```
 black --check allennlp_demo/
would reformat /home/runner/work/allennlp-demo/allennlp-demo/api/allennlp_demo/info/api.py
Oh no! 💥 💔 💥
1 file would be reformatted, 80 files would be left unchanged.
make: *** [format] Error 1
Makefile:12: recipe for target 'format' failed
##[error]Process completed with exit code 2.
```

I didn't change that file, so it's probably ok to merge this as is. That said for good measure I tried to fix the issue:

```
[I] ~/h/a/api ❯❯❯ black allennlp_demo/                                                                                                                                                                                 (allennlp-demo)
All done! ✨ 🍰 ✨
81 files left unchanged.
```

...but nothing was reformatted. So maybe CI and the version of `black` we install locally is different?

schmmd (approved):
I don't know re: black.  I'm not terribly familiar with it, but Pete probably has a good guess.

epwalsh:
Ah, yea.. black recently released a new version which changes the formatting. We should probably have this pinned in the `dev-requirements.txt`.

codeviking:
@epwalsh Ok for me to merge w/ the failure and let you make that update (and formatting fix) separately?

epwalsh:
@codeviking yeup!

codeviking:
🤘

PR #535, opened 2020-09-18
add initial work for vqa ui

jonborchardt:


dirkgr (approved):
Only took a cursory glance

PR #540, opened 2020-10-02
More memory while building images

dirkgr:


dirkgr:
This is the wrong file to do this in.

PR #545, opened 2020-10-06
Update allennlp, utililze pretrained module from allennlp-models, run tests on GH Actions

epwalsh:
Sorry @codeviking, this is a big PR 😉 

TODO:
- [x] Add `allennlp-semparse` to `api/requirements.txt` and remove custom requirements and Dockerfiles for the semparse demos. First https://github.com/allenai/allennlp-semparse/pull/28 needs to be merged.
- [x] Run CI for all demos on the self-hosted runner
- [x] Fix vilbert test
- [ ] Update allennlp-models once next nightly release goes out.
- [x] Make new release of semparse and then update the dependency here.
- [x] Update instructions for adding new models


.github/workflows/ci.yml
```diff
@@ -7,13 +7,15 @@ on:
...
+  schedule:
+  - cron: '7 6 * * *'
 
 jobs:
-  # Runs common light-weight checks like linting and type checking.
   api_checks:
     name: API Checks
     runs-on: ubuntu-latest
-
+    # Don't run nightly builds for forks.
```
codeviking, line 17:
```suggestion
    # Don't run nightly builds for forks, unless it's a PR to master
```

I'm not sure if that's right, but I think this explains the `github.event_name != 'schedule'` bit.

.github/workflows/ci.yml
```diff
@@ -50,17 +56,66 @@ jobs:
...
+          "nmn_drop",
+          "open_information_extraction",
+          "roberta_mnli",
+          "roberta_sentiment_analysis",
+          "roberta_snli",
+          "semantic_role_labeling",
+          "transformer_qa",
+          "vilbert_vqa",
+          "wikitables_parser",
+        ]
```
codeviking, line 92: It's a bummer that we can't derive this. It might be worth writing up a doc about what it takes to add an endpoint. I know there was some confusion with how to add the `vilgert_vqa` endpoint, for instance (the trigger didn't get created).

If you get something started let me know, I can help fill in the Google Cloud Build / Skiff specific bits.

epwalsh: We already have [this](https://github.com/allenai/allennlp-demo/blob/master/doc/AddingModelEndpoint.md). Do you think that's enough? I did add a step to that explaining how you have to update the CI config.

I also added a CI check that ensures the CI config is up-to-date:

https://github.com/allenai/allennlp-demo/blob/update-allennlp/.github/workflows/ci.yml#L100-L121

So CI will fail if someone adds a new model without updating the CI config.

.github/workflows/ci.yml
epwalsh: I think it's right, we have the same `if` condition in other repos. We want this to run on PRs and master commits for the main repo and all forks, but we don't want it to run on a nightly schedule for forks (because people have complained about this before).

api/allennlp_demo/common/config.py
```diff
@@ -11,11 +16,87 @@ class Model:
...
+
+            return load_predictor(self.pretrained_model_id)
+
+        assert self.archive_file is not None
+        o = json.dumps(self.overrides) if self.overrides is not None else ""
+
+        if self.use_old_load_method:
+            from allennlp.models.archival import load_archive
+
+            o = json.dumps(self.overrides) if self.overrides is not None else ""
```
codeviking, line 96: I think this can be dropped, as it's a repeat of line 91.

It's also kind of odd that the empty state is `""`, when I imagine `overrides` is a `dict`, right? Maybe it should be `{}` or `None`?

I suppose it really depends on how it's handled in the APIs we pass it to, so do whatever they expect, which might be `""` :).

epwalsh: Good catch

api/allennlp_demo/common/http.py
```diff
@@ -114,21 +108,29 @@ def load_interpreters(self) -> Mapping[str, SaliencyInterpreter]:
...
-            "smooth_gradient": SmoothGradient(self.predictor),
-            "integrated_gradient": IntegratedGradient(self.predictor),
-        }
+        interpreters: Dict[str, SaliencyInterpreter] = {}
+        if self.model.interpreters is None or "simple_gradient" in self.model.interpreters:
+            interpreters["simple_gradient"] = SimpleGradient(self.predictor)
+        if self.model.interpreters is None or "smooth_gradient" in self.model.interpreters:
+            interpreters["smooth_gradient"] = SmoothGradient(self.predictor)
+        if self.model.interpreters is None or "integrated_gradient" in self.model.interpreters:
+            interpreters["integrated_gradient"] = IntegratedGradient(self.predictor)
```
codeviking, line 117: It looks like we now instantiate all three gradient types if there's no preference specified, otherwise we use only what they told us too.

Maybe we do something like:

```
interpreters = [ "simple_gradient", "smooth_gradient", "integrated_gradient" ] if self.interpreters is None else self.interpreters
```

Then we can leave the prior code unchanged (we just instantiate the gradients we need based on what's in `interpreters` instead of `self.interpreters`).

It's a small change, I think I just makes it a little easier to understand what's happening (and puts the default handling bit in one spot).  That said this is also a bit of a subjective comment, so feel free to ignore if you feel otherwise!

api/allennlp_demo/common/http.py
```diff
@@ -114,21 +108,29 @@ def load_interpreters(self) -> Mapping[str, SaliencyInterpreter]:
...
-        hotflip = Hotflip(self.predictor)
-        hotflip.initialize()
-        return {"hotflip": hotflip, "input_reduction": InputReduction(self.predictor)}
+        attackers: Dict[str, Attacker] = {}
+        if self.model.attackers is None or "hotflip" in self.model.attackers:
+            hotflip = Hotflip(self.predictor)
+            hotflip.initialize()
+            attackers["hotflip"] = hotflip
+        if self.model.attackers is None or "input_reduction" in self.model.attackers:
+            attackers["input_reduction"] = InputReduction(self.predictor)
```
codeviking, line 132: 👆  We could probably do something similar here too, if you like the idea I shared above.

dev/check_models_ci.py
```diff
@@ -0,0 +1,47 @@
...
+        assert model in tested_models, (
+            f"test for '{model}' model not found in {WORKFLOW_FILE_PATH}. "
+            f"Did you forget to add '{model}' to the 'Endpoint Test' model matrix?"
+        )
+    for model in tested_models:
+        assert model in all_models, (
+            f"'{model}' is in the GitHub Actions 'Endpoint Test' job, but does not "
+            f"appear to correspond to an actual demo model in 'api/allennlp_demo'. "
+            f"Did you forget to delete '{model}' from the 'Endpoint Test' model matrix?"
+        )
```
codeviking, line 43: These errors are great.

codeviking (approved):
LGTM!

.github/workflows/ci.yml
codeviking: Yes, that's great. I forgot about that doc!

api/allennlp_demo/common/http.py
epwalsh: That's much cleaner!

Issue #548, opened 2020-10-09
The VQA demo breaks when trying to upload a large image


jonborchardt:
its ok that we dont allow loading an image over 1MB
but, we should give a proper error message
AND it appears that when we get into this state, we must reload the page to fix it...

1- add a better error message
2- fix it so that the page is not broken requiring a reload

schmmd:
@jonborchardt we might make the limit 5 MB and add a better error message.

codeviking:
You may need to adjust this in two places:

1. Via an Ingress NGINX annotation, like [so](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size).

2. As an NGINX rule that you add to the local reverse proxy that sits in front of the service.

aimichal:
This isn't fixed completely, #571 and #569 were partial fixes.

aimichal:
Just confirmed that large images can be uploaded to https://demo.allennlp.org/visual-question-answering and prediction works.

I tried sizes of images up to 5,214,629 bytes, under the 5MB (5*1024*1024 byte) limit.

aimichal:
Confirmed nice error message appears when large file is uploaded:

> ![image](https://user-images.githubusercontent.com/19393950/96931024-82029780-1471-11eb-91a0-1cc7277a8eea.png)

aimichal:
> 2- fix it so that the page is not broken requiring a reload

I can't reproduce this situation; attempting to use a large image shows a nice error message, and I can use a smaller image after that and predictions work fine. I didn't have to reload anything.

There's another problem I found along the way (https://github.com/allenai/allennlp-demo/issues/572), maybe that's what someone ran into?

Anyway, I'm closing this now.

PR #569, opened 2020-10-21
parametrize body size, use large body size for permalinks

aimichal:
Large images are rejected (with HTTP 413 - "Payload Too Large") when when uploading them on the form at https://demo.allennlp.org/visual-question-answering. The rejection is happening on a POST to https://demo.allennlp.org/api/permalink/noop which is the permalink service

So this change makes that service accept uploads of 5MB (5242880 bytes).

I set this value manually on the ingress and confirmed that a 5,214,629 byte file is accepted, but a 5,315,752 file is rejected.

(This partially addresses https://github.com/allenai/allennlp-demo/issues/548)


epwalsh (approved):

codeviking (approved):
LGTM 👍

PR #575, opened 2020-10-22
show nice error when image is too large

aimichal:
This change makes a nice error appear when the image to be uploaded is too large.

Looks like this:

> image

This is another step towards fixing https://github.com/allenai/allennlp-demo/issues/548

### DIFF
--- a/ui/src/components/ImageParamControl.jsx
+++ b/ui/src/components/ImageParamControl.jsx
@@ -90,7 +90,19 @@ export const ImageParamControl = (props) => {
                 );
             }
         } else if (info.file.status === 'error') {
-            message.error(`${info.file.name} file upload failed: ${info.file.error.message}`);
+            // If the image is too large, then info.file.error.message will
+            // look like "cannot post api/permalink/noop 413'"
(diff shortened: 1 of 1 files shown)


codeviking (approved):
👍

ui/src/components/ImageParamControl.jsx
```diff
@@ -90,7 +90,19 @@ export const ImageParamControl = (props) => {
                 );
             }
         } else if (info.file.status === 'error') {
-            message.error(`${info.file.name} file upload failed: ${info.file.error.message}`);
+            // If the image is too large, then info.file.error.message will
```
jonborchardt, line 93: im actually confused we even get here see: https://github.com/allenai/allennlp-demo/blob/4a7586cb91bcc784a83a42ce9b00a5862ab95f39/ui/src/components/ImageParamControl.jsx#L44

aimichal: Yeah, I initially wrote the catch right before the call to `compressAndSubmit`, but discovered empirically that the `if (info.file.status === 'done')` condition was false, and we fell into the `else if` block on line 92.

jonborchardt (approved):

PR #576, opened 2020-10-24
Migrate AllenNLP to the latest build tools.

codeviking:
This PR updates the AllenNLP UI build to use the latest, canonical
setup that we've arrived at on ReViz.

Specifics:

- Add TypeScript support, which I validated via the new `/info` route.
  This route is probably something we can remove down the road once
  we have other TypeScript code in place.
- Remove `react-scripts` and `react-scripts-rewired` and use `webpack`
  instead.

One thing that's different still:

- Because the majority of the code is JavaScript and accordingly not
  compiled via `tsc`, we use Babel instead to transpile the JavaScript
  (and importantly, the JSX) to the appropriate execution target.

Things I still need to do:

- I've validated that things work locally by clicking around, but
  I'd like to deploy an adhoc of the UI to make sure things work
  in a deployed context. I plan on doing that first thing next
  week.

This work accomplishes the goal of this ticket:
https://github.com/allenai/reviz/issues/149.


ui/src/App.js
```diff
@@ -103,7 +105,7 @@ const Logo = styled.img.attrs({
 })``;
 
 const HeaderColumnsWithSpace = styled(HeaderColumns)`
-    padding: 6.5px 0;
+    padding: ${({ theme }) => theme.spacing.md} 0;
```
jonborchardt, line 108: `sm` is closer to 6.5

jonborchardt (approved):
awesome-possum

ui/src/App.js
codeviking: Yea, `sm` felt too tight:

Here's `md` (what it is now):
Screen Shot 2020-10-26 at 9 24 06 AM

Here's `sm`:
Screen Shot 2020-10-26 at 9 25 10 AM

The `md` looks better to me, lmk what you think.

jonborchardt: yep, you are right.... 
it actually should really be `lg`, since the white part of the header should stay 70-80px...
your call on `md` vs `lg`, but 6.5 was wrong for sure!

codeviking: I like where it's at now. We can tweak it after updating the latest n' greatest Varnish (which I think might take care of this for us).

Issue #572, opened 2020-10-22
Trailing slash breaks image uploads


aimichal:
The page https://demo.allennlp.org/visual-question-answering/ (with trailing slash) accepts an image and uploads it to https://demo.allennlp.org/visual-question-answering/api/permalink/noop which returns a 405 and this error appears:

![image](https://user-images.githubusercontent.com/19393950/96909897-8029db80-1453-11eb-8f95-ec9bf2c7c656.png)

However, the page https://demo.allennlp.org/visual-question-answering (no trailing slash) accepts an image and uploads it to https://demo.allennlp.org/api/permalink/noop

**The working page is linked from the navigation**, so this isn't a critical issue. I just stumbled upon .../visual-question-answering/ (with trailing slash) by manually typing it and completing with a trailing / out of habit.

Possible solutions:

* Return 404 on https://demo.allennlp.org/visual-question-answering/
* Redirect from https://demo.allennlp.org/visual-question-answering/ to https://demo.allennlp.org/visual-question-answering
* Make the page work regardless of the URL
* Maybe something even more obvious I can't think of now

(I suspect the root of the issue is that "api/..." is missing a leading "/".)

PR #593, opened 2020-11-05
Fix permalinks.

codeviking:
Permalinks have been broken since this change:
https://github.com/allenai/allennlp-demo/commit/bf3b1e989637e44c8236613ae7037a58c256579c

In that change I ported the UI to our newer build tools, but failed to
test permalinks (oops!). Permalinks were broken because the `<script />`
tag that's injected by `webpack` at build time was using a relative
path instead of an absolute one. This caused permalinks to fail
because instead of downloading `/main.[hash].js` the browser would
attempt to download `/[task-name]/main.[hash].js`, which doesn't
exist and actually causes `/index.html` to be served instead (which
clearly isn't a JavaScript document).

Don't worry, this time I actually tested with permalinks. They work
now :).

Oof. SPAs. So many footguns!

This fixes this issue:
https://github.com/allenai/reviz/issues/152

### DIFF
--- a/ui/public/index.html
+++ b/ui/public/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8" />
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/hierplane@0.0.9/dist/static/hierplane.min.css">
-    <link rel="manifest" href="https://demo.allennlp.org/manifest.json">
     <link rel="shortcut icon" href="https://demo.allennlp.org/favicon.ico">
     <title>AllenNLP - Demo</title>
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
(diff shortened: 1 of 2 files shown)


ui/public/index.html
```diff
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8" />
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/hierplane@0.0.9/dist/static/hierplane.min.css">
-    <link rel="manifest" href="https://demo.allennlp.org/manifest.json">
```
codeviking, line 6: This file isn't served anymore, so I'm removing it now, but it's not related to the fix.

aimichal (approved):
I wonder if this will also fix https://github.com/allenai/allennlp-demo/issues/572

codeviking:
> I wonder if this will also fix #572

Unfortunately not. I tested locally and the behavior still occurs.

PR #594, opened 2020-11-05
document using the permalinks svc and stop populating requester field

aimichal:
This addresses https://github.com/allenai/reviz/issues/153

Confirmed my IP was written to the requester column before this change, and not after this change.


codeviking (approved):
Looks great!

epwalsh (approved):
LGTM!

PR #595, opened 2020-11-06
Update model's usage to reflect the actual model file

AkshitaB:
The demo's usage text is outdated, and mentions the old tar.gz


schmmd (approved):

schmmd:
Thanks!  It sure is annoying how many places the model URL is stored in ;-P

PR #600, opened 2020-11-09
log request and response json

epwalsh:
This will be a huge help with debugging.


dirkgr (approved):
Looks OK, but will this make the logs huge?

epwalsh:
It shouldn't affect the readability of the logs in Google Cloud since these fields will be collapsed unless expanded by clicking on them.

image

epwalsh:
Not sure if there's any privacy concerns with this? @schmmd

schmmd:
We already log requests for permalinks, so I don't think there's a significant privacy issue here.  Particularly if they are not retained indefinitely.

epwalsh:
I'm going to go ahead and merge this now for the sake of debugging that weird issue we're seeing with next-token-lm.

codeviking (commented):
@dirkgr is right that these might be quite large in some cases. Did you test what this outputs for VQA? I think the image data might be base64-encoded and included in the request -- that might be a lot of data to log.

I imagine Google can handle routing, storing and displaying large log messages without issue. We do, however, pay Google per GB logged -- so this will unquestionably cost more. I'm not sure how much more, but it might be worth checking in in a week to see how this impacts our spend. It'll probably be fine, but if it isn't we could consider enabling it selectively (i.e. for a single task) as to reduce the verbosity.

PR #602, opened 2020-11-11
disable permalinks for vqa

aimichal:
The goal is to disable permalinks for VQA. See https://github.com/allenai/reviz/issues/151 for motivation.

I have no idea if this is the "right way" to do it, but I've renamed the disablePermadata variable to disablePermadataForRun, and distinguished it from a new model-level prop usePermalinksForModel. Then I set it usePermalinksForModel only in the VQA component to false.

Testing locally, http://localhost:8080/visual-question-answering doesn't trigger the permalink service (and the URL doesn't grow a permalink). OTOH, http://localhost:8080/reading-comprehension works as before.

Requesting http://localhost:8080/visual-question-answering/MzU= still results in a lookup of the permalink data for identifier `MzU=` though.


ui/src/components/Model.js
```diff
@@ -22,27 +22,28 @@ class Model extends React.Component {
...
     }
 
-    runModel(inputs, disablePermadata = false) {
-      const { selectedModel, apiUrl } = this.props
-
+    runModel(inputs, disablePermadataForRun = false) {
+      const { selectedModel, apiUrl } = this.props;
       this.setState({outputState: "working", interpretData: undefined, attackData: undefined});
 
+      const usePermalinks = permalinksDecider(this.props.usePermalinksForModel, disablePermadataForRun);
```
codeviking, line 37: Since `undefined` is "falsy" (in evaluates to `false`), you could call this `disablePermalinksForModel` and just do this:

```
const disablePermadata = disablePermadataForRun or this.props.disablePermalinksForModel
```

Then you'd swap `disablePermadata` back in for `!usePermalinks` and call it good, I think.

codeviking (approved):
This solves the problem, but I think could be simplified a bit, per my comment.

codeviking:
> Requesting http://localhost:8080/visual-question-answering/MzU= still results in a lookup of the permalink data for identifier MzU= though.

Right, and this is hard to fix, since the code for fetching permalinks is in `App.js`, which I think is the parent of various `Model.js` instances:
https://github.com/allenai/allennlp-demo/blob/master/ui/src/App.js#L14

Now, the various `modelComponents` are available, so you might be able to patch that function and do something like this:

```js
componentDidMount() {
   // Note: I'm not sure if this actually works (i.e. if props are accessible this way). You'd also probably want to be more
   // careful and make sure modelComponents[this.state.selectedModel] exists.
   if (!modelComponents[this.state.selectedModel].props.disablePermalinks) {
        /* Fetch data for permalink */
   }
}
```

Seems a little hacky, and could be done in a separate PR. Up to you if you want to give it a shot :).

ui/src/components/Model.js
aimichal: Yep, that is simpler. Thanks :-)

aimichal:
> Seems a little hacky, and could be done in a separate PR. Up to you if you want to give it a shot :).

I just had another idea. Will run it by you outside this PR.

ui/src/components/Model.js
```diff
@@ -22,27 +22,28 @@ class Model extends React.Component {
...
     }
 
-    runModel(inputs, disablePermadata = false) {
-      const { selectedModel, apiUrl } = this.props
-
+    runModel(inputs, disablePermalinksForRun = false) {
+      const { selectedModel, apiUrl } = this.props;
       this.setState({outputState: "working", interpretData: undefined, attackData: undefined});
 
+      const disablePermalinks = this.props.disablePermalinksForModel || disablePermalinksForRun;
```
codeviking, line 37: Should `disablePermalinksForRun` come first, so that it takes precedence?

aimichal: For conveying the intention, yeah, I can flip it around. But logically, there is no difference in order :-)

aimichal:
Yikes, looks like this will collide with https://github.com/allenai/allennlp-demo/compare/jon-upgrade-varnish.

@jonborchardt who merges first? This is ready to go, and it's urgent-ish, but it's also pretty surgical, so I don't mind re-applying it over your changes once yours hit master.

LMK or let's figure it out in Slack tomorrow.

jonborchardt:
> Yikes, looks like this will collide with https://github.com/allenai/allennlp-demo/compare/jon-upgrade-varnish.
> 
> @jonborchardt who merges first? This is ready to go, and it's urgent-ish, but it's also pretty surgical, so I don't mind re-applying it over your changes once yours hit master.
> 
> LMK or let's figure it out in Slack tomorrow.

feel free to merge and ill fix up on my side

PR #612, opened 2020-11-13
add model card data to /info when available

epwalsh:


api/allennlp_demo/common/config.py
```diff
@@ -69,6 +74,7 @@ def from_file(cls, path: str) -> "Model":
                 model_card = get_pretrained_models()[raw["pretrained_model_id"]]
                 raw["archive_file"] = model_card.archive_file
```
AkshitaB, line 75: Is there a reason to include archive_file and predictor_name twice? It'll be there in the model_card_data too.

AkshitaB (changes requested):

codeviking (approved):

api/allennlp_demo/common/config.py
epwalsh: Only because the `archive_file` field is required in this dataclass, but `model_card_data` might be `None`.

epwalsh: For `predictor_name`, some of our configs override this.

AkshitaB: Fwiw, we do test that all modelcards at least have an archive_file. It may not be worth it to change this dataclass and all the places it is used right now, but we should clean it up in the future when the demo is upgraded.

AkshitaB (approved):

api/allennlp_demo/common/config.py
epwalsh: I agree, that would be nice. Currently there are only a handful of demo models that don't have model cards, such as VilbertVQA (easy to add), NMN Drop (harder because this is maintained in a third-party repo, but personally I'm fine with removing this from the demo), and the semantic parsing models.

Not sure what to do about the semparse models, because I think we want to avoid having `allennlp-models` be a dependency of `allennlp-semparse`.

PR #623, opened 2020-11-16
initial tugboat work

jonborchardt:
should have no affect on prod

adhoc
https://allennlp-demo.jon-tugboat.apps.allenai.org/reading-comprehension
https://allennlp-demo.jon-tugboat.apps.allenai.org/hidden/reading-comprehension

all menu items go to /x
you can try /hidden/x to see new pages
(only /hidden/reading-comprehension works right now

NOTE:
the purpose is to agree on menu and demo file structure
next week we will agree on demo structure and tugboat files


ui/src/tugboat/shared.tsx
```diff
@@ -0,0 +1,63 @@
+import styled from 'styled-components';
+import { Button } from 'antd';
+
+export interface DemoConfig {
```
aimichal, line 4: Nice name :-)

ui/src/tugboat/shared.tsx
```diff
@@ -0,0 +1,63 @@
...
+    demoGroup: string;
+    title: string;
+    order: number;
+    disabled?: boolean;
+    status: 'active' | 'hidden' | 'disabled';
```
aimichal, line 9: What's the difference between `disabled: true` and `status: "disabled"`?

aimichal (commented):
I want to see this on an adhoc :-)

jonborchardt:
https://allennlp-demo.jon-tugboat.apps.allenai.org/reading-comprehension

https://allennlp-demo.jon-tugboat.apps.allenai.org/hidden/reading-comprehension

ui/src/tugboat/shared.tsx
jonborchardt: oops, ill delete `disabled:`
its old

jonborchardt: removed

ui/package.json
```diff
@@ -11,16 +11,19 @@
     "@babel/plugin-proposal-class-properties": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
+    "@types/commonmark": "^0.27.4",
```
codeviking, line 14: Do we still need this, if we're not using mdx?

ui/package.json
```diff
@@ -44,6 +47,7 @@
     "react-router-dom": "^5.0.1",
     "react-syntax-highlighter": "^15.3.0",
     "react-tooltip": "4.2.10",
+    "slug": "^4.0.2",
```
codeviking, line 50: What's this for?

ui/src/App.js
```diff
@@ -25,6 +25,15 @@ import '@allenai/varnish/dist/theme.css';
 
 const DEFAULT_PATH = '/reading-comprehension';
 
+// all routes in the menu
+let routes = [
+    ...demoMenuGroups.reduce((acc, c) => acc.concat(c.routes), []), // flatmap
+];
+// move any routes that are to be hidden to be prefixed by /hidden
+// this frees up the original url to open up any existing route of the same name not iun the menu
```
codeviking, line 33:
```suggestion
// this frees up the original url to open up any existing route of the same name not in the menu
```

ui/src/App.js
```diff
@@ -60,10 +81,10 @@ const App = () => (
 // This is the top-level demo component.
 // It handles the chrome for header and menus,
 // and it renders the specific task.
+// Note, this is older code that will be removed once the conversion ios complete
```
codeviking, line 84:
```suggestion
// Note, this is older code that will be removed once the conversion is complete
```

ui/package.json
jonborchardt: yes, if we were using mdx, we would NOT need commonmark
commonmark is used for sticking some md on the page (will be used for descriptions of tasks and models)

ui/package.json
jonborchardt: instead of having a hardcoded separate route per task, we use the task name as the route... but use slug to convert to a url

ui/src/App.js
```diff
@@ -25,6 +25,15 @@ import '@allenai/varnish/dist/theme.css';
...
 
+// all routes in the menu
+let routes = [
+    ...demoMenuGroups.reduce((acc, c) => acc.concat(c.routes), []), // flatmap
+];
+// move any routes that are to be hidden to be prefixed by /hidden
+// this frees up the original url to open up any existing route of the same name not iun the menu
+routes = routes.map((r) => {
+    return r.status === 'hidden' ? { ...r, path: `/hidden/${r.path}` } : r;
+});
```
codeviking, line 36: You fancy functional coder you! 

I think all of this could be done like so, too:

```
const routes = []
for (const g in demoMenuGroups) {
    for(const r in g.routes) {
         if (route.status === "hidden") {
             const hiddenRoute = { ...r, path: `/hidden/${r.path}` };
             routes.push(hiddenRoute);
             continue;
         }
         routes.push(r)
    }
}
```

Which I think is a little easy to read. Though admittedly my former functional self would've done what you did here. I'm regressing :)

ui/src/App.js
```diff
@@ -86,6 +107,30 @@ const Demo = (props) => {
...
+// and it renders the specific task.
+const DemoWrapper = (props) => {
+    return (
+        <Layout bgcolor="white">
+            <Header>
+                <HeaderColumnsWithSpace columns="auto 1fr">
+                    <Header.Logo href="http://www.allennlp.org/">
+                        <Logo width="147px" height="26px" alt="AllenNLP" />
+                    </Header.Logo>
+                </HeaderColumnsWithSpace>
```
codeviking, line 121: Doesn't Varnish do this for us now?

ui/src/components/Menu.js
```diff
@@ -3,16 +3,79 @@ import styled from 'styled-components';
...
+                    !demo.demoConfig.status !== 'disabled' &&
+                    ((!g.label && demo.demoConfig.demoGroup === 'Other') ||
+                        demo.demoConfig.demoGroup === g.label)
+            )
+            .sort((a, b) => a.demoConfig.order - b.demoConfig.order)
+            .map((demo) => {
+                return {
+                    path: demo.demoConfig.routePathOverride
+                        ? demo.demoConfig.routePathOverride
+                        : `${slug(demo.demoConfig.title, '-')}`,
```
codeviking, line 70: Ah, nice. So we convert the title to a URL friendly version if they don't provide one.

I might just call `routePathOverrite` `path` and make it optional:

```
/* The path used to access your demo. If it's not set we make a URL safe version of your title and use that instead.
    For instance a title of "My Cool Demo" would be accessible at "/my-cool-demo". */
```

ui/src/demos/ConstituencyParsing.tsx
```diff
@@ -0,0 +1,8 @@
+import { DemoConfig } from '../tugboat';
+
+export const demoConfig: DemoConfig = {
+    demoGroup: 'Annotate a sentence',
```
codeviking, line 4: Maybe this could be `group` for brevity?

ui/src/App.js
jonborchardt: at this point in my life, i disagree... but we can discuss more if you like
i find what i have easier and faster to understand...

ui/src/demos/ConstituencyParsing.tsx
```diff
@@ -0,0 +1,8 @@
+import { DemoConfig } from '../tugboat';
+
+export const demoConfig: DemoConfig = {
```
codeviking, line 3: Does it have to be named `demoConfig`? I wonder if this could just be a file instead:

```
ui/src/demos/constituency-parsing/config.ts
                                                           /Main.tsx
                                                           /...
```

More or less anticipating that some demos will be complicated enough to be broken out into multiple files, and having some commons structure for people to patten match off of helps avoid wild inconsistencies down the road.

ui/src/App.js
jonborchardt: good point, ill validate what we need here and fix it up

ui/src/components/Menu.js
jonborchardt: will update and comment
the intent is to actually remove routePathOverrite on as many as we can

ui/src/demos/ConstituencyParsing.tsx
jonborchardt: deal

ui/src/demos/ConstituencyParsing.tsx
jonborchardt: sure, i can go either way (at this point im not sure), ill break it out

ui/src/App.js
jonborchardt: header spacing is left to the client, since we dont know how tall the text/logo will be

ui/src/components/Menu.js
jonborchardt: done

ui/src/demos/ReadingComprehension.tsx
```diff
@@ -0,0 +1,111 @@
...
+            setModelId(modelCards[0].id);
+        }
+    }, [modelCards]);
+
+    React.useEffect(() => {
+        const found = modelCards.filter((m: ModelCard) => m.id === modelId);
+        if (found.length) {
+            setModel(found[0]);
+        }
+    }, [modelId]);
```
codeviking: These seem like they could all be combined:

```
interface ModelState {
   cards: ModelCard[],
   selectedModel?: Model
}

const [ modelState, setModelState ] = React.useState<ModelState>()

React.useEffect(() => {
   fetch('/api/info').then(r => r.json()).then(r => {
       const cards: getModelCards(...);
       const selectedModel = cards[0];
       setModelState({ cards, selectedModel })
   });
}, [ false ])
```

Then whatever code changes the selected model simply mutates that property...

This is simpler, and I think it's more efficient with respect to rendering. I'm pretty sure the individual `setState()` calls each enqueue a render -- so mutating the properties in isolation probably makes things a little slower.

ui/src/demos/ConstituencyParsing.tsx
jonborchardt: done

ui/src/demos/ConstituencyParsing.tsx
jonborchardt: done

ui/src/demos/ReadingComprehension.tsx
```diff
@@ -0,0 +1,111 @@
...
+                </Form.Field>
+                <Form.Field>
+                    <Form.Label>Passage</Form.Label>
+                    <Form.TextArea />
+                </Form.Field>
+                <Form.Field>
+                    <Form.Label>Question</Form.Label>
+                    <Form.Input />
+                </Form.Field>
+                <RunButton>Run Model</RunButton>
```
codeviking: This is great!

ui/src/tugboat/form.tsx
```diff
@@ -0,0 +1,55 @@
+import styled, { css } from 'styled-components';
+import { Input as AntInput, Select as AntSelect } from 'antd';
+
+export const Form = styled.form`
+    margin: ${({ theme }) => `0 0 ${theme.spacing.lg}`};
+    max-width: 36rem;
+`;
```
codeviking, line 7: Is there any reason to not use `antd`'s `<Form />` component?

It's pretty nice...and I think would be simpler for peopel to use and extend, since it's not custom.

ui/src/demos/ReadingComprehension.tsx
jonborchardt: im not going to do this in this pr.. i agree... in fact, i think that ill make a hook to do this... but im calling that out of scope for getting this up

ui/src/tugboat/form.tsx
jonborchardt: yes, we should. this page is not ready for review

ui/src/tugboat/shared.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const getModelCards = (data: ModelInfo[] | undefined, ids: string[]): ModelCard[] => {
+    const ret = (data || [])
+        .filter((m) => m.id && m.info && m.info.model_card_data && ids.includes(m.id))
+        .map((m) => {
+            m.info.model_card_data.id = m.id; // todo: remove this once we have real data
+            return m.info.model_card_data;
+        });
+    return ret;
+};
```
codeviking, line 49: I'd lift the model card stuff into its own file. It's feels a bit weird to have a bunch of css styles intermingled with this type of logic.

jonborchardt: will do... later

codeviking (approved):
This looks like a great start. None of my comments are blocking, as they're all things we can tend to as we iron this out further. Great work!

ui/src/demos/ReadingComprehension.tsx
codeviking: Ah, good call! I like the idea of a dedicated hook that we use.  `useModels(...)`!

PR #631, opened 2020-11-20
Rework the way we resolve the list of demos a bit.

codeviking:
This reworks the way we resolve the list of demos in a few ways:

1. Rather than importing the component and it's config from `Main.tsx`,
   we use the `index.ts` file for this. This is a pattern that's already
   well established in the community to it feels right to do. Note, if
   this PR is accepted I can rename all instances of `demoConfig`
   to `config`. I'll do this before merging if we like the direction.

2. Factor the code for loading demos into code that's clearly separated
   from the AllenNLP code. This way we have more of an idea of how
   an individual demo might use it.

3. Clearly document how things work and where the warts are. Knowledge
   is power.

4. Moved a few things out of `shared.tsx` into their own files. I like
   keeping the UI specific code and the models separate. In this case
   it felt like we were combining a few too many concerns into one
   file. Let me know what you think.

5. Wrote a hook for the model card info bit, and refactored the approach
   so it queries the info for the specific models that are requested
   rather than querying them all. In this case this reduces the size of
   the response from 32k to 7k, which seems worth it. There's of course
   overhead for additional connections, though the requests can and
   are processed in parallel -- my gut intuition is this will be faster,
   but I could be wrong. That said I like that it gives us a little
   flexibility, in that if we only need one model we only request it's
   data, rather than fetching them all.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -18,6 +18,7 @@
     "@types/react-router-dom": "^5.1.6",
     "@types/slug": "^0.9.1",
     "@types/styled-components": "^5.1.4",
+    "@types/webpack-env": "^1.15.3",
     "@typescript-eslint/eslint-plugin": "^4.5.0",
     "@typescript-eslint/parser": "^4.7.0",
     "antd": "^4.7.2",
@@ -61,5 +62,6 @@
     "build": "webpack --mode production",
     "lint": "eslint '**/*.{js,jsx,ts,tsx,json}' && echo '💫  Lint complete.'",
     "lint:fix": "eslint '**/*.{js,jsx,ts,tsx,json}' --fix && echo '🛠  Lint --fix complete'"
-  }
+  },
+  "devDependencies": {}
 }
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -4,13 +4,16 @@ import { BrowserRouter as Router, Route, Redirect, Switch } from 'react-router-d
 import { Content, Footer, Header, Layout, VarnishApp } from '@allenai/varnish/components';
 import { ScrollToTopOnPageChange } from '@allenai/varnish-react-router';
 
+import * as tugboat from './tugboat';
+
 import allenNlpLogo from './components/allennlp_logo.svg';
-import Menu, { demoMenuGroups } from './components/Menu';
+import Menu from './components/Menu';
 import ModelIntro from './components/ModelIntro';
 import { modelComponents, modelRedirects } from './models';
 import { PaneTop } from './components/Pane';
 import WaitingForPermalink from './components/WaitingForPermalink';
 import Info from './components/Info';
+import { groups } from './groups';
 
 import './css/App.css';
 import './css/fonts.css';
@@ -25,15 +28,12 @@ import '@allenai/varnish/dist/theme.css';
 
 const DEFAULT_PATH = '/reading-comprehension';
 
-// all routes in the menu
-let routes = [
-    ...demoMenuGroups.reduce((acc, c) => acc.concat(c.routes), []), // flatmap
-];
-// move any routes that are to be hidden to be prefixed by /hidden
-// this frees up the original url to open up any existing route of the same name not in the menu
-routes = routes.map((r) => {
-    return r.status === 'hidden' ? { ...r, path: `/hidden/${r.path}` } : r;
-});
+const demos = tugboat.Demos.load();
+const demosByGroup = groups.map((g) => ({
+    ...g,
+    demos: demos.forGroup(g.label),
+}));
+
 /*
 The App is just a react-router wrapped around the Demo component.
 The design is a bit convoluted so that the same code can run
@@ -60,14 +60,13 @@ const App = () => (
             <Switch>
                 <Route exact path="/" render={() => <Redirect to={DEFAULT_PATH} />} />
                 <Route path="/info" component={Info} />
-                {routes.map(({ path, exact, component: Component, componentProps }) => (
+                {demos.all().map(({ config, Component }) => (
                     <Route
-                        exact={exact}
-                        key={path}
-                        path={path}
+                        key={config.path}
+                        path={config.path}
                         render={(props) => (
                             <DemoWrapper>
-                                <Component {...props} {...componentProps} />
+                                <Component {...props} />
                             </DemoWrapper>
                         )}
                     />
@@ -95,7 +94,7 @@ const Demo = (props) => {
                 </HeaderColumnsWithSpace>
             </Header>
             <Layout>
-                <Menu redirectedModel={redirectedModel} />
+                <Menu items={demosByGroup} redirectedModel={redirectedModel} />
                 <Layout>
                     <FullSizeContent main>
                         <SingleTaskDemo model={redirectedModel} slug={slug} />
@@ -121,7 +120,7 @@ const DemoWrapper = (props) => {
                 </HeaderColumnsWithSpace>
             </Header>
             <Layout>
-                <Menu />
+                <Menu items={demosByGroup} />
                 <Layout>
                     <FullSizeContent main>{props.children}</FullSizeContent>
                     <Footer />
--- a/ui/src/components/Menu.js
+++ b/ui/src/components/Menu.js
@@ -3,81 +3,14 @@ import styled from 'styled-components';
 import { Menu as AntdMenu } from 'antd';
 import { textStyles, LeftSider } from '@allenai/varnish/components';
 import { Link } from '@allenai/varnish-react-router';
-import slug from 'slug';
 
 import { ImgIcon } from './ImgIcon';
-import annotateIcon from '../icons/annotate-14px.svg';
-import otherIcon from '../icons/other-14px.svg';
-import parseIcon from '../icons/parse-14px.svg';
-import passageIcon from '../icons/passage-14px.svg';
-import questionIcon from '../icons/question-14px.svg';
-import addIcon from '../icons/add-14px.svg';
 
 const { Item, SubMenu } = AntdMenu;
 
 /*******************************************************************************
   <Menu /> Component
 *******************************************************************************/
-
-// find all demo pages to show in menu
-const demoCtxs = require.context('../demos', true, /Main\.tsx$/);
-const demos = demoCtxs.keys().map(demoCtxs);
-
-// predefined order of known types
-const demoGroups = [
-    {
-        label: 'Answer a question',
-        iconSrc: questionIcon,
-    },
-    {
-        label: 'Annotate a sentence',
-        iconSrc: annotateIcon,
-    },
-    {
-        label: 'Annotate a passage',
-        iconSrc: passageIcon,
-    },
-    {
-        label: 'Semantic parsing',
-        iconSrc: parseIcon,
-    },
-    {
-        label: 'Other',
-        iconSrc: otherIcon,
-    },
-    {
-        label: 'Contributing',
-        iconSrc: addIcon,
-    },
-];
-
-export const demoMenuGroups = demoGroups
-    .map((g) => {
-        return {
-            label: g.label,
-            iconSrc: g.iconSrc,
-            routes: demos
-                .filter(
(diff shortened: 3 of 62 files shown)


ui/package.json
```diff
@@ -18,6 +18,7 @@
     "@types/react-router-dom": "^5.1.6",
     "@types/slug": "^0.9.1",
     "@types/styled-components": "^5.1.4",
+    "@types/webpack-env": "^1.15.3",
```
codeviking, line 21: This adds type information for `require.context()`, which was essential in understanding it.

ui/src/App.js
```diff
@@ -60,14 +60,13 @@ const App = () => (
             <Switch>
                 <Route exact path="/" render={() => <Redirect to={DEFAULT_PATH} />} />
                 <Route path="/info" component={Info} />
-                {routes.map(({ path, exact, component: Component, componentProps }) => (
+                {demos.all().map(({ config, Component }) => (
                     <Route
-                        exact={exact}
```
codeviking, line 65: This wasn't ever set, so this was always: `exact={undefined}`.

ui/src/App.js
```diff
@@ -60,14 +60,13 @@ const App = () => (
...
-                        key={path}
-                        path={path}
+                        key={config.path}
+                        path={config.path}
                         render={(props) => (
                             <DemoWrapper>
-                                <Component {...props} {...componentProps} />
```
codeviking, line 70: The `componentProps` were never set, so this was always `{...undefined}`.

ui/src/demos/masked-language-modeling/config.ts
```diff
@@ -3,8 +3,7 @@ import { DemoConfig } from '../../tugboat';
 export const demoConfig: DemoConfig = {
     group: 'Other',
     title: 'Masked Language Modeling',
-    path:
-        'masked-lm?text=The%20doctor%20ran%20to%20the%20emergency%20room%20to%20see%20%5BMASK%5D%20patient.',
+    path: '/masked-lm',
```
aimichal, line 6: Whoa, what was The%20doctor%20ran%20to%20the%20emergency%20room ?

ui/src/demos/masked-language-modeling/config.ts
```diff
@@ -3,8 +3,7 @@ import { DemoConfig } from '../../tugboat';
 export const demoConfig: DemoConfig = {
     group: 'Other',
     title: 'Masked Language Modeling',
-    path:
-        'masked-lm?text=The%20doctor%20ran%20to%20the%20emergency%20room%20to%20see%20%5BMASK%5D%20patient.',
```
codeviking, line 7: We probably shouldn't hard-code query parameters. It's too likely to break, when we change `text` to `t` in the file where that logic is defined.

I figure we can just implement behavior that uses a `<Redirect />` to set the default params when they aren't set...

ui/src/tugboat/DemoConfig.ts
```diff
@@ -0,0 +1,12 @@
...
+    group: string;
+    title: string;
+    order: number;
+    status: 'active' | 'hidden' | 'disabled';
+    /**
+     * The path used to access the demo. If it's not set we make a URL safe version of your
+     * title and use that instead.  For instance a title of "My Cool Demo" would be accessible
+     * at "/my-cool-demo".
+     */
+    path?: string;
```
aimichal, line 11: Hmm, is that a new feature we're adding or has it always been this way?

I'd like (in the future) if paths were always explicit and required, and never inferred from some arbitrary string.

Then we should simply verify that paths of various demos don't collide.

aimichal (commented):
Preliminary thoughts.

ui/src/tugboat/DemoConfig.ts
codeviking: It's the way it was, I'm just preserving existing behavior.

@jonborchardt implemented the original logic, so I'll let him chime in other whether we want to retain or change this.

ui/src/demos/masked-language-modeling/config.ts
codeviking: The language modeling demos complete a given sentence. Both demos start with some canned text so they're more compelling to begin with.

This merely removes them from the path, as they don't belong there. We can implement behavior in each demo that sets the appropriate demo, which prevents the need to hard-code some url encoded query parameters.  Doing it this way seems brittle.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -12,41 +13,30 @@ import {
...
-            setModel(found[0]);
-        }
-    }, [modelId]);
+    if (!models) {
+        return (
+            <Content>
+                <Spin indicator={<LoadingOutlined style={{ fontSize: '2rem' }} spin />} />
+            </Content>
+        );
+    }
```
codeviking, line 39: This is crucial. If we don't do this it's kind of frustrating to use on a slow connection, because the form's displayed without any content. It's also a bit startling when the UI suddenly refreshes with data. I tested by throttling by connection, and while this loading icon isn't perfect it felt ideal to not having any type of loading treatment at all.

ui/package.json
jonborchardt: nice!

ui/package.json
```diff
@@ -61,5 +62,6 @@
     "build": "webpack --mode production",
     "lint": "eslint '**/*.{js,jsx,ts,tsx,json}' && echo '💫  Lint complete.'",
     "lint:fix": "eslint '**/*.{js,jsx,ts,tsx,json}' --fix && echo '🛠  Lint --fix complete'"
-  }
+  },
+  "devDependencies": {}
```
jonborchardt, line 66: delete?

ui/src/App.js
jonborchardt: this came from other code where props are passed
i think we wil eventually do this.. but ok to cut for now

ui/src/demos/constituency-parsing/index.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * TODO: If we like the pattern proposed I'll:
```
jonborchardt, line 2: yes to all please!

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: can you pull out the style into a styled component?

ui/src/tugboat/DemoConfig.ts
jonborchardt: the intent was to make less config for users, but we could abandon this and jsut hardcode a path...
in the future we need a way to register a path so 2 tasks dont collide

ui/src/tugboat/Demos.ts
```diff
@@ -0,0 +1,92 @@
...
+ * below to `'../../../src/demos'`, assuming the module is published as `@allenai/varnish`, which
+ * means it'll be installed at `node_modules/allenai/varnish` and that `node_modules` lives
+ * at the root of their application.
+ *
+ * It's worth noting that npm and yarn tolerate moving where `node_modules` is, in which case
+ * the above mentioned mechanism will stop working. Maybe that's ok, given it's *very* unlikely
+ * one of our users will do this.
+ *
+ */
+const ctx = require.context('../demos', true, /\/index\.ts$/);
```
jonborchardt, line 32: lets just abandon this...
we will just have a hardcoded menu
and in the future, we can use a manifest supplied by users

ui/src/tugboat/Demos.ts
```diff
@@ -0,0 +1,92 @@
...
+            }
+            if (!('config' in d)) {
+                throw new ConfigError(`${path}/config.ts doesn't exist`);
+            }
+
+            const config: DemoConfig = d.config;
+
+            // If there isn't a path, we generate a URL safe version of the title and use that.
+            if (!config.path) {
+                config.path = `/${slug(config.title, '-')}`;
```
jonborchardt, line 79: and lets abandon this and use a hard coded path

ui/src/tugboat/Demos.ts
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            // We prefix demos that are hidden from view with a prefix that makes them harder to
+            // find. Someone can still access them, they're just less likely to.
+            if (config.status === 'hidden') {
+                config.path = `/hidden${config.path}`;
```
jonborchardt, line 85: [edited]
i get it now, this is fine
// i dont think this works (i have not run the code)
// we DONT want /hidden to show up in the menu

ui/src/tugboat/ModelCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface ModelInfo {
+    id: string;
+    model_card_data?: ModelCard;
+}
+
+export function fetchModelInfo(id: string): Promise<ModelInfo> {
+    return fetch(`/api/${id}`).then((r) => r.json() as Promise<ModelInfo>);
+}
+
+export function useModels(...ids: string[]): ModelInfo[] | undefined {
```
jonborchardt, line 48: so nice!

ui/src/components/Menu.js
```diff
@@ -115,13 +48,20 @@ export default class Menu extends React.Component {
...
-                                        <textStyles.Small>{m.title}</textStyles.Small>
-                                    </Link>
-                                </Item>
-                            ))}
+                            {g.demos.map((m) => {
+                                // TODO: Remove this, as it's an artifact of the incremental
+                                // transition. All demos are currently marked as hidden as
+                                // a mechanism for using the old code. Once we've fully
+                                // transition we won't need to do this anymore.
+                                const path = m.config.path.replace(/^\/hidden/, '');
```
jonborchardt, line 56: oic... this was confusing, but i get it now

jonborchardt (approved):
some work needed, but this is all in a great direction

ui/src/App.js
codeviking: Sure, it's easy to bring back if needed, and then we'll know what the shape looks like. Hopefully it's not an `{ [k: string]: any }`.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Sure.

ui/package.json
codeviking: `yarn` added it automatically, so I'll leave it in place (I think)?

ui/src/tugboat/Demos.ts
codeviking: Per our discussion we'll keep for now and let it brew. It's easy to use the hard-coded route down the road if we decide that's right.

ui/src/tugboat/Demos.ts
codeviking: Sure, ok.

ui/src/tugboat/Demos.ts
codeviking: Yea, I didn't really know where to put it.  Before the addition of `/hidden` was being handled when building up the routes, rather than when building the links.

Honestly I don't like it being in either location. The whole mechanism where if it's hidden it actually links to the old URL is pretty opaque.  But I also can't think of a better way without undoing the change we made to `Menu`, so I'll just leave it in it's new home for now (and the improvement being that it at least has a large comment clarifying it's intent).

ui/package.json
jonborchardt: k

ui/src/tugboat/Demos.ts
jonborchardt: yep, agreed

ui/src/tugboat/Demos.ts
codeviking: I think I'll take care of this in a separate PR, as right now there's a lot of changes in flight so I don't want to risk screwing something by tacking this on too. Let me know if you feel otherwise.

jonborchardt: deal, you can also leave it...
im ok either way

PR #639, opened 2020-11-25
Wire up the predict I/O for the new Reading Comprehension demo.

codeviking:
This wires up the I/O for Reading Comprehension predictions. There's
still a bunch of TODOs, but it at least gets the basic wiring in place.
Right now the output is just a big blob of JSON that's pretty printed.

Some things to note:

- Some of the types are a little incomplete, as I didn't dive too deeply
  into the backends as to understand the precise shape of each response.
  I figured we could do this while porting over the proper output
  treatments.
- I moved a lot of the model related code and models into `lib` from
  `tugboat`. This was an explicit decision because the models and
  API mechanics are AllenNLP specific. It might be that in the long
  run we anticipate forcing other similar applications to adopt the
  structure put forth by the AllenNLP demo, but at this point I don't
  feel like that assertion holds true. It's of course relatively easy
  to move this back when and if that situation arises.
- I mucked around with the shared form libraries a bit, as to use the
  `antd` "native" components wherever possible. This made it easier for
  me to get the form working with some semblance of validation n' such.


ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -47,58 +127,69 @@ export const Main = () => {
...
+                                key={m.info.id}
+                                value={m.info.id}
+                                label={m.info.model_card_data.display_name}>
+                                <b>{m.info.model_card_data.display_name}</b>
+                                <Markdown>{m.info.model_card_data.description}</Markdown>
+                            </Select.Option>
+                        ) : null
+                    )}
+                </Form.Select>
+            </Form.Field>
```
codeviking, line 156: I moved this, since it doesn't need to be a part of the data that we submit to the API, but rather changes the exact endpoint (and thereby model) that we invoke.

ui/src/tugboat/form.tsx
```diff
@@ -1,26 +1,24 @@
...
+export const Form = styled(AntForm).attrs(() => ({
+    layout: 'vertical',
+    requiredMark: true,
+}))`
     margin: ${({ theme }) => `0 0 ${theme.spacing.lg}`};
     max-width: 36rem;
 `;
 
-export const Field = styled.div`
+export const Field = styled(AntForm.Item)`
```
codeviking, line 12: Should we consider calling this `Item` instead of `Field` so it's consistent with `antd`'s API?

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -6,26 +6,84 @@ import React from 'react';
...
+}
+
+// TODO: Figure out what other types exist
+enum NAQANetAnswerType {
+    PassageSpan = 'passage_span',
+}
+
+interface NAQANetOutput {
+    answer: {
+        'answer-type': NAQANetAnswerType;
```
jonborchardt, line 49: yuck... we should change this key in the backend

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -6,26 +6,84 @@ import React from 'react';
...
+    question_id: string;
+    question_tokens: string[];
+}
+
+type Output = BiDAFOutput | TransformerQAOutput | NAQANetOutput;
+
 export const Main = () => {
-    const models = useModels('bidaf-elmo', 'bidaf', 'nmn', 'transformer-qa', 'naqanet');
-    const [selectedModel, setSelectedModel] = React.useState<ModelInfo>();
+    // TODO: NMN doesn't return anything right now (there's no `/info` route, I think), so it's
```
jonborchardt, line 67: ya, nmn is not in the model cards yet... but we will need it there before shipping

ui/src/lib/Model.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ModelInfo, fetchModelInfo } from './ModelInfo';
+
+const HTTP_HEADERS = { Accept: 'application/json', 'Content-Type': 'application/json' };
+
+interface Prediction<I, O> {
+    input: I;
+    output: O;
+}
+
+export class Model<I, O> {
```
jonborchardt, line 11: nice

jonborchardt (approved):
this is directional
ship it
(there is lots to do ... im worried about the length of Main.tsx)

codeviking:
> there is lots to do ... im worried about the length of Main.tsx

Yea, agree'd. I figure we'll arrive at generalizations as we write more code that makes it smaller and smaller. But the "Write less code in Main.tsx" mantra is a good north star for us to keep in mind.

PR #640, opened 2020-11-26
Generalize model selection a bit.

codeviking:
This factors things a bit so that model selection is captured
in the `useModels` hook rather than the task code itself. The
motivation being to reduce the amount of boilerplate that exists
in each demo.

To do this I modified `useModels` to return a class I call a
`ModelList`, which is just a wrapper for a list of models with
a reference to the selected one. This way demos can simply refer
to `list.selected`, rather than managing that value themselves.

The function also returns a second parameter, which is a function
that updates the selected model.

I'm still getting used to using React hooks, but this feels pretty
nice. Curious to see what y'all think.

### DIFF
--- a/ui/src/demos/reading-comprehension/Main.tsx
+++ b/ui/src/demos/reading-comprehension/Main.tsx
@@ -8,7 +8,7 @@ import { Content } from '@allenai/varnish/components';
 
 import { Form, Title, Description, Markdown, RunButton, Loading } from '../../tugboat';
 import { ModelUsageModal, ModelCardModal } from '../../components';
-import { useModels, Model } from '../../lib';
+import { useModels } from '../../lib';
 import { config } from './config';
 
 interface Input {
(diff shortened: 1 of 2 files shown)


ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -100,13 +87,13 @@ export const Main = () => {
     const handleFormSubmit = async (i: Input) => {
         // TODO: Show some form of loading.
         // TODO: Implement better error handling (show something nice to the user).
```
jonborchardt, line 89: I believe this whole method could live to model...

jonborchardt (approved):
Very nice

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Yea, I think you're right. I put that on my list of things to explore today.

ui/src/lib/Model.ts
```diff
@@ -22,12 +22,49 @@ export class Model<I, O> {
     }
 }
 
-export function useModels<I, O>(...ids: string[]): Model<I, O>[] | undefined {
-    const [models, setModels] = React.useState<Model<I, O>[]>();
+class ModelNotFoundError extends Error {
```
jonborchardt, line 25: nice

PR #644, opened 2020-12-03
More useModels factoring.

codeviking:
This PR commits additional factoring the `useModels()` API to encapsulate more functionality. After this change we also capture code related to requesting predictions from the currently selected model.

This took a bit to put together due to the nuance of `React.useReducer()`.  There's a lot of glue here, the ultimate goal of which is to make the API provided by `useModels()` as easy for the end user to use.  While I don't love all the complexity behind the scenes, I do like the resulting API that the user interacts with.  All of the logic related to handling race conditions created by asynchronous behavior are taken care of transparently, so the user has a simple interface for display purposes.

I snuck in a few extra things too, sorry for the scope creep:

1. I added a top-level `<ErrorBoundary />`. When things `throw` in the actual model code things are caught an ok error message is displayed. It's better than the prior treatment -- a blank page.
2. I added a `<ModelSelect />` control to `tugboat`. 

@jonborchardt @aimichal Please TAL.


ui/src/lib/Model.ts
```diff
@@ -28,22 +27,76 @@ class ModelNotFoundError extends Error {
...
+    selectModel(id: string): ModelList<I, O> {
+        const model = this.all.find((m) => m.info.id === id);
+        if (!model) {
             throw new ModelNotFoundError(id);
         }
-        return new ModelList(this.all, m);
+        return new ModelList(this.all, new SelectedModel(model));
+    }
+
+    updateSelectedModel(state: State) {
```
codeviking, line 93: This name needs rethinking.

ui/src/tugboat/index.ts
```diff
@@ -1,18 +1,4 @@
...
-// </Form>
-const Form = FormImpl as any; // any is so we can add the random statics below
-Form.Field = Field;
-Form.TextArea = TextArea;
-Form.Input = Input;
-Form.Select = Select;
-Form.OptDesc = OptDesc;
-
-export { Form };
-
```
codeviking, line 15: @jonborchardt I imagine we'll want to chat about this. This felt wrong to maintain, as it makes it tricky to change the things exported in `form.tsx`.

The existing acrobatics felt a little silly -- so I just moved things into a `form/` directory and export them in concert.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -66,44 +62,28 @@ type Output = BiDAFOutput | TransformerQAOutput | NAQANetOutput;
...
     // not present. We need to fix this.
-    const [models, selectModel] = useModels<Input, Output>(
-        'bidaf-elmo',
-        'bidaf',
-        'nmn',
-        'transformer-qa',
-        'naqanet'
-    );
-    const [output, setOutput] = React.useState<Output>();
+    const ml = useModels<Input, Output>('bidaf-elmo', 'bidaf', 'nmn', 'transformer-qa', 'naqanet');
```
codeviking, line 65: I used `ml` as a short form of `modelList`.  It might be a little confusing for people to pattern match off of. Let me know what you think.

ui/src/lib/store/models/error.ts
```diff
@@ -0,0 +1,26 @@
...
+    constructor(a: Action) {
+        super(`Unknown action: ${JSON.stringify(a)}`);
+    }
+}
+
+export class InvalidState extends Error {
+    constructor(actual: State) {
+        super(`Current state is ${actual.constructor.name}, which is unexpected`);
+    }
+}
```
codeviking, line 14: @codeviking You need to verify this tomorrow (I didn't test the tweak to print the constructor name, but I think it should work).

codeviking: This works.

ui/src/tugboat/form/ModelSelect.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Form>
+        <Field label="Model">
+            <Select
+                value={selected?.info.id}
+                onChange={(id) => {
+                    if (onChange === undefined) {
+                        return;
+                    }
+                    // TODO: This cast shouldn't be necessary.
+                    onChange(id as string);
```
codeviking, line 28: @jonborchardt Do you know why this is necessary? It seems `antd` sets `id` to `SelectValue`, and doesn't try to infer it from the anything?

ui/src/lib/store/models/Store.ts
```diff
@@ -0,0 +1,186 @@
...
+     * derived from React nomenclature: https://reactjs.org/docs/hooks-reference.html#usereducer
+     *
+     * If the action can't be processed because of a situation that's indicative of a programming
+     * error, an exception is thrown.
+     *
+     * In other cases the current state is returned with no change, and the effect of the action
+     * is discarded. This usually occurs due to asynchronous behavior -- for instance, when a
+     * response is received for a request that is no longer relevant to the UI.
+     */
+    static reducer<I, O>(cs: state.State, a: action.Action): state.State {
```
codeviking, line 115: This is the real meat n' potatoes.

The reducer processes actions and emits the resulting state (the current state being an argument of the function).

So we end up with a FSM, and this function acts to transition between states.  Each state has a distinct set of properties, and certain states inherit from prior ones (as when we're `Predicting` we're also `Loaded`, by necessity).

ui/src/lib/store/models/Store.ts
```diff
@@ -0,0 +1,186 @@
...
+            return new state.Predicting(cs.models, cs.selected, a.input);
+        }
+
+        if (a instanceof action.ReceivedPrediction) {
+            if (!(cs instanceof state.Predicting)) {
+                return cs;
+            }
+            if (cs.input !== a.prediction.input) {
+                return cs;
+            }
```
codeviking, line 170: This code handles a race. If the returned prediction isn't the one we're currently waiting on, we just return the current state.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -66,44 +62,32 @@ type Output = BiDAFOutput | TransformerQAOutput | NAQANetOutput;
...
-                return;
-            }
-            setOutput(p.output);
-        } catch (err) {
-            // TODO: Implement better error handling (show something nice to the user).
-            console.error(err);
-        }
-    };
+    if (!modelStore.hasLoadedModels()) {
+        const message = modelStore.failedToLoadModels()
```
jonborchardt, line 82: this logic should move to model store...
add a method called `AlertIfNoModels()`

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -114,61 +98,53 @@ export const Main = () => {
...
-                        <Markdown>{models.selected.info.model_card_data.description}</Markdown>
-                        <ModelUsageModal model={models.selected.info} />
-                        <ModelCardModal model={models.selected.info} />
-                    </>
-                ) : null}
-                <Form.Field label="Select an Example">
-                    <Form.Select placeholder="Examples..." />
-                </Form.Field>
-                <Form.Field
+            <form.ModelSelect
```
jonborchardt, line 101: ModelSelect should just take a modelStore rather than options, selected, and onChange

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -114,61 +98,53 @@ export const Main = () => {
...
+                options={modelStore.models}
+                selected={modelStore.selectedModel}
+                onChange={(id) => modelStore.selectModelById(id)}
+            />
+            <form.Form
```
jonborchardt, line 106: Form can also take a modelStore

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -114,61 +98,53 @@ export const Main = () => {
...
+                <form.Field>
+                    <form.Submit loading={modelStore.isPredicting()}>Run Model</form.Submit>
+                </form.Field>
+            </form.Form>
             <Divider />
-            {output ? (
-                <code>
-                    <pre>{JSON.stringify(output, null, 2)}</pre>
-                </code>
+            {modelStore.failedToPredict() ? (
```
jonborchardt, line 138: also should move to modelStore
`AlertIfPredictFailed()`

ui/src/tugboat/ErrorBoundary.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
jonborchardt, line 1: move to varnish... or cut an issue?

ui/src/tugboat/form/ModelSelect.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+// TODO: Right now the Model definition is AllenNLP specific, so it doesn't live in Tugboat.
```
jonborchardt, line 3: comment is old

ui/src/tugboat/form/ModelSelect.tsx
jonborchardt: ya, its some kind of bug...
i changed a similar thing in ai2-web as
` onChange={(v: SelectValue) => myFunc(v.toString())}`

ui/src/tugboat/index.ts
jonborchardt: seems fine... we can reexamine later if need be

jonborchardt (approved):
i like it!

ui/src/App.js
```diff
@@ -66,7 +66,9 @@ const App = () => (
                         path={config.path}
                         render={(props) => (
                             <DemoWrapper>
-                                <Component {...props} />
+                                <tugboat.ErrorBoundary>
```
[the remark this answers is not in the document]
codeviking: Agree'd. I'll cut an issue.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Ah, you mean, the logic for determining what message to be displayed. Good call.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Yea, I thought about that. I chose not to in the favor of composability. If we wanted to use the `<ModelSelect />` with a different type of store, (or no store at all!), we can.  If we bind it to a `modelStore` then the control is less flexible, and can't be as easily used in those context.

I realize I'm trying to predict the future in some sense, as we don't know if we're going to need to use the `<ModelSelect />` elsewhere. But it feels like a reasonable trade-off, as the code here isn't terribly complex (and I'd be ok repeating throughout the various demos).

I also like the existing implementation as it makes it easy for the user to understand the behavior bits.  For instance, this makes it very clear what happens when the user selects a model:

```
<form.ModelSelect onChange={(id) => store.selectModel(id)} />
```

Whereas this alternative hides that behavior. We have to step into `ModelSelect` to understand how it changes the `modelStore`:

```
<formModelSelect store={store} />
```

WDYT?

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Per above, I'm hesitant to do that. I think doing so risks inflexibility that will bite us later -- as there might be demos where we need to customize the behavior.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: I don't think the store should output anything. I think it should capture the relevant state and how it changes -- but any display layer logic (like outputting some HTML) should live elsewhere.

Does that make sense?

ui/src/tugboat/ErrorBoundary.tsx
codeviking: Sure, I'll cut an issue.

ui/src/tugboat/form/ModelSelect.tsx
codeviking: Err, no, it's still relevant.

All of the store related to models is in `src/lib`, not `src/tugboat`. This is because it's tightly coupled to AllenNLP's notion of what a model is (it's shape) and how to get that information (the `/api/:modelid` routes).

We need to figure out how to lift something general into Tugboat. Maybe it's some sort of adapter pattern, where the app using `tugboat` tells it how to retrieve the information for the models it contains (the API route to use, or possibly even just a hard-coded list, etc).]

Maybe that's something I take on next instead of `emory`? WDYT?

ui/src/tugboat/ErrorBoundary.tsx
codeviking: https://github.com/allenai/varnish/issues/364

ui/src/tugboat/form/ModelSelect.tsx
codeviking: Gotcha, so, in my case I only needed to cast it, so I think it's actually already a `string`. Turns out `string`s also have a `toString()`, so the route you went is ok too.

Thanks.

ui/src/tugboat/index.ts
codeviking: Sounds good.

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: AND popping the message
we should have as little code like this in Main as possible.

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: i like the later...
make Main as simple as possible
we can always support a the other variables as optional

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: i think some other file should have this code... but it does not need to be in the store

jonborchardt: `tugboat.AlertIfPredictFailed(modelStore)`

ui/src/tugboat/form/ModelSelect.tsx
jonborchardt: k
its up to you

ui/src/demos/reading-comprehension/Main.tsx
codeviking: There's no message that's popped. It's just some HTML that's displayed inline.

> we should have as little code like this in Main as possible.

I agree, to the extent that it doesn't damage the flexibility of each demo. The demos will necessarily derivate a bit from one another.  What looks good for one demo might not for another, etc.

So, the code we have here is minimal.  It's a few components.  I think it'd be ok for these to be duplicated throughout the individual demos. This way if one decides they want to make the alert green they can without unwinding some generalization that we've put in place.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Hmm. I think we're going a bit too far down the path of generalization. I worry we're headed towards an inflexible state -- and when we started this effort we mentioned that flexibility was key.

Here's a thoughtful article that makes the case for composition, rather than amassing overly complicated, high-level components:
https://epicreact.dev/soul-crushing-components/

ui/src/demos/reading-comprehension/Main.tsx
codeviking: What's so bad about it being here?  I like that I can:

1. Very easily change it, if I want to.
2. Easily follow what's displayed when (the "behavior" of the demo).

ui/src/App.js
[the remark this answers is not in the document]
codeviking: https://github.com/allenai/varnish/issues/364

ui/src/demos/reading-comprehension/Main.tsx
codeviking: @jonborchardt Please TAL, then let's sync in a Google Meet.

I consolidated the display logic, and the defaults. See `<ErrorMessage />` and its use here.

codeviking: We sync'd and came up with some ideas for further consolidation. The things we're building now are electrons, next we'll form some atoms and then...fission?

I'll merge as is and get some code out for us to review later that takes a stab at more generalization, but still gives us the freedom to compose things if / when we need to.  Thx @jonborchardt for the helpful feedback.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Per above, we've got an idea of how to form some components that connect these lower-level atoms to the larger entities they compose.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Yet another atoms will become molecules, and molecules will become organisms -- but we're still building atoms now.

ui/src/tugboat/form/ModelSelect.tsx
codeviking: We chatted, our takeaway is that:

1. The model card and it's structure is something we assert as something we want all demos to adopt, so we can safely generalize that structure (though it might, of course, change).

2. The way of getting it (i.e. querying it from an endpoint) should be AllenNLP specific, since we don't know how that information will be retrieved across all of our demos.

3. The endpoint for a given model needs to be demo specific, but the store that captures it's output should be generalized into Tugboat as some sort of "invoker" pattern.

So more work soon on this o'er the next week.

PR #647, opened 2020-12-05
Rework how we handle querying for predictions, take two.

codeviking:
This reworks the way I had previously factored code for querying predictions per our discussion on Friday.  At a high level this change tries to capture all code related to state management in generalizations, so that the `<Main />` component can be as succinct as possible. I also moved more code into `tugboat` and figured out way of accommodating the AllenNLP specific details in a way that should work for other demos too.

Here's a high level overview of the changes:

- I partitioned the files in `tugboat` into `lib` (data and logic), `components`, `context` and `error`.
- I moved `Model` and `ModelCard` into `tugboat` but left `ModelInfo` (which is AllenNLP specific in `lib`).
- I added a top-level components called `<MultiModelDemo />` in tugboat. This component sets up a `Models` context (which is also part of tugboat). That context keeps a list of models and the one that's currently selected, so that other TugBoat components can retrieve that information, regardless of their position in the DOM.  This lets this information transparently flow to the various tugboat components, regardless of where the user puts them in their demo.
- I added an AllenNLP `<MultiModelDemo />` which wraps the tugboat one with some glue that accounts for the process of querying `ModelInfo` instances from the API and converting them to `Model`s.
- I redacted my change that modified `fetchModelInfo` to query individual API endpoints per model, as connections are expensive and even a 3x size increase probably isn't worth the extra time spent negotiating sockets.
- I captured the whole state-machine notion in two new ideas -- `useAsync` and `<AsyncOutput />`.  The `useAsync` API is a hook that sets up a state machine that captures the state of an asynchronously retrieved thing.  The `<AsyncOutput />` component uses `useAsync` and handles rendering a `<Loading />` indicator while things are being fetched, an `<ErrorMessage />` if things failed and otherwise passes the retrieved thing on to be rendered by the component's children.

Despite my initial grumblings about over generalization I really like where this ended up. Thanks @jonborchardt for nudging me in this direction.

I'm really curious to hear what y'all think of the `useAsync` bit. It handles things that are queried asynchronously in a generic way that should be easy to apply elsewhere. It takes care of the race conditions that exist in this scenario -- which usually require some special-casing to handle. If we like how it came together we might consider publishing it as an OSS module that we can use throughout the things we're working on.


ui/src/components/ModelInfo.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import { AsyncOutput } from '../tugboat/components';
+
+import { ModelInfo as MInfo, fetchModelInfo } from '../lib';
+
+interface Props {
+    /**
```
aimichal, line 8: test123

aimichal: This PR-in-IDE thing is gonna take some getting used to.

ui/src/tugboat/components/AsyncOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * React components visualizing it.
+ *
+ * @example
+ *  <AsyncOutput<Input, Output>
+ *      input={"modelId"}
+ *      fn={fetchModelInfo}>{ (output) => (
+ *      <FancyOutputDisplay output={output} />
+ *  )}</AsyncOutput>
+ */
+export const AsyncOutput = <I, O>({
```
aimichal, line 29: Consider s/Async/Inflight/ so it's unique to this codebase, and not confusable with js's async.

aimichal:
> I'm really curious to hear what y'all think of the `useAsync` bit.

It's taking me longer than I expected to warm up to this. That we had a long conversation where I stumbled through the ideas doesn't help :-) So it's probably just me, because it sounds like this approach fits nicely with how large react apps are *supposed to* be built.

But I can't help remembering the days when I could write simple components with internal state that was updated by an async operation that waited for a response from a server. It felt so neat and contained.

So, at a very high level, do we need this generalized state machine (uninitialized->loading->success/failure) because we'll have *so many* of such individually simple components that the copypasta will be too hard to maintain and too error prone?

codeviking:
> So, at a very high level, do we need this generalized state machine (uninitialized->loading->success/failure) because we'll have so many of such individually simple components that the copypasta will be too hard to maintain and too error prone?

That's part of the motivation.

The other is that when we implement that pattern, we usually do it without properly handling the possibility of a race.  For instance, if we do something like this:

```
const [ state, setState ] = useState();
return (
   <>
     <Form>
           <Form.Input name="q">
          <Button onClick={() => {
                doSomeAsyncThing().then(r => setState(r));
           }}>
     </Form>
      {output && <Viz output={state} />}
   </>
);
```

...it works most of the time, but there's a race that can be really confusing, and even cause the end user to be misled.

If the backend is slow (like a lot of ours are) and the form doesn't prevent me from submitting something while something else is being processed, then we can end up with multiple in-flight requests at once, with no guarantee about which will be displayed. 

If we don't try to handle this in the `then()` callback, by checking that the input associated with the returned response is the most recent input submitted, we might display the wrong output to the user. So, we need code like so:

```
const [ state, setState ] = useState();
return (
   <>
     <Form>
           <Form.Input name="q">
          <Button onClick={() => {
                doSomeAsyncThing().then(r => {
                      if (r.input === currentInput) {
                           setState(r));
                      }
                })
           }}>
     </Form>
      {output && <Viz output={state} />}
   </>
);

Most of the time we don't do this, and rely on the fact that this race probably doesn't occur very often (that we know of!). 

What I like about the `useAsync` approach is it takes care of this transparently.

**Comment by codeviking** [comment]
I should say, another option is to opt-out of functional components. If we did this we could rely on the `setState()` method of `React.Component`. Importantly, you can pass `setState()` a callback to ensure that the values expressed in `state` are up-to-date.  This mechanism avoids the complexity of the `useReducer` API.

File: ui/src/tugboat/components/SelectModel.tsx [hunk 3]
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
```
jonborchardt, line 2: needs a nice 'sams' description comment ;-P

ui/src/tugboat/components/form/Form.tsx
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    // We do our best to make sure the children match the format we expect. That said we can't
+    // actually make sure that they're the specific compoent type's we expect -- or at least, I
+    // wasn't able to figure out how to do that.
+    const children = React.Children.toArray(props.children);
+    if (!children || children.length !== 2) {
+        throw new InvalidPredictChildrenError();
+    }
+    const [firstChild, secondChild] = React.Children.toArray(children);
```
jonborchardt, line 62: nice

ui/src/tugboat/components/form/Form.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if (!React.isValidElement(secondChild)) {
+        throw new InvalidPredictChildrenError();
+    }
+
+    return (
+        <>
+            <FormElement
+                onFinish={(i) => {
+                    setInput(i as I);
```
jonborchardt, line 74: can you just do `(i:I) =>` on line 73?

ui/src/tugboat/components/form/Passage.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
jonborchardt, line 1: love it... needs a comment on when to use

ui/src/tugboat/components/form/Question.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
jonborchardt, line 1: needs comment

jonborchardt (approved):
this is great!!!
left some thoughts

ui/src/App.js
```diff
@@ -28,7 +29,7 @@ import '@allenai/varnish/dist/theme.css';
 
 const DEFAULT_PATH = '/reading-comprehension';
 
-const demos = tugboat.Demos.load();
+const demos = Demos.load();
```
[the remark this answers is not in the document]
codeviking: Yea, I'm not sure.

Maybe `Demos.buildListFromFilesystem()` or `Demos.loadFromFilesystem()`?

ui/src/components/ModelInfo.tsx
```diff
@@ -0,0 +1,20 @@
...
+     * If set, only info for the specified models will be returned. If unset information for
+     * all models will be returned.
+     */
+    ids?: string[];
+    children: (info: MInfo[]) => React.ReactNode;
+}
+
+export const ModelInfo = ({ ids, children }: Props) => (
+    <AsyncOutput<string[] | undefined, MInfo[]> input={ids} fn={fetchModelInfo}>
```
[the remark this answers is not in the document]
codeviking: Good call.

ui/src/components/MultiModelDemo.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * of a model (which is queried via API routes) to the shape expected by the tugboat package.
+ */
+export const MultiModelDemo = ({ ids, children }: Props) => (
+    <ModelInfo ids={ids}>
+        {(info) => {
+            const models = [];
+            for (const i of info) {
+                if (!i.model_card_data) {
+                    console.warn(
+                        `Model ${i.id} won't be displayed because it doesn't have a model card.`
```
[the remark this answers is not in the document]
codeviking: We could log an event, yea. We'd have to look for it thought.

I think in most cases this will get caught while the person is building out the demo. It'll be pretty obvious when their model isn't available as thing to select :).

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -1,146 +1,45 @@
...
+            <SelectModel />
+            <ModelCardModal />
+            <ModelUsageModal />
+            <Predict<Input, Prediction>>
+                <Fields>
+                    <Passage />
+                    <Question />
+                    <Submit>Run Model</Submit>
+                </Fields>
+                <Output>{(p: Prediction) => <pre>{JSON.stringify(p, null, 2)}</pre>}</Output>
```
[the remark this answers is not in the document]
codeviking: Gotcha, yea. I can do that.

ui/src/tugboat/components/AsyncOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+    input,
+    fn,
+    children,
+    errorMessage,
+}: AsyncOutputProps<I, O>) => {
+    const output = useAsync(fn, input);
+
+    if (output.isLoading() || output.isUnitialized()) {
+        return <Loading />;
```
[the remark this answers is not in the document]
codeviking: Yea, that'd be nice.

ui/src/tugboat/components/SelectModel.tsx
codeviking: Be careful what you wish for... 😉

ui/src/tugboat/components/form/Form.tsx
codeviking: I'll try that.

ui/src/tugboat/lib/useAsync.ts
```diff
@@ -0,0 +1,99 @@
...
+    isSuccess(): this is Success<I, O> {
+        return this instanceof Success;
+    }
+
+    isFailure(): this is Failure<I, O> {
+        return this instanceof Failure;
+    }
+}
+class Uninitialized<I, O> extends AsyncState<I, O> {
+    readonly isEmpty = true;
```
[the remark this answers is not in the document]
codeviking: Done, thanks.

ui/src/demos/reading-comprehension/config.ts
```diff
@@ -1,8 +1,9 @@
-import { DemoConfig } from '../../tugboat';
+import { DemoConfig } from '../../tugboat/lib';
 
 export const config: DemoConfig = {
     group: 'Answer a question',
     title: 'Reading Comprehension',
     order: 1,
+    modelIds: ['bidaf-elmo', 'bidaf', 'nmn', 'transformer-qa', 'naqanet'],
```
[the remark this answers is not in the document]
codeviking: I like this, but I think I'll let this creep in during a future PR since there's already a lot changing here.

ui/src/tugboat/lib/useAsync.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (action.desiredState instanceof Success || action.desiredState instanceof Failure) {
+            // We need to make sure the desired state is still relevant. If we're not currently
+            // loading something, then we shouldn't transition to the new state.
+            if (!(currentState instanceof Loading)) {
+                return currentState;
+            }
+
+            // ...and if whatever we loaded (or failed to load) isn't what we're currently waiting
+            // on, then drop the desired state on the floor.
+            if (currentState.input !== action.desiredState.input) {
```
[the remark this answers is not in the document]
codeviking: So, I did this, and then reverted the change because:

- Debugging was a little odd, in that instead of inspecting `input` that had some shape I expected, I'd be looking at a number I didn't generate.
- If I resubmit the same input several times, the first request to respond "wins" and is displayed (so the user gets output faster). Now, if the backend isn't idempotent then this technically might display the wrong thing...but that feels like a bit of an edge case.

WDYT? Let me know if there's another reason to use a counter.

ui/src/tugboat/components/AsyncOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A generic component that's responsible for rendering UI states associated with a resource
+ * that's obtained asynchronously. While the resource is being fetched a `<Loading />` indicator
+ * is rendered, and when things fail an `<ErrorMessage />` is rendered.
+ *
+ * The component expects a single child which is a function that takes the output and returns
+ * React components visualizing it.
+ *
+ * @example
+ *  <AsyncOutput<Input, Output>
```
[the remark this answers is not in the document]
codeviking: I think I did what you're looking for. TAL.

ui/src/tugboat/components/AsyncOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    if (output.isLoading() || output.isUnitialized()) {
+        return <Loading />;
+    }
+
+    if (output.isFailure() || !output.isSuccess()) {
+        return <ErrorMessage message={errorMessage} />;
+    }
+
+    return <>{children(output.output)}</>;
```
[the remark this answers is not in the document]
codeviking: I asked him to write this while we were reviewing the code. Ideally on line 37 we should be able to omit the `!ouput.isSuccess()` bit:

```
if (output.isFailure() || !output.isSuccess()) {P
  return <ErrorMessage message={errorMessage} />
}
```

ui/src/tugboat/components/AsyncOutput.tsx
codeviking: I went with `usePromise` and `<Promised />`. 

We can tweak it later too :). It's a hard thing to name.

ui/src/tugboat/components/form/Form.tsx
codeviking: The compiler gets angry:

```
ui_1          | [tsl] ERROR in /ui/src/tugboat/components/form/Form.tsx(73,17)
ui_1          |       TS2769: No overload matches this call.
ui_1          |   Overload 1 of 2, '(props: Pick<Pick<FormProps<unknown> & { children?: ReactNode; } & { ref?: ((instance: FormInstance<unknown> | null) => void) | RefObject<FormInstance<unknown>> | null | undefined; } & { ...; }, "children" | ... 279 more ... | "requiredMark"> & Partial<...>, "children" | ... 281 more ... | "hideRequiredMark"> & { ...; } & { ...; }): ReactElement<...>', gave the following error.
ui_1          |     Type '(i: I) => void' is not assignable to type '(values: unknown) => void'.
ui_1          |       Types of parameters 'i' and 'values' are incompatible.
ui_1          |         Type 'unknown' is not assignable to type 'I'.
ui_1          |           'I' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.
ui_1          |   Overload 2 of 2, '(props: StyledComponentPropsWithAs<FormInterface, any, { layout: "vertical"; hideRequiredMark: true; }, "layout" | "hideRequiredMark">): ReactElement<...>', gave the following error.
ui_1          |     Type '(i: I) => void' is not assignable to type '(values: unknown) => void'.
```

I think `antd` kind of ties our hands here, unfortunately.

ui/src/components/ModelInfo.tsx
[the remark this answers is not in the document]
codeviking: Done, I just called it `fetch`. I think this looks nice when it's used:

```
<Promised input={i} fetch={...} />
```

Let me know what you think.

ui/src/tugboat/components/form/Form.tsx
codeviking: Oh, wait.

I think the key is this bit: `<FormProps<Unknown>>`, so maybe I need to pass a type to `FormElement`?

codeviking: Ok, so, I couldn't get it to work with `styled`. Maybe we can do that later.

But `antd.Form` does take a `type` that fixes this:

```
<antd.Form<Input> onFinish={(input) => /* input instanceof Input == true */} />
```

I wonder if `antd.Select` is similar?

codeviking: Hmm. I couldn't get it to work with `Select`.

ui/src/tugboat/components/form/Question.tsx
codeviking: Done.

ui/src/tugboat/components/form/Passage.tsx
codeviking: Done.

ui/src/tugboat/components/form/Form.tsx
codeviking: Anyway, the immediate issue is handled. Thanks!

ui/src/tugboat/components/SelectModel.tsx
codeviking: Fixed.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -1,146 +1,45 @@
...
-        );
-    }
-
-    return (
-        <Content>
+export const Main = () => (
+    <Content>
+        <MultiModelDemo ids={config.modelIds}>
             <Title>{config.title}</Title>
+            {/* TODO: It might be nice to put the description in the config too, or put it in
```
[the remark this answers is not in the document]
codeviking: Cool, we'll get there. I'll leave the `TODO` in place to remind us :).

ui/src/App.js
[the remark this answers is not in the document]
codeviking: I'm going to leave as-is for now, but we can rev on the name when something comes to us that's better!

ui/src/demos/reading-comprehension/types.ts
```diff
@@ -0,0 +1,51 @@
+export interface Input {
```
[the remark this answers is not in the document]
codeviking: Yup, we'll get there.  Especially for demos for the same task I imagine.

ui/src/tugboat/components/AsyncOutput.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import { useAsync } from '../lib';
+import { Loading } from './shared';
+import { ErrorMessage } from './ErrorMessage';
+
+interface AsyncOutputProps<I, O> {
+    input: I;
+    fn: (i: I) => Promise<O>;
```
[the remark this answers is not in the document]
codeviking: Fixed.

ui/src/tugboat/components/AsyncOutput.tsx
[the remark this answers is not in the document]
codeviking: Ok, I revised this a bit.

The compiler seems to think there's a potential `state` that might exist that isn't one of the known states. I think it might be because the base class could in theory be extended elsewhere, as there's no notion of a `sealed` class? That doesn't really make sense though, as the compiler should be able to see all derivations of the abstract class that's used to define the return type.  ¯\_(ツ)_/¯

Anyway, I'm not going to worry too much about it. I revised the code so that it's clear that if we reach that state that it's some sort of programming error, at which point we could dig in more.

ui/src/demos/reading-comprehension/Main.tsx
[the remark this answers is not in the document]
codeviking: Done. I didn't add a context object for doing this -- rather maintained the existing paradigm, where the end user passes the output to the component that's used.

Per our discussion today -- I'm not sure if we'll end up hiding this portion of the data-flow from the user or not. It seems like the spot they're more likely to customize, which is why I hesitated to do so right away.

PR #658, opened 2020-12-15
Add a new endpoint for listing taskcards.

codeviking:
This adds two new endpoints:
- `/api/tasks ~> Dict[str, TaskCard]`. This is effectively a HTTP wrapper for `get_tasks()`.
- `/api/task/:task_id ~> Optional[TaskCard]`. This fetches a single task card by it's id.

We plan on using this in TugBoat (the new demo UI architecture) for powering the examples
that we display in the UI.

### DIFF
--- a/api/allennlp_demo/tasks/api.py
+++ b/api/allennlp_demo/tasks/api.py
@@ -21,7 +21,7 @@ class TasksService(flask.Flask):
         def info():
             return flask.jsonify({"id": "tasks", "allennlp_models": VERSION})
 
-        @self.route("/tasks", methods=["GET"])
+        @self.route("/all", methods=["GET"])
         def tasks():
             tasks = get_tasks()
             return flask.jsonify(tasks)
--- a/api/allennlp_demo/tasks/api.py
+++ b/api/allennlp_demo/tasks/api.py
@@ -1,12 +1,12 @@
 """
 The tasks endpoint lists all demo tasks and some info about them.
 """
-import logging  # noqa: E402
-import flask  # noqa: E402
+import logging
+import flask
 from werkzeug.exceptions import NotFound
 
-from allennlp_demo.common.logs import configure_logging  # noqa: E402
-from allennlp_models.pretrained import get_tasks  # noqa: E402
+from allennlp_demo.common.logs import configure_logging
+from allennlp_models.pretrained import get_tasks
 from allennlp_models.version import VERSION
 
 logger = logging.getLogger(__name__)
--- a/api/allennlp_demo/tasks/api.py
+++ b/api/allennlp_demo/tasks/api.py
@@ -1,0 +1,30 @@
+"""
+The tasks endpoint lists all demo tasks and some info about them.
+"""
+import logging  # noqa: E402
+import flask  # noqa: E402
+
+from allennlp_demo.common.logs import configure_logging  # noqa: E402
+from allennlp_models.pretrained import get_tasks # noqa: E402
+from allennlp_models.version import VERSION
+
+logger = logging.getLogger(__name__)
+
+class TasksService(flask.Flask):
+    def __init__(self, name: str = "info"):
+        super().__init__(name)
+        configure_logging(self)
+
+        @self.route("/", methods=["GET"])
+        def info():
+            return flask.jsonify({ "id": "tasks", "allennlp_models": VERSION })
+
+        @self.route("/tasks", methods=["GET"])
+        def tasks():
+            tasks = get_tasks()
+            print(tasks)
+            return flask.jsonify(tasks)
+
+if __name__ == "__main__":
+    app = TasksService()
+    app.run(host="0.0.0.0", port=8000)
--- a/api/allennlp_demo/tasks/api.py
+++ b/api/allennlp_demo/tasks/api.py
@@ -5,10 +5,8 @@ import logging  # noqa: E402
 import flask  # noqa: E402
 from werkzeug.exceptions import NotFound
 
-from typing import List, Dict  # noqa: E402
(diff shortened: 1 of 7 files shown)


dev/check_models_ci.py
```diff
@@ -36,6 +48,8 @@ def main():
             f"Did you forget to add '{model}' to the 'Endpoint Test' model matrix?"
         )
     for model in tested_models:
+        if model in NON_MODEL_ENDPOINTS:
+            continue
         assert model in all_models, (
             f"'{model}' is in the GitHub Actions 'Endpoint Test' job, but does not "
             f"appear to correspond to an actual demo model in 'api/allennlp_demo'. "
```
codeviking, line 55: @epwalsh @AkshitaB @dirkgr Let me know if these changes aren't ideal -- I wanted to get to a state where we can stop force-merging without waiting for the VQA fixes to land.  Given this might happen again in the future this felt like an ok change.

codeviking:
I should also note that when I start the `tasks` HTTP server I see an alert like this:

```
/usr/local/lib/python3.8/site-packages/torch/cuda/__init__.py:52: UserWarning: CUDA initialization: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx (Triggered internally at  /pytorch/c10/cuda/CUDAFunctions.cpp:100.)
```

I'm assuming `cuda` is now a dependency of `allennlp` and/or `allennlp_models`, so this is getting included in the container due to that dependency. If that's not the case let me know.

codeviking:
Ok, this is finally passing CI. I'll stop making changes in flight so it can be reviewed. Apologies for the noise.

AkshitaB (approved):
LGTM! I'll wait for @dirkgr and @epwalsh for approving the VQA stuff.

api/allennlp_demo/tasks/api.py
```diff
@@ -0,0 +1,39 @@
+"""
```
jonborchardt, line 1: do we need __init__.py ?

api/allennlp_demo/tasks/api.py
```diff
@@ -0,0 +1,39 @@
...
+The tasks endpoint lists all demo tasks and some info about them.
+"""
+import logging  # noqa: E402
```
jonborchardt, line 4: what is `# noqa: E402`

dev/check_models_ci.py
```diff
@@ -4,11 +4,17 @@
...
 import os
+import logging
 from typing import Iterable
 
-
 WORKFLOW_FILE_PATH = ".github/workflows/ci.yml"
+logging.basicConfig()
+
+# Sometimes models are broken, so we allow their tests to be temporarily omitted.
+SKIPPED = set([ "vilbert_vqa" ])
```
jonborchardt, line 14: interesting...
i wonder if instead of listing the skips here, you could still fetch all tasks, and check an attribute in the response if it should be skipped?
for example, we may want to allow 'skipped' models to be seen in the demo when in debug mode...
this would centralize 'skip/ignore/hide' so that it does not need to be specified in multiple places....

jonborchardt (approved):
lg2m, i left some thoughts...

api/allennlp_demo/tasks/api.py
codeviking: Yup, it tells Python that it's a package / module with code that can be imported elsewhere.

See https://docs.python.org/3/tutorial/modules.html#packages.

api/allennlp_demo/tasks/api.py
codeviking: Y'know, I'm not sure. I copied it because it was in existing code.

Looks like it's a lint override that tells the linter to not get angry when an import isn't at the top of the file:
https://www.flake8rules.com/rules/E402.html

So it might not be needed here. I'm going to remove it and see if the linter gets mad. Thanks for asking the question!

dev/check_models_ci.py
codeviking: In this case we're just disabling CI -- it'll still be present in the demo.

We're disabling CI because it doesn't build right now -- which means the VQA endpoint won't deploy, but others will.

So I feel like disabling the demo and disabling the CI checks are probably different enough that it makes sense to keep them separate. WDYT?

api/allennlp_demo/tasks/api.py
epwalsh: Yea I don't think we need that

api/allennlp_demo/tasks/api.py
```diff
@@ -0,0 +1,39 @@
...
+import flask  # noqa: E402
+from werkzeug.exceptions import NotFound
+
+from allennlp_demo.common.logs import configure_logging  # noqa: E402
+from allennlp_models.pretrained import get_tasks  # noqa: E402
```
epwalsh, line 9: Shouldn't need the `noqa: E402` comments here either

dev/check_models_ci.py
epwalsh: I think this is fine for now. We should only need this temporarily until VilBERT VQA is fixed.

epwalsh (approved):
LGTM other than removing the `# noqa: E402` comments 👍👍👍

dev/check_models_ci.py
jonborchardt: sounds right, carry on

codeviking: Sounds good, thanks!

api/allennlp_demo/tasks/api.py
codeviking: Awesome, removed -- thx!

PR #666, opened 2020-12-18
Pass input too when rendering the output of `usePromise`.

codeviking:
Before this change only output was accessible, now both output
and the associated input is.


ui/src/tugboat/components/Promised.tsx
```diff
@@ -8,7 +8,7 @@ import { UnknownStateError } from '../error';
 interface Props<I, O> {
     input?: I;
     fetch: (i?: I) => Promise<O>;
-    children: (o: O) => React.ReactNode;
+    children: (o: O, i: I) => React.ReactNode | JSX.Element;
```
jonborchardt, line 11: should this be:
React.ReactNode | JSX.Element | JSX.Element[]

jonborchardt (approved):
great!

aimichal (approved):
Funny to see the order be output followed by input. I'm used to I then O :-)

ui/src/tugboat/components/Promised.tsx
codeviking: I'm not sure. I get really confused about what this type should be. I wish React would just provide `React.ReactChildren`, or something.

I'm going to leave it and err on the side of adding things as we need them. For instance, by doing this I discovered I didn't need `React.ReactNode[]`.

codeviking:
> Funny to see the order be output followed by input. I'm used to I then O :-)

Yea, my gut said I should do that. I'm going to before merging.

My initial rationale was that people are probably always going to use `output`, and will less-often use `input`. So the idea was we might have callbacks that look like this a lot:

```
(output: O) => ...
```

...but they can also do:

```
(_: I, output: O) => ...
```

So I think preserving the typical ordering is the right-call. Thanks!

ui/src/tugboat/components/Promised.tsx
jonborchardt: deal

codeviking:
@aimichal Thanks for the suggestion. I arrived at something that I think is even more elegant.

Instead of passing them as separate properties, I simply pass the `Success` state, which has two properties, `input` and `output`.

This means I can use them via destructuring and only extract what's needed in the callback:

```
({ output }) => /* output */
({ input, output }) => /* input, output */
({ output: predictions }) => /* predictions */
(result) => /* result.input, result.output */
```

jonborchardt:
i think that input should include the model name/id...
i will be needing this

PR #668, opened 2020-12-18
Jon tugboat answers

jonborchardt:
This is the start of tugboat answers
this pr has answer sections that are styleable
it includes how i intend to return different answers based on the model and answertype
it includes a simple highlight component
and some general cleanup

this code is 'ready' for staging, but there are a lot of todos in there before its ready fro prod


ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,38 +14,116 @@ import {
...
+                <SelectExample displayProp="question" placeholder="Select a Question…" />
+                <Predict<Input, Prediction>>
+                    <Fields>
+                        <Passage />
+                        <Question />
+                        <Submit>Run Model</Submit>
+                    </Fields>
+                    <Output>
+                        {({ output, input }: { output: Prediction; input: Input }) => {
+                            let answerType = 'oldInterface';
```
codeviking, line 44: What's `oldInterface` do? Maybe we add a comment here.

jonborchardt: will do

it is a historical code path that does not use the answer type.... reading comp is a mess

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,114 @@ import {
...
+                            if ('answerType' in output) {
+                                answerType = (output as any).answerType;
+                            }
+                            // TODO: we need model as part of input, hacking it in here
+                            // TODO, should there be a set of const ids somewhere?
+                            if (model.id === 'nmn') {
+                                answerType = 'nmn';
+                            }
+
+                            switch (answerType) {
```
codeviking, line 52: Where's `answerType` come from? I don't actually see it in the output that's being returned.

What if we just used `model.id` here instead?

```
switch(model.id) {
   'nmn': { /* NMN Output */
   'bidaf': ...
}
```

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,114 @@ import {
...
+                                }
+                                case NAQANetAnswerType.Arithmetic: {
+                                    // TODO: add other types
+                                    return <>has Arithmetic answer</>;
+                                }
+                                case 'nmn': {
+                                    // TODO: add other types
+                                    return <>has nmn answer</>;
+                                }
+                                default: {
```
codeviking, line 73: I don't think we should assume that `default` maps to a certain type of output. What if instead we tied `model.id == 'bidaf'` to this state, and let `default` throw an error:

```
default: {
   throw new UnknownModelId(model.id);
}
```

This way if the default value in the select changes (a new model is added that's displayed per BiDAF) the person responsible for the change will get a loud n' clear indication of fact that they need to add it's output mechanism here too.

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: its a double switch... but maybe... ill look into it

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: actually, this is just not done, it will instead display <Answer.NoAnswer />

jonborchardt: maybe this was not ready to review... its hard, since its getting bigger every thing i add

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,114 @@ import {
...
+                                    // oldInterface falls to here
+                                    // default is a BiDAFPrediction prediction
+                                    const pp = output as BiDAFPrediction;
+                                    // TODO: there is a bug in the response, so we need to calculate the best_span locally
+                                    const start = input.passage.indexOf(pp.best_span_str);
+                                    const best_span = [start, start + pp.best_span_str.length];
```
codeviking, line 79: We might reword the comment, as it confused me.

It's not that there's a bug in the response, but rather that the response doesn't include the indices of `best_span`, so we have to determine them on the client.

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Yea, it's ok. I think in this case it's expected because it's foundational.  We're laying the grounds for a complex series of output mechanics -- plus Reading Comprehension is the most complex!

What if for this PR we just try to get the mechanics in place for switching output styles, but don't try to wire up any actual output (meaning it won't look nice yet).  Then we could do each output component as an isolated PR?

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: nope, its a bug... it does have the values, but it is [start, start]
its a bug

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: no longer relevant due to code changes

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: ok, i think its ready for that

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+            return (
+                <>
+                    <Answer.Section label="Answer">
+                        <div>{pp.best_span_str}</div>
+                    </Answer.Section>
+
+                    <Answer.Section label="Passage Context">
+                        <TextWithHighlight
+                            text={input.passage}
+                            highlightRanges={[
```
codeviking, line 101: I could see calling this `highlights` for brevity (I'm not sure `Ranges` is necessary) -- but don't feel strongly one way or another.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+    // Now, if the model is NMN, we use yet another output display, so
+    // we are setting the answertype to nmn.
+    // TODO, should there be a set of const ids somewhere?
+    switch (model.id) {
+        case 'bidaf':
+        case 'bidaf-elmo': {
+            // TODO: is bidaf the only model we need here?
+            const pp = output as BiDAFPrediction;
+            // TODO: there is a bug in the response, so we need to calculate the best_span locally
+            const start = input.passage.indexOf(pp.best_span_str);
```
codeviking, line 90: We should probably handle the case where it's not found, maybe by throwing?

```
if (start === -1) {
   throw new UnableToFindHighlightedSpanInInputError(...)
}
```

...or we could just omit highlights when that happens?

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+                </>
+            );
+        }
+        // TODO: I dont see this in the model selection... does it need to be added, or
+        // can we remove this case?
+        case 'nmn': {
+            // TODO: add other types
+            return <>has nmn answer</>;
+        }
+        // default catches the rest of the models that support the newer interface.
```
codeviking, line 130: What about the transformer QA model?

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+        default: {
```
codeviking, line 131: I don't think the `default` case should be NAQANet. If we add another model, it presumably shouldn't be handled here.

I'd suggest `throw`ing in the `default` class so that when someone adds a new model they can't miss the fact that they need to add output for it too.

Also, `answerType` is specific to `NAQANet`, it's not meant to indicate the type of output that's provided.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+                        return <>has QuestionSpan answer</>;
+                    }
+                    case NAQANetAnswerType.Count: {
+                        // TODO: add other types
+                        return <>has Count answer</>;
+                    }
+                    case NAQANetAnswerType.Arithmetic: {
+                        // TODO: add other types
+                        return <>has Arithmetic answer</>;
+                    }
```
codeviking, line 149: I'd consider lifting this into a component dedicated to `NAQANet`:

```
case 'naqanet':
   return <NaqanetAnswer .. />
```

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+}: {
+    input: Input;
+    output: Prediction;
+    model: Model;
+}) => {
+    // Determining which output code to display is non trivial
+    // the original output interface did not include answerType, so
+    // if we will use the older (default) display in the switch below.
+    // Now, if the model is NMN, we use yet another output display, so
+    // we are setting the answertype to nmn.
```
codeviking, line 82: `answerType` isn't meant to be replicated across models -- it's a NAQANet specific thing. So I don't think this comment makes sense.

For this reason I'd adjust the `switch / case` bit below to toggle output based on `model.id` and add a clause for all models.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,143 @@ import {
...
+    // TODO, should there be a set of const ids somewhere?
+    switch (model.id) {
+        case 'bidaf':
+        case 'bidaf-elmo': {
+            // TODO: is bidaf the only model we need here?
+            const pp = output as BiDAFPrediction;
```
codeviking, line 88: I wonder if we could use type-guards here to omit the need for a cast. What if we defined a function for each output type like so:

```
function isBidafOutput(model: Model, output: Output): output is BiDAFOutput {
   return model.id === 'bidaf' || model.id === 'bidaf-elmo';
}
```

Then, we could do this:

```
if (isBidafOutput(model, output)) {
   // ...do something with output here, and it's type will be BiDAFOutput, no cast required.
}
```

WDYT?

ui/src/tugboat/components/Answer.tsx
```diff
@@ -0,0 +1,41 @@
...
+    margin-bottom: ${({ theme }) => theme.spacing.md};
+`;
+
+const NoAnswer = () => {
+    return (
+        <Section label="Answer">
+            <div>No answer returned.</div>
+        </Section>
+    );
+};
```
codeviking, line 39: So, what's interesting about this case is there's actually an answer, we just don't know how to render it. So I'd `throw` in the interest of being loud.

We have an `<ErrorBoundary />` now, so the user will see a big ole' error message (and ideally catch this before merging!).

ui/src/tugboat/components/Answer.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <AnswerWrapper>
+            <Label>{props.label}</Label>
+            {props.children}
+        </AnswerWrapper>
+    );
+};
+
+const AnswerWrapper = styled.div`
+    padding-bottom: ${({ theme }) => theme.spacing.md};
+`;
```
codeviking, line 22: So, I learned about the `<Space />` component recently. I wonder if we should consider using that instead, as to avoid writing a lot of one-off containers like this.

I think we'd just be able to omit `<AnswerWrapper />` and the `margin-*` styles in `Label`.

ui/src/tugboat/components/TextWithHighlight.tsx
```diff
@@ -0,0 +1,53 @@
...
+    color?: string;
+}
+
+interface Props {
+    text: string;
+    highlightRanges: HighlightRange[]; // expected to be in order
+}
+
+/**
+ * Nicely displays the provided text with all its highlightTanges highlighted.
```
codeviking, line 18:
```suggestion
 * Displays a passage of text, highlighting the tokens in the provided `highlightRanges`.
```

ui/src/tugboat/components/TextWithHighlight.tsx
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const TextWithHighlight = (props: Props) => {
+    let lastEndIndex = 0;
+    const ranges = props.highlightRanges.map((r: HighlightRange) => {
+        if (r.start < lastEndIndex || r.start >= r.end) {
+            throw new InvalidAttributesError(
+                'TextWithHighlight takes highlightRanges that need to be in order and non overlapping.'
```
codeviking, line 25: Could we sort the list first, to ensure this doesn't happen?

ui/src/tugboat/components/TextWithHighlight.tsx
```diff
@@ -0,0 +1,53 @@
...
+            );
+        }
+        const leadin = props.text.slice(lastEndIndex, r.start);
+        const highlight = props.text.slice(r.start, r.end);
+        lastEndIndex = r.end;
+        return (
+            <span>
```
codeviking, line 32: I think we might need a `key` here? Probably could just use the tokens in the span?

ui/src/tugboat/components/TextWithHighlight.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const Highlight = styled.span<{ color?: string }>`
+    color: ${({ theme, color }) =>
+        theme.color[color || 'B6'].useContrastText
+            ? theme.palette.text.contrast
+            : theme.palette.text.primary};
+    background: ${({ theme, color }) => theme.color[color || 'B6']};
+    font-weight: 700;
+    padding: 0.1875em;
+    margin: 0 0.125em;
```
codeviking, line 52: Should we use `rem`s here instead? An `em` is relative to the font-size of it's parent -- maybe that's what we want?

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: instead of the switch?

codeviking: Yea, I think so.

While I like `switch / case` statements, I like that it'd omit the need for casts.

codeviking (approved):
LGTM; I left some ideas.

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: well we can... but right now we will always fail....
im partial to just fixing here.. 
what's you call?

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: transformer qa has the same interface as NAQANetPrediction
i dont see the properties of TransformerQAPrediction used anywhere in the code...
TransformerQA follows the same path as NAQANet

am i wrong?

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: yup

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: agreed, ill change how default works...

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Ah, I didn't catch that.

We should figure out a more general name. `NAQANetAndTransformerOutput` (there's probably something better).

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: added to use same output as naqa

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: fixed

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: fixed

ui/src/tugboat/components/Answer.tsx
jonborchardt: done

ui/src/tugboat/components/TextWithHighlight.tsx
jonborchardt: fixed

ui/src/tugboat/components/TextWithHighlight.tsx
jonborchardt: done

ui/src/tugboat/components/TextWithHighlight.tsx
jonborchardt: done

ui/src/tugboat/components/TextWithHighlight.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: ok, i cleaned it up a bit, let me know what you think

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,140 @@ import {
...
+                            );
+                        }}
+                    </Output>
+                </Predict>
+            </MultiModelDemo>
+        </Content>
+    );
+};
+
+const OutputByModel = ({
```
codeviking, line 71: What do you think about lifting this into a separate component, maybe just `Output.tsx`?

It'd be a nice pattern (I think) for people to follow. The motivation being to keep a single component per file pattern for the code that demo developers see (so they can pattern match off of it).  This way we're less likely to end up with 4 or 5 components in a single file :).

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,140 @@ import {
...
+    output,
+    model,
+}: {
+    input: Input;
+    output: Prediction;
+    model: Model;
+}) => {
+    switch (model.id) {
+        case ModelId.Bidaf:
+        case ModelId.BidafElmo: {
```
codeviking, line 82: I wonder if we should use the capitalization scheme that each model uses, so `BiDAFELMo`, and `BiDAF`?

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,140 @@ import {
...
+                    </Answer.Section>
+                </>
+            );
+        }
+        // TODO: I dont see this in the model selection... does it need to be added, or
+        // can we remove this case?
+        case ModelId.Nmn: {
+            // TODO: add other types
+            return <>has nmn answer</>;
+        }
```
codeviking, line 87: It's filtered right now because there's no model card for it. You should see a `console.warn()` statement when you load the page:

https://github.com/allenai/allennlp-demo/blob/master/ui/src/components/MultiModelDemo.tsx#L47

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,140 @@ import {
...
+        case ModelId.Naqanet:
+        case ModelId.TransformerQa:
+            if ('answer_type' in (output as any).answer) {
```
codeviking: We should probably cast here too, like we did above:

```
const answerOutput = output as MultiTypeAnswer
switch(answer.answer.answerType) {
 ...
}
```

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -14,40 +14,140 @@ import {
...
+                    }
+                    case NAQANetAnswerType.Arithmetic: {
+                        // TODO: add other types
+                        return <>has Arithmetic answer</>;
+                    }
+                }
+            }
+    }
+    // If we dont have an output throw.
+    throw new InvalidModelForTaskError(model.id);
```
codeviking, line 162: Maybe `UnexpectedModel` instead? This is ok too, just an idea.

ui/src/demos/reading-comprehension/types.ts
```diff
@@ -26,9 +27,11 @@ export interface TransformerQAPrediction {
     span_start_logits: number[];
 }
```
codeviking, line 28: It's not, remove it!

codeviking: Err, nevermind. Per our conversation in Slack space I think we need to retain it.

ui/src/lib/ModelInfo.ts
```diff
@@ -1,5 +1,13 @@
 import { ModelCard } from '../tugboat/lib/ModelCard';
 
+export enum ModelId {
+    Bidaf = 'bidaf',
+    BidafElmo = 'bidaf-elmo',
+    Naqanet = 'naqanet',
+    TransformerQa = 'transformer-qa',
+    Nmn = 'nmn',
+}
```
codeviking, line 9: So, we should probably make an `AllenNLPDemoConfig` that extends `DemoConfig` and sets the type of `modelId` to `ModelId` instead of string, right?

ui/src/tugboat/components/TextWithHighlight.tsx
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Displays a passage of text, highlighting the tokens indicated by `highlights`.
+ */
+export const TextWithHighlight = (props: Props) => {
+    let lastEndIndex = 0;
+    const ranges = props.highlights
+        .sort((a, b) => a.start - b.start)
+        .map((r: Highlights, i: number) => {
+            if (r.start < lastEndIndex || r.start >= r.end) {
+                throw new InvalidAttributesError(
```
codeviking, line 26: I'd suggest making this more specific, say `InvalidHighlightRange`.

The purpose of these typed errors is twofold:

1. To make it easier for the user to catch errors and decide how to handle them.
2. To make it clear to the developer what's wrong.  The better the name the easier it is for them to trace it down.

So I think `InvalidAttributesError` is a bit too generic, I'm not sure what `Attributes` are precisely, and lots of things could have them -- which I think might cause the error to be overused.  In that case it provides little beyond `throw Error()`.

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: added type guards

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: i think that gets hard to do...
im inclined to not... your call

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: k

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: fixed with typeguards

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: k

ui/src/tugboat/components/Answer.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

ui/src/lib/ModelInfo.ts
jonborchardt: maybe, im going to call that out of scope for right now

ui/src/demos/reading-comprehension/types.ts
jonborchardt: will it come back?
i see a model with its name in the dropdown...

ui/src/tugboat/components/TextWithHighlight.tsx
jonborchardt: done

ui/src/lib/ModelInfo.ts
codeviking: Hmm. It feels a little silly to be opting out of strings here but not in the config files. I'd argue we should be consistent in either direction.

Either we use the `enum` across the board, or we use strings -- having both seems confusing, as a developer I'm not sure which to use when...

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: leaving as is

ui/src/demos/reading-comprehension/types.ts
jonborchardt: leaving for now, remove if you think we should

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -9,45 +9,34 @@ import {
...
+                <ModelCardModal />
+                <ModelUsageModal />
+                <SelectExample displayProp="question" placeholder="Select a Question…" />
+                <Predict<Input, Prediction>>
+                    <Fields>
+                        <Passage />
+                        <Question />
+                        <Submit>Run Model</Submit>
+                    </Fields>
+                    <Output />
```
codeviking, line 37: Ah! Nice.

PR #669, opened 2020-12-18
Only run API checks when there's API changes.

codeviking:


codeviking:
Closing in favor of #670.

PR #672, opened 2020-12-21
add model card as tab

jonborchardt:
moves model card to a tab
and fills it in

image


aimichal:
What's a "registered name" and "id"?

What would this table look like if the left-most column was wide enough to hold the longest field name like "Performance Measure" on a single line?

ui/package.json
```diff
@@ -61,7 +61,8 @@
     "start": "webpack-dev-server --mode development",
     "build": "webpack --mode production",
     "lint": "eslint '**/*.{js,jsx,ts,tsx,json}' && echo '💫  Lint complete.'",
-    "lint:fix": "eslint '**/*.{js,jsx,ts,tsx,json}' --fix && echo '🛠  Lint --fix complete'"
+    "lint:fix": "eslint '**/*.{js,jsx,ts,tsx,json}' --fix && echo '🛠  Lint --fix complete'",
+    "dockerui": "docker-compose -f ../docker-compose.ui-only.yaml up --build"
```
aimichal, line 65: 👍

aimichal (approved):
I think some (?) bubbles would help, but can be added in a future rev.

jonborchardt:
> What's a "registered name" and "id"?
> 
> What would this table look like if the left-most column was wide enough to hold the longest field name like "Performance Measure" on a single line?

1- im not sure, ill show this to the allennlp team before we ship to prod
2- ill see about making it wider

jonborchardt:
> I think some (?) bubbles would help, but can be added in a future rev.

good idea

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -25,17 +27,25 @@ export const Main = () => {
...
+                            </Fields>
+                            <Output />
+                        </Predict>
+                    </Tabs.TabPane>
+                    <Tabs.TabPane tab="Model Card" key="Card">
+                        <ModelCard />
+                    </Tabs.TabPane>
+                    <Tabs.TabPane tab="Model Usage" key="Usage">
+                        <ModelUsage />
+                    </Tabs.TabPane>
```
codeviking, line 47: 👍  Do we anticipate adding URL paths that trigger the tabs too?

codeviking (commented):
LGTM; The first row of the table is hard for me to parse, since there's multiple fields on one line.  

This `antd` component seems like it might be well suited for this information:
https://ant.design/components/descriptions/

Maybe we give it a shot and see how it looks?

PR #673, opened 2020-12-21
add usage info

jonborchardt:
pulled over syntaxhighlighter into ts
added a new tab for usage
added a usage file per task
updated global usage component

image

### DIFF
--- a/ui/src/demos/reading-comprehension/Usage.tsx
+++ b/ui/src/demos/reading-comprehension/Usage.tsx
@@ -1,14 +1,14 @@
-/*
-  This file defines the text used when explaining to the user how to use this model.
-  It is specific to the task and changes slightly based on the model.
-*/
+/**
+ * This file defines the text used when explaining to the user how to use this model.
+ * It is specific to the task and changes slightly based on the model.
+ */
 
 import React from 'react';
 
-import { ModelUsage } from '../../tugboat/components';
+import { ModelUsage } from '../../lib';
 import { Models, Examples } from '../../tugboat/context';
-import { NoSelectedModel } from '../../tugboat/error';
-import { GroupedExamples } from '../../tugboat/lib';
+import { NoSelectedModel, InvalidExamplesFormat } from '../../tugboat/error';
+import { isGroupedExamples } from '../../tugboat/lib';
 
 export const Usage = () => {
     const models = React.useContext(Models);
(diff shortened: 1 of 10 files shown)


ui/src/demos/reading-comprehension/Usage.tsx
```diff
@@ -0,0 +1,47 @@
+/*
+  This file defines the text used when explaining to the user how to use this model.
+  It is specific to the task and changes slightly based on the model.
+*/
+
+import React from 'react';
+
```
jonborchardt, line 7: this file ins intended to bridge the task details and the global usage component, how do we like this level of abstraction?

ui/src/demos/reading-comprehension/Usage.tsx
```diff
@@ -0,0 +1,47 @@
+/*
+  This file defines the text used when explaining to the user how to use this model.
+  It is specific to the task and changes slightly based on the model.
+*/
```
codeviking, line 4: I've been using this syntax for multi-line comments:

```
/**
 * Text...
 * more...
 */
```

We might stick with that in the interest of consistency. WDYT?

ui/src/demos/reading-comprehension/Usage.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { GroupedExamples } from '../../tugboat/lib';
+
+export const Usage = () => {
+    const models = React.useContext(Models);
+    const examples = React.useContext(Examples);
+    if (!models.selectedModel) {
+        throw new NoSelectedModel();
+    }
+    // Reading comp uses GroupedExamples
+    const ex = (examples.examples as GroupedExamples)['SQuAD-like Argument Finding'][2]; // matrix example
```
codeviking, line 20: There's a method called `isGroupedExamples` in `src/tugboat/lib/Example.ts`. I'd suggest using that here instead of the `cast`.

You might do something simple, like exit if they're not grouped with an error:

```
if (!isGroupedExample(examples)) {
   throw new InvalidExamplesFormat(`The examples aren't grouped, but we expect them to be`).
}
```

I'll probably have more thoughts after I see how `examples` are used below.

ui/src/demos/reading-comprehension/Usage.tsx
```diff
@@ -0,0 +1,47 @@
...
+    // TODO: we need to get these paths from model card
```
codeviking, line 21: 👍

ui/src/tugboat/components/ModelUsage.tsx
```diff
@@ -1,24 +1,84 @@
...
         <>
-            <h4>Installing AllenNLP</h4>
-            {models.selectedModel.card.display_name}
-            <h4>Prediction</h4>
-            <h5>On the command line (bash):</h5>
-            <h5>As a library (python):</h5>
-            <h4>Evaluation</h4>
-            <h4>Training</h4>
-            <h4>Installing AllenNLP</h4>
+            <h5>Installing AllenNLP</h5>
```
codeviking, line 29: This should probably be a prop too, since other demos that use Tugboat probably won't want people to install AllenNLP 😂.

I wonder how applicable this component is to other demos, given that I'd expect a lot of variety with respect to installation and setup instructions.

ui/src/tugboat/components/SyntaxHighlight.tsx
```diff
@@ -0,0 +1,39 @@
...
+  Supported Languages:
+  https://github.com/conorhastings/react-syntax-highlighter/blob/HEAD/AVAILABLE_LANGUAGES_HLJS.MD
+
+  Supported Styles:
+  https://github.com/conorhastings/react-syntax-highlighter/blob/HEAD/AVAILABLE_STYLES_HLJS.MD
+
+  Demo:
+  https://highlightjs.org/static/demo/
+
+*******************************************************************************/
```
codeviking, line 25: 👍

codeviking (approved):
This seems like a great first step.

I find myself wondering if we'll end up wanting to just load a `.md` file and display it. Maybe MDX is a good fit here?

If we use Markdown instead of a component it'll make it easier for people to hedge in additional instructions, which might be necessary. Say a specific model needs some third-party thing -- or a system setting to be tweaked.  If we use Markdown they simply edit the document, if we use a component then we might end up with a confusing number of `prop`s.

Either way I think it's ok to wait and see -- unless this suggestion prompts you to re-evaluate.  When we do go the markdown right we might look into the Markdown loader, so that the Markdown content is just baked into the bundle:
https://github.com/peerigon/markdown-loader



1

ui/src/demos/reading-comprehension/Usage.tsx
jonborchardt: agreed
(i wonder if we can lint this)

ui/src/demos/reading-comprehension/Usage.tsx
jonborchardt: great!

codeviking:
Oh, one more thing -- I might add a bit of padding between the light-grey background and the text -- I think it'd look nice.

ui/src/tugboat/components/ModelUsage.tsx
jonborchardt: oh, ya, this should come out of tugboat...

jonborchardt:
> Oh, one more thing -- I might add a bit of padding between the light-grey background and the text -- I think it'd look nice. Also, I see light gray underlines on every line -- is that intentional?

ill take a. look, ty

jonborchardt:
> This seems like a great first step.
> 
> I find myself wondering if we'll end up wanting to just load a `.md` file and display it. Maybe MDX is a good fit here?
> 
> If we use Markdown instead of a component it'll make it easier for people to hedge in additional instructions, which might be necessary. Say a specific model needs some third-party thing -- or a system setting to be tweaked. If we use Markdown they simply edit the document, if we use a component then we might end up with a confusing number of `prop`s.
> 
> Either way I think it's ok to wait and see -- unless this suggestion prompts you to re-evaluate. When we do go the markdown right we might look into the Markdown loader, so that the Markdown content is just baked into the bundle:
> https://github.com/peerigon/markdown-loader

interesting...
ill look into mdx for this.... we wantt o preserve getting values from the model, but for the rest, ya maybe...
i think i will clean this up and merge, than look into mdx

codeviking:
> i think i will clean this up and merge, than look into mdx

👍  Sounds great.

ui/src/demos/reading-comprehension/Usage.tsx
jonborchardt: fixed

ui/src/demos/reading-comprehension/Usage.tsx
jonborchardt: fixed

ui/src/tugboat/components/ModelUsage.tsx
jonborchardt: fixed

jonborchardt:
all done

PR #674, opened 2020-12-22
Interpret + output refactoring.

codeviking:
This adds the `<Interpret />` component in the simplest form --
right now it just dumps the response as JSON.

I also did a little refactoring of the output, given it felt
like the demo-specific `Output.tsx` file was getting a little
overloaded:

- I renamed `Output.tsx` to `Predictions.tsx`
- I lifted the output that's not related to the predict method into
  the top-level component (the interpret stuff).
- I refactored the `<Answer.Section />` bits into `<Output />`
  and stopped using `antd`'s `<Space />` given it only has 3 levels
  and we need more.


ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -36,7 +38,28 @@ export const Main = () => {
...
+                                {({ model, input, output }) => (
+                                    <Output.Sections>
+                                        <Output.Section title="Model Predictions">
+                                            <Predictions
+                                                model={model}
+                                                input={input}
+                                                output={output}
+                                            />
+                                        </Output.Section>
+                                        <Output.Section title="Model Interpretations">
```
jonborchardt, line 51: these should be part of output....IMO

ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -21,28 +17,24 @@ import {
...
+        <OutputByModel input={input} output={output} model={model} />
 
-                    <Answer.Section label="Output JSON (DEBUG)">
-                        <PrettyPrintedJSON json={output} />
-                    </Answer.Section>
-                </>
-            );
-        }}
-    </TBOutput>
+        <Output.SubSection title="Debug Output">
```
jonborchardt, line 24: nice

jonborchardt (approved):
looks good, but i was expecting (see my pr: https://github.com/allenai/allennlp-demo/pull/676)
that the interpretations and attacks would be part of the 'Predictions' below

the reason is that showing each is model dependent, which is handled by the answerbyModel function

ui/src/demos/reading-comprehension/Main.tsx
codeviking: Yea, I decided to lift things out of that component because it was quickly becoming a 400+ line monster component with a few too many things stuffed into there. 

Specifically the `<Intepret />` and to be implemented `<Attack />` bits aren't `/predict` outputs, so they don't belong there. They use the same input, but they're separate types of output, so I think they deserve their own packaging.

Also, I renamed `Output.tsx` to `Prediction.tsx` to better capture that it's the visualization of a model's predictions to further cement this notion.

WDYT?

codeviking:
> the reason is that showing each is model dependent, which is handled by the answerbyModel function

This is still possible with these changes. We have access to the `model` at the top-level, and could branch in `Main.tsx` as needed. That said I think it'd be a bit cleaner to add `Interpretations.tsx` to `src/demos/reading-comprehension` and add that branching there.

My motivation is to reduce the size and complexity of `Output.tsx`, as it feels like that file is getting a bit overloaded. WDYT?

ui/src/demos/reading-comprehension/Interpreters.tsx
```diff
@@ -0,0 +1,40 @@
...
+                </Interpret>
+            </Collapse.Panel>
+            <Collapse.Panel key="smooth" header="SmoothGrad Visualization">
+                <Interpret<Input, any> interpreter="smooth_gradient" input={input}>
+                    {({ output }) => <PrettyPrintedJSON json={output} />}
+                </Interpret>
+            </Collapse.Panel>
+        </Collapse>
+    );
+};
```
codeviking, line 40: @jonborchardt Take a look. I lifted the interpret stuff into a separate file. I think this might mesh nicely with your work?

Let's sync tomorrow and decide what route we want to go.

codeviking: We're going to try this route and re-assess in the future if we need to.

PR #676, opened 2020-12-22
initial code to get saliency to work

jonborchardt:
adds interpretation viz


ui/src/@types/colormap.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'colormap' {
+    const src: any;
+    export default src;
+}
```
codeviking, line 4: What's this for? Can we add a comment explaining why we need this, for our future selves?

ui/src/demos/reading-comprehension/Output.tsx
```diff
@@ -58,13 +65,13 @@ const OutputByModel = ({
         case ModelId.Bidaf:
         case ModelId.BidafElmo: {
             if (isBiDAFPrediction(output)) {
-                return <BasicAnswer input={input} output={output} />;
+                return <BiDAFPredictionAnswer input={input} output={output} />;
```
codeviking, line 68: I would drop the `Answer` suffix. These are actually predictions, not answers per say.  So I think `<BiDAFPrediction />` would do the trick.

I know @aimichal has thoughts on this verbiage too :).

ui/src/demos/reading-comprehension/Output.tsx
```diff
@@ -85,12 +92,63 @@ const OutputByModel = ({
...
 };
 
-const BasicAnswer = ({
+const BiDAFPredictionAnswer = ({ input, output }: { input: Input; output: BiDAFPrediction }) => {
+    let best_span = output.best_span;
+    if (best_span[0] >= best_span[1]) {
+        // TODO: there is a bug in the response, so we need to calculate the best_span locally
+        const start = input.passage.indexOf(output.best_span_str);
+        best_span = [start, start + output.best_span_str.length];
+    }
```
codeviking, line 101: Should we cut this as an issue here: https://github.com/allenai/allennlp-demo/milestone/2?

ui/src/demos/reading-comprehension/Output.tsx
```diff
@@ -85,12 +92,63 @@ const OutputByModel = ({
...
+            </Answer.Section>
+
+            {interpretData ? (
+                <MySaliencyMaps
+                    interpretData={interpretData}
+                    questionTokens={output.question_tokens}
+                    passageTokens={output.passage_tokens}
+                    interpretModel={interpretModel}
+                    requestData={input}
+                />
```
codeviking, line 136: As noted in [this PR](https://github.com/allenai/allennlp-demo/pull/674) I don't this component is getting a little too complicated. I proposed renaming it to `<Predictions />` and lifting anything that's not explicitly related to visualizing the predictions into other components.

ui/src/demos/reading-comprehension/Output.tsx
```diff
@@ -153,3 +207,54 @@ const NaqanetAnswer = ({ output }: { output: NAQANetPrediction }) => {
...
     }
 };
+
+const getGradData = ({
+    grad_input_1: gradInput1,
+    grad_input_2: gradInput2,
+}: GradInputInstance): [number[], number[]] => {
+    // Not sure why, but it appears that the order of the gradients is reversed for these.
+    return [gradInput2, gradInput1];
+};
```
codeviking, line 217: Whacky. We should verify this with someone and if it's the case ask if we can just change this in the back-end. This feels like the type of code-smell we should squash if we can while we're making the changes.

Fixing this should make the API easier for all parties (given AllenNLP has users who *aren't* the demo).

ui/src/demos/reading-comprehension/Output.tsx
```diff
@@ -153,3 +207,54 @@ const NaqanetAnswer = ({ output }: { output: NAQANetPrediction }) => {
...
+        Interpreter.SG_INTERPRETER in interpretData
+            ? getGradData(interpretData[Interpreter.SG_INTERPRETER].instance_1)
+            : undefined;
+    const inputTokens: [string[], string[]] = [questionTokens, passageTokens];
+    const inputHeaders = [<h5>Question:</h5>, <h5>Passage:</h5>];
+    const allInterpretData: AllInterpreterData = {
+        simple: simpleGradData,
+        ig: integratedGradData,
+        sg: smoothGradData,
+    };
```
codeviking, line 250: I'm confused as to why these are all being aggregated into one struct here. Aren't they separate things.  When I visit this section of the demo currently I see three distinct chunks:

image

Each of these chunks has their own API endpoint, and thus their own IO.  Maybe that separation means we don't need to aggregate them together in this fashion (and thus do less work to be defensive)?

ui/src/lib/Saliency.tsx
```diff
@@ -0,0 +1,475 @@
...
+                    target="_blank"
+                    rel="noopener noreferrer">
+                    SmoothGrad
+                </a>
+                .
+            </p>
+        );
+    }
+    return description;
+};
```
codeviking, line 79: Ah. I see. I think you and I stepped on each other's toes here a bit.

I was thinking we'd end up with something like this, instead of a single component that handles the output for all three:

```tsx
<Interpret type="simple_gradient">
   <Intepret.Description>
          <p>See salience maps generated using <a href="...">this approach.</a></p>. 
          <Submit>See Interpretations</Submit>
    </Interpret.Description>
     {({ output }) => <SimpleGradientMap gradients={output} />
</Interpret>
<Interpret type="smooth">
   <Intepret.Description>
          <p>See salience maps generated using <a href="...">this approach.</a></p>. 
          <Submit>See Interpretations</Submit>
    </Interpret.Description>
     {({ output }) => <SmoothGradientMap gradients={output} />
</Interpret>
```

I think this makes sense because each has its own API endpoint, with its own types, output styles and behavior. By making them distinctly separate we avoid large bits of branching code in a lot of places.

Let me know what you think!

ui/src/lib/Saliency.tsx
```diff
@@ -0,0 +1,475 @@
...
+    | 'hot'
+    | 'cool'
+    | 'spring'
+    | 'summer'
+    | 'autumn'
+    | 'winter'
+    | 'bone'
+    | 'copper'
+    | 'greys'
+    | 'YIGnBu'
```
codeviking, line 217: Is this really a type? This looks like a styled-component class name :-P.

ui/src/lib/Saliency.tsx
```diff
@@ -0,0 +1,475 @@
...
+    | 'velocity-blue'
+    | 'velocity-green'
+    | 'cubehelix';
+
+interface TokensWithWeight {
+    token: string;
+    weight: number;
+}
+
+export class SaliencyComponent extends React.Component<
```
codeviking, line 255: Can we use `PureComponent` here instead? 

Also, this feels like it's complicated enough that it might deserve it's own file?

ui/src/lib/Saliency.tsx
```diff
@@ -0,0 +1,475 @@
...
+        background: ${({ theme }) => theme.color.B4};
+        cursor: pointer;
+    }
+
+    &::-webkit-slider-thumb {
+        appearance: none;
+    }
+`;
+
+export default SaliencyMaps;
```
codeviking, line 475: This feels like you and I stepped on each other's toes again, as I think a lot of this code will (and should) be reworked after #674 lands.

For instance, there's bits that handle the `loading` state here, which we won't need after that code merges.

Also, could we just opt into the `antd` `<Slider />` now rather than porting over the custom one? I bet the `antd` one would solve all sorts of bugs and accessibility issues that exist in the current one...so it seems like the right thing to do, even if it's visually different.

codeviking (commented):
I left a few thoughts; Let's chat about order of operations n' all that, as I think this and #674 collide a bit.

ui/src/@types/colormap.d.ts
jonborchardt: done

ui/src/demos/reading-comprehension/Output.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Output.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Output.tsx
jonborchardt: no longer relevent

ui/src/demos/reading-comprehension/Output.tsx
jonborchardt: no longer done this way

ui/src/lib/Saliency.tsx
jonborchardt: discussed offline

ui/src/lib/Saliency.tsx
jonborchardt: yes.. it is yellow to indigo to blue

ui/src/lib/Saliency.tsx
jonborchardt: done

ui/src/lib/Saliency.tsx
jonborchardt: ill port the slider to antd after the break
the rest has been resolved offline

ui/src/demos/reading-comprehension/Output.tsx
jonborchardt: done: https://github.com/allenai/allennlp-demo/issues/684

ui/src/components/Interpret.tsx
```diff
@@ -5,24 +5,29 @@ import { Form, Field, Fields, Output, Result, Submit } from '../tugboat/componen
 
 interface Props<I, O> {
     input: I;
+    description?: React.ReactNode;
```
codeviking, line 8: I'd propose passing this as a child. I think we should avoid passing components as props as a general rule of thumb, as it ends up being clunky to use and generates code that's less readable. Cramming DOM into an attribute of a tag isn't natural or easy to read.

So, instead we'd do something like what we do in `<Form />` to parse the children and display things in the correct order.  So we could do:

```
<Interpret>
   <p>Get some cool salience map things.</p>
   <InterpretOutput>{(output) => ...}</InterpretOutput>
</Interpret>
```

In this case we could simply display all children as part of the `<Form />`, and make the last child our output handler.

WDYT?

ui/src/components/Interpret.tsx
```diff
@@ -5,24 +5,29 @@ import { Form, Field, Fields, Output, Result, Submit } from '../tugboat/componen
...
-                <Field key={fieldName} name={fieldName} hidden>
-                    <Input value={input[fieldName]} />
-                </Field>
-            ))}
-            <Submit>Interpret Prediction</Submit>
-        </Fields>
-        <Output>{children}</Output>
-    </Form>
+    <>
+        {description}
```
codeviking, line 20: If you preserve the DOM as prop pattern, I'd just make this a child of `<Fields />`.

It's ok to have things that aren't `<Input />`s there. They won't be sent as part of the submission, and it avoids the wrapping `< />`.

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export interface InputTokens {
+    passage_tokens: string[];
+    question_tokens: string[];
+}
+
+export const isInputTokens = (x: any): x is InputTokens => {
+    const xx = x as InputTokens;
+    return xx.passage_tokens !== undefined && xx.question_tokens !== undefined;
```
codeviking, line 26:
```suggestion
    return Array.isArray(xx.passage_tokens) && Array.isArray(xx.question_tokens);
```

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        black-box) neural models, leave researchers and practitioners scratching
+                        their heads asking, <i>why did my model make this prediction?</i>
+                    </p>
+                    <a
+                        href="https://allennlp.org/interpret"
+                        target="_blank"
+                        rel="noopener noreferrer">
+                        Learn More
+                    </a>
+                </div>
```
codeviking, line 64: I'm going to repeat myself here, but I wonder if this should also be handled via composition as noted above.

Maybe we could get rid of the `title` bit and just convert this to DOM that's a child of `<Output.Section />`, no tricky composition bits required:

```
<Output.Section>
   <Output.SectionTitle>Model Interpreters. <Popover help=...>?</Popover></Output.SectionTitle>
    ...
</Output.Section>
```

The advantage being that we make `<Output.Section />` that more flexible (so we can accommodate similar patterns elsewhere).

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+    instance_1: {
+        grad_input_1: number[];
+        grad_input_2: number[];
+    };
+}
+
+export interface InputTokens {
+    passage_tokens: string[];
+    question_tokens: string[];
+}
```
codeviking, line 22: Since we're exporting these, we might consider moving them to `lib/`.  Maybe `lib/InterpretResponse.ts`?

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { ModelInfoList } from '../context';
+import { InterpreterId } from '../lib';
+import { Input } from '../demos/reading-comprehension/types';
+
+export interface InterpreterData {
+    instance_1: {
+        grad_input_1: number[];
+        grad_input_2: number[];
+    };
+}
```
codeviking, line 17: Is this the shape of the output, or response? If so maybe we call this `InterpretResponse` or `InterpretOutput`? I'd probably go with `Output` to be consistent with nomenclature used elsewhere.

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        <Interpret<Input, InterpreterData>
+                            interpreter={InterpreterId.SimpleGradient}
+                            input={input}
+                            description={
+                                <p>
+                                    See saliency map interpretations generated by{' '}
+                                    <a
+                                        href="https://arxiv.org/abs/1312.6034"
+                                        target="_blank"
+                                        rel="noopener noreferrer">
```
codeviking, line 80:
```suggestion
                                        rel="noopener,noreferrer">
```

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        <Interpret<Input, InterpreterData>
+                            interpreter={InterpreterId.IntegratedGradient}
+                            input={input}
+                            description={
+                                <p>
+                                    See saliency map interpretations generated using{' '}
+                                    <a
+                                        href="https://arxiv.org/abs/1703.01365"
+                                        target="_blank"
+                                        rel="noopener noreferrer">
```
codeviking, line 103:
```suggestion
                                        rel="noopener,noreferrer">
```

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        <Interpret<Input, InterpreterData>
+                            interpreter={InterpreterId.SmoothGradient}
+                            input={input}
+                            description={
+                                <p>
+                                    See saliency map interpretations generated using{' '}
+                                    <a
+                                        href="https://arxiv.org/abs/1706.03825"
+                                        target="_blank"
+                                        rel="noopener noreferrer">
```
codeviking, line 126:
```suggestion
                                        rel="noopener,noreferrer">
```

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+                            {({ output }) => <Interpreter output={output} tokens={tokens} />}
+                        </Interpret>
+                    </Collapse.Panel>
+                ) : null}
+            </Collapse>
+        </Output.Section>
+    );
+};
+
+export const Interpreter = ({
```
codeviking, line 141: Does this need to be `export`ed? It feels unlikely to be used outside of this file?

ui/src/components/Interpreters.tsx
```diff
@@ -0,0 +1,155 @@
...
+    tokens,
+}: {
+    output: InterpreterData;
+    tokens: InputTokens;
+}) => {
+    return (
+        <Saliency
+            interpretData={[output.instance_1.grad_input_2, output.instance_1.grad_input_1]}
+            inputTokens={[tokens.question_tokens, tokens.passage_tokens]}
+            inputHeaders={[<h6>Question</h6>, <h6>Passage</h6>]}
```
codeviking, line 152: Some models take different input -- so this probably will need to be generalized.

Also, I'd suggest omitting the `<h6 />` tags. The display component should take care of those, instead we should just pass a list of strings:

```
inputHeaders={["Question", "Passage" ]}
```

I wonder if we can derive this, by title casing the fields in the response?

ui/src/demos/reading-comprehension/types.ts
```diff
@@ -1,14 +1,14 @@
+import { InputTokens, isInputTokens } from '../../components';
+
 export interface Input {
     passage: string;
     question: string;
 }
 
-export interface BiDAFPrediction {
+export interface BiDAFPrediction extends InputTokens {
```
codeviking, line 8: What do you think of renaming`InputTokens` to `HasInputTokens` or `Tokenized`?

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
+/**
+ * Renders a component that displays each word of the tokens with a different color based on the
+ * associated weight.
+ *
+ * There is also a slider used to decide a cutoff on what weights to not render.
+ */
```
codeviking, line 6: Let's throw this above the component declaration (that's what I've been doing) for consistency.

I may have a few top level comments -- but in that case they're for the entire module, i.e. `This file contains code for doing _____`.  Usually this only exists when a file contains a bunch of related things, like a `types.ts` or `utils.ts`.

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+
+export class Saliency extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+
+        this.state = {
+            topK: { all: 3 }, // 3 words are highlighted by default
+        };
+    }
```
codeviking, line 92: WDYT about trying to rewrite this as a functional component?

I think it'd be nice to be consistent. It means those who look at the code after us aren't confused about when to use a class component instead of a functional one.

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+        colormapProps: {
+            colormap: 'copper',
+            format: 'hex',
+            nshades: 20,
+        },
+    };
+
+    getTokenWeightPairs(grads: number[], tokens: string[]) {
+        return tokens.map((token, idx: number) => {
+            const weight = grads[idx];
```
codeviking, line 104: Could `weight` be `undefined`? Should we handle that case?

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+            // We do 1 - weight because the colormap is inverted
+            return { token, weight: 1 - weight };
+        });
+    }
+
+    colorize(tokensWithWeights: TokensWithWeight[], topKIdx: Set<number>) {
+        const { colormapProps } = this.props;
+        // colormap package takes minimum of 6 shades
+        colormapProps.nshades = Math.min(Math.max(colormapProps.nshades, 6), 72);
+        const colors = colormap(colormapProps);
```
codeviking, line 114:
```suggestion
        const nshades = Math.min(Math.max(colormapProps.nshades, 6), 72);
        const colors = colormap({ ...colormapProps, nshades });
```

This avoids mutating the property (which is a reference, so anyone with a handle to it will receive any mutations).

ui/src/lib/Saliency.tsx
jonborchardt: ok, i converted the slider to use antd

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+        const colorizedString = tokensWithWeights.map((obj, idx) => {
+            // Again, 1 -, in this case because low extreme is blue and high extreme is red
+            return (
+                <Tooltip title={(1 - obj.weight).toFixed(3)}>
```
codeviking: Are these decimal values below 0? If not, we may want to use `Intl.NumberFormat` to add commas n' all that.

I generally try to use `Intl.NumberFormat` over `toFixed`, given it in theory makes our interfaces more accessible to people outside of the US (I imagine this demo actually has quite a few folks in that category).

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+            );
+        });
+        return colorizedString;
+    }
+
+    // when the user changes the slider for input 1, update how many tokens are highlighted
+    handleInputTopKChange = (inputIndex: number) => (e: any) => {
+        const stateUpdate = Object.assign({}, this.state);
+        if (e.target.value.trim() === '') {
+            stateUpdate.topK[inputIndex] = e.target.value;
```
codeviking: So, if it's not set we set an empty string? Would `null` or `undefined` be more canonical here?

I suppose I'm really asking what the type is. Below we set it to an integer, so it feels odd to set it to a string here.

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+        } else {
+            stateUpdate.topK[inputIndex] = parseInt(e.target.value, 10);
```
codeviking: I think you can drop the radix, as base-10 is the default.

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+        }
+        this.setState(stateUpdate);
+    };
+
+    // Extract top K tokens by saliency value and return only the indices of the top tokens
+    getTopKIndices(tokensWithWeights: TokensWithWeight[], inputIndex: number) {
+        function gradCompare(obj1: TokensWithWeight, obj2: TokensWithWeight) {
+            return obj1.weight - obj2.weight;
+        }
```
codeviking, line 147: I'd consider writing this as an arrow function. Technically inline function declarations that use the `function` keyword aren't valid and would fail in strictmode.  I'm pretty sure TypeScript and/or some layer of transpilation prevents this from being an issue.

```
const gradCompare = (o1, o2) => o1.weight - o2.weight;
```

ui/src/components/Interpreters.tsx
jonborchardt: removed

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+        // Add indices so we can keep track after sorting
+        const indexedTokens = tokensWithWeights.map((obj, idx) => {
+            return { ...obj, ...{ idx } };
```
codeviking, line 151: I think you can just do:

```
{ ...obj, idx }
```

ui/src/tugboat/components/Saliency.tsx
```diff
@@ -0,0 +1,253 @@
...
+    margin: 1px;
+    display: inline-block;
+    border-radius: ${({ theme }) => theme.shape.borderRadius.default};
+`;
+
+const Info = styled.span`
+    color: ${({ theme }) => theme.color.B6};
+`;
+
+// TODO: just use antd slider
```
codeviking: 👍

ui/src/tugboat/components/form/Output.tsx
```diff
@@ -60,17 +62,47 @@ Output.Sections = styled.section`
     gap: ${({ theme }) => theme.spacing.xl};
 `;
 
+const PopoverWidthFix = createGlobalStyle`
+  .ant-popover{
+    max-width: 70%;
```
codeviking, line 67: Would `70vw` work?

ui/src/tugboat/components/form/Output.tsx
```diff
@@ -60,17 +62,47 @@ Output.Sections = styled.section`
...
+        <TitleRow>
+            <OutputSectionTitle>{title}</OutputSectionTitle>
+            {helpContent ? <PopoverTarget>{helpLabel}</PopoverTarget> : null}
+        </TitleRow>
+    );
+    return (
+        <OutputSection>
+            {helpContent ? (
+                <>
+                    <PopoverWidthFix />
```
codeviking, line 94: So, I'm pretty sure we'll be changing the width of all popovers.

Do we actually want that? This seems like an odd side-effect.

Can we somehow scope the selector so it only impacts this pop-over? I'd hate for someone elsewhere to write more global CSS to tweak this down the road when they decide they want a different size and for things to regress here due to that.

ui/src/tugboat/components/Saliency.tsx
jonborchardt: outdated

ui/src/tugboat/components/form/Output.tsx
```diff
@@ -82,6 +114,13 @@ const OutputSectionTitle = styled.h4`
     margin: 0;
 `;
 
+const TitleRow = styled.div`
+    margin: 0;
+    display: grid;
+    grid-template-columns: auto 1fr;
+    align-items: center;
```
codeviking, line 121: Should we have a `gap` for small screens?

ui/src/tugboat/components/Saliency.tsx
jonborchardt: outdated

ui/src/tugboat/components/form/Output.tsx
```diff
@@ -99,3 +138,9 @@ const OutputSubSectionTitle = styled.h5`
     margin: 0;
     text-transform: none;
 `;
+
+const PopoverTarget = styled.span`
+    ${LinkCSS.default()}
+    padding-left: ${({ theme }) => theme.spacing.md};
```
codeviking, line 144: My guess is if you use `gap` above you can omit this (and I think it's a bit more canonical,  as if things wrap the `gap` mechanic will work correctly).

codeviking (approved):
Nice work. I left some ideas.

ui/src/components/Interpret.tsx
codeviking: We can always do this later, too. If you like the idea maybe throw a TODO in there.

codeviking: We talked offline, and decided to preserve this.

We landed a nice rule of thumb:

1. If the markup is something that's explicitly placed by the component, it should be passed as a property.
2. If the markup is truly something that's completely composable, then it should be provided as a child, and the render props / child as a function pattern should be used to passed whatever data is provided by the parent component.

This *really* makes sense to me, and helped provide some structure to something I was noodling on but hadn't effectively put to words. Thanks!

ui/src/components/Interpreters.tsx
codeviking: This is a place where the placement of the resulting HTML is controlled, so a property makes sense.

ui/src/components/Interpret.tsx
jonborchardt: we discussed offline and decided that as is is ok for now (or possibly forever)

ui/src/components/Interpret.tsx
jonborchardt: we discussed offline and decided that as is is ok for now (or possibly forever)

ui/src/components/Interpreters.tsx
jonborchardt: we discussed offline and decided that as is is ok for now (or possibly forever)

ui/src/components/Interpreters.tsx
jonborchardt: done

ui/src/components/Interpreters.tsx
jonborchardt: done

ui/src/components/Interpreters.tsx
jonborchardt: this is a non generic file
but i did move the h6's out

ui/src/demos/reading-comprehension/types.ts
jonborchardt: done

ui/src/tugboat/components/Saliency.tsx
jonborchardt: done

ui/src/tugboat/components/Saliency.tsx
jonborchardt: done

ui/src/tugboat/components/Saliency.tsx
jonborchardt: fixed

ui/src/tugboat/components/Saliency.tsx
jonborchardt: fixed

ui/src/tugboat/components/Saliency.tsx
jonborchardt: fixed

ui/src/tugboat/components/Saliency.tsx
jonborchardt: fixed

ui/src/tugboat/components/form/Output.tsx
jonborchardt: fixed

ui/src/tugboat/components/form/Output.tsx
jonborchardt: fixed

ui/src/tugboat/components/form/Output.tsx
jonborchardt: removed

ui/src/tugboat/components/form/Output.tsx
jonborchardt: so, ya, you are right... but i dont know how to do this....

the popup is outside of this part of the dom

skipping this for now

Issue #679, opened 2020-12-23
None of the attackers or interpreters work for the Transformer QA model.


codeviking:
### Task
Reading Comprehension

### Model
Transformer QA

### Feature
All model attacks and interpretations.

### Issue
Trying to run an attack or interpretation for the transformer QA model results in a blank white page (this is what happens when an error is thrown in the demo).

I believe it's because transformer QA doesn't support these types of requests.

PR #681, opened 2020-12-23
Use the model info to toggle interpreters on and off.

codeviking:
This uses the list of `intepreters` in the `ModelInfo` returned
for the current model to toggle the interpreters that are displayed.

After this change when Transformer QA is selected, the option to
interpret the results no longer displays (which avoids part of this
bug: https://github.com/allenai/allennlp-demo/issues/679).

On the other hand when models like BiDAF or ELMo-BiDAF are selected,
the option to invoke each fo the interpreters remains (since they're
supported for those models).


jonborchardt (approved):
looks great
(we overlaped again on much of this)

PR #685, opened 2020-12-24
Add support for model attacks.

codeviking:
This wires up model attacks in the new demo architecture. As part of
this effort I discovered that the use of `<Form />` was causing some
interesting side-effects. Specifically, the `useEffect()` calls in
`<Form />` were mutating the data in the `<Interpret />` and
`<Attack />` instances. In retrospect this makes sense, and might
suggest that the `<Form />` name is too generic.

I left the `<Form />` abstraction alone and instead decided to wire
`<Attack />` and `<Interpret />` up as straight XHR calls (via
`<Promised />`. This makes it clear there's no user-input in either
case. The `<Interpret />` interface wasn't changed -- as the two
implementation are functionally equivalent.

### DIFF
--- a/ui/src/components/Attack.tsx
+++ b/ui/src/components/Attack.tsx
@@ -25,9 +25,10 @@ interface Props<I, O> {
     target: keyof I & string;
     gradient: GradientInputField;
     children: (io: { input: AttackRequest<I>; output: O }) => React.ReactNode | JSX.Element;
+    action: string;
 }
 
-export const Attack = <I, O>({ type, input, children, target, gradient }: Props<I, O>) => {
+export const Attack = <I, O>({ type, input, children, target, gradient, action }: Props<I, O>) => {
     const [submitted, setSubmitted] = React.useState(false);
 
     const ctx = React.useContext(Models);
@@ -53,7 +54,7 @@ export const Attack = <I, O>({ type, input, children, target, gradient }: Props<
     return (
         <>
             <Button type="primary" onClick={() => setSubmitted(true)}>
-                Submit Attack
+                {action}
             </Button>
             {submitted ? (
                 <Promised input={attackInput} fetch={fetchAttackOutput}>
--- a/ui/src/components/Attack.tsx
+++ b/ui/src/components/Attack.tsx
@@ -41,13 +41,13 @@ export const Attack = <I, O>({ type, input, children, target, gradient, action }
         grad_input_field: gradient,
         inputs: input,
     };
-    const fetchAttackOutput = (i?: AttackRequest<I>) => {
-        if (!i) {
+    const fetchAttackOutput = (attack?: AttackRequest<I>) => {
+        if (!attack) {
             throw new EmptyAttackRequestError();
         }
         return fetch(`/api/${model.id}/attack/${type}`, {
             method: 'POST',
-            body: JSON.stringify(i),
+            body: JSON.stringify(attack),
         }).then((r) => r.json());
     };
 
--- a/ui/src/components/Attack.tsx
+++ b/ui/src/components/Attack.tsx
@@ -25,10 +25,10 @@ interface Props<I, O> {
     target: keyof I & string;
     gradient: GradientInputField;
     children: (io: { input: AttackRequest<I>; output: O }) => React.ReactNode | JSX.Element;
-    action: string;
+    label: string;
 }
 
-export const Attack = <I, O>({ type, input, children, target, gradient, action }: Props<I, O>) => {
+export const Attack = <I, O>({ type, input, children, target, gradient, label }: Props<I, O>) => {
     const [submitted, setSubmitted] = React.useState(false);
 
     const ctx = React.useContext(Models);
@@ -54,7 +54,7 @@ export const Attack = <I, O>({ type, input, children, target, gradient, action }
     return (
         <>
             <Button type="primary" onClick={() => setSubmitted(true)}>
-                {action}
+                {label}
             </Button>
             {submitted ? (
                 <Promised input={attackInput} fetch={fetchAttackOutput}>
--- a/ui/src/components/Attack.tsx
+++ b/ui/src/components/Attack.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import { Button } from 'antd';
+
+import { Promised } from '../tugboat/components';
+import { Models } from '../tugboat/context';
+import { NoSelectedModelError } from '../tugboat/error';
+
+import { AttackType, GradientInputField } from '../lib';
+
+class EmptyAttackRequestError extends Error {
+    constructor() {
+        super('An attack request cannot be empty.');
+    }
+}
+
+interface AttackRequest<I> {
+    input_field_to_attack: keyof I & string;
+    grad_input_field: GradientInputField;
+    inputs: I;
+}
+
+interface Props<I, O> {
+    input: I;
+    type: AttackType;
+    target: keyof I & string;
+    gradient: GradientInputField;
+    children: (io: { input: AttackRequest<I>; output: O }) => React.ReactNode | JSX.Element;
+}
+
+export const Attack = <I, O>({ type, input, children, target, gradient }: Props<I, O>) => {
+    const [submitted, setSubmitted] = React.useState(false);
+
+    const ctx = React.useContext(Models);
+    if (!ctx.selectedModel) {
+        throw new NoSelectedModelError();
+    }
+    const model = ctx.selectedModel;
+    const attackInput: AttackRequest<I> = {
+        input_field_to_attack: target,
+        grad_input_field: gradient,
+        inputs: input,
+    };
+    const fetchAttackOutput = (i?: AttackRequest<I>) => {
+        if (!i) {
+            throw new EmptyAttackRequestError();
+        }
+        return fetch(`/api/${model.id}/attack/${type}`, {
+            method: 'POST',
+            body: JSON.stringify(i),
+        }).then((r) => r.json());
+    };
+
+    return (
+        <>
+            <Button type="primary" onClick={() => setSubmitted(true)}>
+                Submit Attack
+            </Button>
+            {submitted ? (
+                <Promised input={attackInput} fetch={fetchAttackOutput}>
+                    {children}
+                </Promised>
+            ) : null}
+        </>
+    );
+};
--- a/ui/src/components/Attacks.tsx
+++ b/ui/src/components/Attacks.tsx
@@ -14,7 +14,7 @@ interface Props<I> {
     target: keyof I & string;
 }
 
-export const Attacks = <I,>({ model, input, target }: Props<I>) => {
+export const Attacks = <I, O>({ model, input, target }: Props<I>) => {
     const modelInfoList = React.useContext(ModelInfoList);
 
     const info = modelInfoList.find((i) => i.id === model.id);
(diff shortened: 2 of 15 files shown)


ui/src/components/Interpret.tsx
```diff
@@ -1,28 +1,56 @@
...
+                Interpret Prediction
+            </Button>
+            {submitted ? (
+                <Promised input={input} fetch={fetchInterpretOutput}>
+                    {children}
+                </Promised>
+            ) : null}
+        </>
+    );
+};
```
codeviking, line 56: @jonborchardt Let's merge you work first this time, I can resolve conflicts (the description stuff should be easy to port over to this, I think.

For that reason we might hold off on toggling the `description` property in favor of `children` until later. If you like the idea maybe I can knock it out after both your changes and this merge.  We can discuss in the new year 🍾 .

ui/src/tugboat/error/index.ts
```diff
@@ -40,7 +40,7 @@ export class InvalidHighlightRangeError extends Error {
     }
 }
 
-export class InvalidExamplesFormat extends Error {
+export class InvalidExamplesFormatError extends Error {
```
codeviking, line 43: The changes in this file are just a bit of cleanup. The other errors in this file all end in `Error` but these three didn't (this inconsistency was my doing).

ui/src/components/Attack.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface AttackRequest<I> {
+    input_field_to_attack: keyof I & string;
+    grad_input_field: GradientInputField;
+    inputs: I;
+}
+
+interface Props<I, O> {
+    input: I;
+    type: AttackType;
+    target: keyof I & string;
```
codeviking, line 25: This type is *awesome*. Another 💯  for TypeScript!

ui/src/components/Attack.tsx
```diff
@@ -0,0 +1,65 @@
...
+        return fetch(`/api/${model.id}/attack/${type}`, {
+            method: 'POST',
+            body: JSON.stringify(i),
+        }).then((r) => r.json());
+    };
+
+    return (
+        <>
+            <Button type="primary" onClick={() => setSubmitted(true)}>
+                Submit Attack
```
jonborchardt, line 56: is this really the name that people use?
could we call it something else?

ui/src/components/Interpret.tsx
jonborchardt: there is some non trivial refactoring required from pr suggestions.  otherwise, i would recommend just merging as is.
we can merge on jan 2nd i bet.

ui/src/demos/reading-comprehension/Attacks.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { AttackType, GradientInputField } from '../../lib';
+import { Input } from './types';
+
+interface Props {
+    model: Model;
+    input: Input;
+}
+
+/**
+ * TODO: Bits and pieces of this can and should move into `../../components` so that other demos
```
jonborchardt, line 18: none of this is specific to reading comp....
lets make this general

ui/src/demos/reading-comprehension/Interpreters.tsx
```diff
@@ -36,6 +36,7 @@ export const Interpreters = ({ model, input }: Props) => {
                     <Collapse.Panel
                         key={InterpreterId.SimpleGradient}
                         header="Simple Gradient Visualization">
+                        {/* TODO: Replace `any` with a better type once we know what it is. */}
                         <Interpret<Input, any>
```
jonborchardt, line 40: my pr has the type

ui/src/tugboat/error/index.ts
jonborchardt: consider moving these errors closer to where the are used, if they are specific to examples, for example...

jonborchardt (approved):

jonborchardt:
feel free to merging to master, i dont mind merging my changes.... i have some refactoring to do anyway....

ui/src/components/Interpret.tsx
codeviking: Okie dokie. Sounds good 👍

ui/src/tugboat/error/index.ts
codeviking: Sure, can do.

ui/src/demos/reading-comprehension/Attacks.tsx
codeviking: Sure. When I started it it wasn't as generic, but it is now -- good call!

ui/src/demos/reading-comprehension/Interpreters.tsx
codeviking: Great!

ui/src/demos/reading-comprehension/Attacks.tsx
codeviking: Done.

ui/src/components/Attack.tsx
codeviking: Yea, good call. I preserved the text that's used now, which is attack specific.

ui/src/components/Attack.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+class EmptyAttackRequestError extends Error {
+    constructor() {
+        super('An attack request cannot be empty.');
+    }
+}
+
+interface AttackRequest<I> {
+    input_field_to_attack: keyof I & string;
+    grad_input_field: GradientInputField;
```
jonborchardt, line 18: yuck, i sure wish we could solve the case issue in a reasonable way

ui/src/components/Attack.tsx
```diff
@@ -0,0 +1,66 @@
...
+        throw new NoSelectedModelError();
+    }
+    const model = ctx.selectedModel;
+    const attackInput: AttackRequest<I> = {
+        input_field_to_attack: target,
+        grad_input_field: gradient,
+        inputs: input,
+    };
+    const fetchAttackOutput = (i?: AttackRequest<I>) => {
+        if (!i) {
```
jonborchardt, line 45: longer prop name?
(can we lint this?)

ui/src/components/Attack.tsx
```diff
@@ -0,0 +1,66 @@
...
+        return fetch(`/api/${model.id}/attack/${type}`, {
+            method: 'POST',
+            body: JSON.stringify(i),
+        }).then((r) => r.json());
+    };
+
+    return (
+        <>
+            <Button type="primary" onClick={() => setSubmitted(true)}>
+                {action}
```
jonborchardt, line 57: its unclear that `action` is going to be used as a label...

ui/src/components/Attacks.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    const supportedAttackTypes = new Set(info.attackers);
+
+    return (
+        <Output.Section title="Model Attacks">
+            <Collapse>
+                {supportedAttackTypes.has(AttackType.InputReduction) ? (
+                    <Collapse.Panel key={AttackType.InputReduction} header="Input Reduction">
+                        {/* TODO: Replace `any` with a better type once we know what it is. */}
```
jonborchardt, line 32: when we have this in multiple places... we could define a type 
`type AttackResponse = any; // TODO: Replace `any` with a better type once we know what it is.`

ui/src/components/Attacks.tsx
```diff
@@ -0,0 +1,59 @@
...
+                        <Attack<I, any>
+                            type={AttackType.InputReduction}
+                            target={target}
+                            gradient={GradientInputField.Input2}
+                            input={input}
+                            action="Reduce Input">
```
jonborchardt, line 38: ya, lets rename `action` to `label`

ui/src/components/Interpret.tsx
```diff
@@ -1,28 +1,56 @@
...
+        super('An interpret request cannot be empty.');
+    }
+}
 
 interface Props<I, O> {
     input: I;
-    interpreter: string;
-    children: (interpretation: Result<I, O>) => React.ReactNode | JSX.Element;
+    interpreter: InterpreterId;
+    children: (io: { input: I; output: O }) => React.ReactNode | JSX.Element;
```
jonborchardt, line 19: we could add a type for this
`type ReactChildren = React.ReactNode | JSX.Element`

ui/src/components/Interpret.tsx
```diff
@@ -1,28 +1,56 @@
...
+    if (!ctx.selectedModel) {
+        throw new NoSelectedModelError();
+    }
+    const model = ctx.selectedModel;
+    const fetchInterpretOutput = (i?: I) => {
+        if (!i) {
+            throw new EmptyInterpretRequestError();
+        }
+        return fetch(`/api/${model.id}/interpret/${interpreter}`, {
+            method: 'POST',
```
jonborchardt, line 39: i would still prefer we use consts... or even a library of consts
FetchMethod.Post
but i think ill loose this battle ;-P

ui/src/demos/reading-comprehension/Interpreters.tsx
```diff
@@ -36,6 +36,7 @@ export const Interpreters = ({ model, input }: Props) => {
                     <Collapse.Panel
                         key={InterpreterId.SimpleGradient}
                         header="Simple Gradient Visualization">
+                        {/* TODO: Replace `any` with a better type once we know what it is. */}
```
jonborchardt, line 39: see my other comment about making a placeholder type

jonborchardt (approved):
i left a few nit picks...
looks great!

ui/src/components/Interpret.tsx
jonborchardt: https://www.npmjs.com/package/http-methods-enum

ui/src/components/Attack.tsx
jonborchardt: https://stackoverflow.com/questions/44437953/angular-typescript-converting-from-snake-case-to-camel-case-in-interfaces

ui/src/components/Interpret.tsx
jonborchardt: https://www.npmjs.com/package/mime-types

ui/src/components/Attack.tsx
jonborchardt: discussed offline, letting this go ;-P

ui/src/components/Attack.tsx
codeviking: Good call.

codeviking: Updated to `attack`.

ui/src/components/Attack.tsx
codeviking: Good point. Maybe `actionText` or `actionLabel`, or even just `label`?

ui/src/components/Attacks.tsx
codeviking: Ah, I like that!

ui/src/components/Interpret.tsx
codeviking: Yea, I've resisted the urge to do this because I feel like it's something React should provide. But...React isn't TypeScript, so they're not going to! 

*Silly Sam, Trix are for kids!*

ui/src/components/Interpret.tsx
codeviking: I don't like the idea of installing a package solely for a few string values that are well known and *very* unlikely to change. Our `node_modules/` tree is already deep enough.

...but I'm amenable to the idea of consolidating these somewhere. I'll try something out and see how it feels.

ui/src/demos/reading-comprehension/Interpreters.tsx
codeviking: 👍

ui/src/components/Attacks.tsx
codeviking: Done!

ui/src/components/Interpret.tsx
codeviking: I decided to leave this as is right now.

- The `'Content-Type': 'application/json'` bit is only in old code. So that'll go away. 
- The  duplication of `method: 'POST'` bit feels right to me, as the `POST` value isn't something we'll need to change enmasse, so I don't think we gain much by sourcing it from a single definition.

PR #691, opened 2021-01-04
Simpler form output.

codeviking:
This simplifies the `<Form />` mechanism by removing the bits that
relied on the structure of `props.children`. This is both simpler
and typesafe. My prior incantation would fail at runtime in
hard-to-decipher ways.


jonborchardt (approved):
nice

PR #692, opened 2021-01-05
add naqanet answer types

jonborchardt:
this adds ui for all answer types in naqanet
we could make some of these more reusable in tugboat... ie
export PassageContextAnswer
but i opted to not do that for now


ui/src/demos/reading-comprehension/types.ts
```diff
@@ -61,11 +61,20 @@ export enum NAQANetAnswerType {
...
+    sign: number;
+}
+
 export interface NAQANetPrediction extends WithTokenizedInput {
     answer: {
-        'answer-type': NAQANetAnswerType;
-        spans: number[];
-        value: string;
+        answer_type: NAQANetAnswerType;
+        // NOTE: the backend returns very different responses based on the type
```
jonborchardt, line 73: @codeviking any thoughts on how to make this cleaner?

ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -110,25 +117,143 @@ const BasicPrediction = ({
...
+                    <Output.SubSection title="Answer">
+                        <div>{output.answer.value}</div>
+                    </Output.SubSection>
+
+                    <Output.SubSection title="Explanation">
+                        {output.answer.numbers ? (
+                            <div>
+                                The model used the arithmetic expression{' '}
+                                <ArithmeticEquation numbersWithSign={output.answer.numbers} /> ={' '}
+                                {output.answer.value}.
```
codeviking, line 221: Should we use `<code />` and/or `<pre />` here, so that the text uses a monospace font?

ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -110,25 +117,143 @@ const BasicPrediction = ({
...
+                            <div>The model decided this was an arithmetic problem.</div>
+                        )}
+                    </Output.SubSection>
+
+                    <Output.SubSection title="Passage Context">
+                        {output.answer.numbers ? (
+                            <TextWithHighlight
+                                text={input.passage}
+                                highlights={output.answer.numbers
+                                    .filter((n) => n.sign !== 0)
```
codeviking, line 233: Do we know what these represent (i..e could we add a command indicating why we filter them)?

ui/src/demos/reading-comprehension/types.ts
aimichal: My two cents:

One idea is to have a function like `getNAQANetAnswer(answer: GeneralAnswer) : NAQANetAnswer { ... }` which returns a NAQANetAnswer or throws an exception. Then call it in when you'd expect a NAQANetAnswer.

There may be something more puzzle-y with TypeScript that makes handling polymorphic responses possible in the general case. I think there isn't, but if there is, it's _probably_ not worth the code complexity.

(I don't know what's in the backend, but if there's a single URL that's returning different responses, maybe that should be split to different URLs.)

ui/src/tugboat/components/ArithmeticEquation.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+interface NumberWithSign {
+    value: number;
+    span: [number, number];
+    sign: number;
+}
```
aimichal, line 7: BTW, [Math.sign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) can be used to neatly extract the sign of a number. I'm not sure what's happening in this ArithmeticEquation component (what's the "span" field for?), but maybe that can be used to simplify things.

ui/src/demos/reading-comprehension/types.ts
codeviking: I'd use a type for each and type-guards to safely process each as the more specific type.

So, let's say there's two types, one that has `spans` and another has `value`, `numbers` and `count`. I'd express them like:

```
abstract class NAQANetPrediction {
   constructor(readonly answer_type: NAQANetAnswerType) {}
   
  isTextPrediction(): this is NAQANetTextPrediction {
     return this instanceof NAQANetTextPrediction;
  }
  isAlgebraicPrediction(): this is NAQANetAlgebraicPrediction {
     return this instanceof NAQANetAlgebraicPrediction;
  }
}

class NAQANetTextPrediction extends NAQANetPrediction {
   constructor(spans: [ number, number ][]) {}
}

class NAQANetAlgebraicPrediction extends NAQANetPrediction {
   constructor(value: string, numbers: NumberWithSign[], count: number) {}
}
```

Then you can safely do something like this:

```
if (prediction.isTextPrediction()) {
   return <HighlightedText ... />
} else if (prediction.isAlgebraicPrediction()) {
   return <FancyAlebgra .. />
}
```

ui/src/tugboat/components/ArithmeticEquation.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface Props {
+    numbersWithSign: NumberWithSign[];
+}
+
+export const ArithmeticEquation = ({ numbersWithSign }: Props) => {
+    if (numbersWithSign) {
```
codeviking, line 14: According to the props the type can't be `undefined` (or "falsy"), so do we need this check?

ui/src/tugboat/components/ArithmeticEquation.tsx
```diff
@@ -0,0 +1,25 @@
...
+        let ret = numbersWithSign
+            .filter((n) => n.sign !== 0)
+            .map((n) => `${n.sign > 0 ? '+' : '-'} ${n.value}`)
+            .join(' ');
+        while (ret.charAt(0) === '+' || ret.charAt(0) === ' ') {
+            ret = ret.substr(1);
+        }
+        return <span>{ret}</span>;
```
codeviking, line 22: I think this would be easier to follow if it were imperative:

```
let expression = '';
for (const { sign, number } of numbersWithSign) {
  // Filter expressions that evaluate to 0
  if (sign === 0 ) {
     continue;
  }
      
   // Trim leading "+" characters
   if (sign === '+' && str.length == 0) {
     expression = `{number}`;
     continue;
   } 

   expression += `${expression} {sign} {number}`
}
```

Then, to removing leading or trailing space use `trim()` where it's rendered:

```
{expression.trim()}
```

This also has the advantage of doing everything in a single pass, whereas the current code loops over the code 3 times, I think.

codeviking (approved):
LGTM; I left some ideas.

ui/src/demos/reading-comprehension/Predictions.tsx
jonborchardt: er.. not here, but im going to move the 'answer' into the ArithmeticEquation and may style that with a code

jonborchardt: done

ui/src/demos/reading-comprehension/Predictions.tsx
jonborchardt: i added a comment, making a func seemed odd

ui/src/tugboat/components/ArithmeticEquation.tsx
jonborchardt: this is how the data is passed from the backend, we dont have the signed number, rather, we have a struct with a number(positive) and a sign (-1, 0, 1)

ui/src/tugboat/components/ArithmeticEquation.tsx
jonborchardt: fixed... much like this, but a bit more
let me know WYT

ui/src/tugboat/error/index.ts
```diff
@@ -22,6 +22,12 @@ export class UnexpectedModelError extends Error {
     }
 }
 
+export class InvalidModelResponseError extends Error {
+    constructor(modelId: string) {
+        super(`The type of the response does not match the model: ${modelId}`);
```
codeviking, line 27:
```suggestion
        super(`Unexpected response from model: ${modelId}`);
```

codeviking (approved):
🤘

PR #699, opened 2021-01-06
Jon/input reduction UI

jonborchardt:
image


ui/src/components/Attack.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO: move to tugboat?
```
jonborchardt, line 1: opinions?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,109 @@
+// TODO: move to tugboat?
```
jonborchardt, line 1: opinions?

ui/src/components/Tokens.tsx
```diff
@@ -0,0 +1,32 @@
+// TODO: move to tugboat?
```
jonborchardt, line 1: optnions?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import { RedToken, TransparentToken, BlankToken } from './Tokens';
+
+// takes in the input before and after input reduction and highlights
+// the words that were removed in red with strikeout. The reducedInput
+// will have a less than or equal to length than the originalInput.
```
codeviking, line 6: Let's use the `/** */` syntax for these comments, and try to make them more formal (capitalization n' all that) for consistency with what we're doing elsewhere.

Here's an example:
https://github.com/allenai/allennlp-demo/blob/master/ui/src/tugboat/lib/usePromise.ts#L79

jonborchardt: ya, i forgot, sorry... many of these changes are copy pasta

jonborchardt: fixed

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { RedToken, TransparentToken, BlankToken } from './Tokens';
+
+/**
+ * Takes in the input before and after input reduction and highlights
+ * the words that were removed in red with strikeout. The reducedInput
+ * will have a less than or equal to length than the originalInput.
+ */
```
codeviking, line 8: I'm still processing the fact that we do this diff on the client. I can't decide if I think that's a good thing or not.

I think it'd be better if the server returned something that told us what the removed tokens are, i.e.:

```json
[ { "token": "The", "removed": false }, { "token": "quick", "removed": true } ]
```

WDYT? Maybe we just add a `TODO` for later?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+const colorizeTokensForInputReductionUI = (originalInput: string[], reducedInput: string[]) => {
+    const originalStringColorized = [];
+    const reducedStringColorized = [];
+    let idx = 0;
+    let idx2 = 0;
```
codeviking, line 13: Maybe `originalIdx` and `reducedIdx`?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+        if (originalInput[idx] === reducedInput[idx2]) {
+            originalStringColorized.push(
+                <TransparentToken key={idx}>{originalInput[idx]}</TransparentToken>
+            );
+            reducedStringColorized.push(
+                <TransparentToken key={idx}>{reducedInput[idx2]}</TransparentToken>
+            );
+            idx++;
+            idx2++;
+        } else {
```
codeviking, line 25: So, something that I've adopted from golang that I like is to try to keep things as "flat" as possible, by minimizing branching.

I think you could do that here by using a `continue` in this block, which means you can de-indent the stuff in the `else` block:

```
if (originalInput[idx] == reducedInput[idx2]) {
   /* ... */
  continue';
}

// They're not the same
...
```

It's a small thing in practice, but I like the larger effect.

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+            /**
+             * If the words are not equal, keep traversing along the original list, making
+             * each token red as you go along. Also add a blank placeholder into the
+             * reduced list to make the words line up spacing wise in the UI.
+             */
```
codeviking, line 30: Since this isn't formal documentation of an API, I'd use a multi-line `//` here. Sorry for the confusion.

The `/** */` syntax is meant to capture documentation for public APIs -- i.e. what we'd capture if we were to automatically generate documentation for the code and distribute it.

Inline comments like this should use `//` to be consistent with the rest of the codebase.

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { RedToken, TransparentToken, BlankToken } from './Tokens';
+
+/**
+ * Takes in the input before and after input reduction and highlights
+ * the words that were removed in red with strikeout. The reducedInput
+ * will have a less than or equal to length than the originalInput.
+ */
+const colorizeTokensForInputReductionUI = (originalInput: string[], reducedInput: string[]) => {
```
codeviking, line 9: I wonder if `highlightRemovedTokens` might be more clear?

In fact, could this be a component of it's own:

```
const RemovedTokens = ({ original, reduced }) => (
   ...
);

<RemovedTokens original={...} reduced={...} />
```

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export const InputReduction = (reducedInput: InputReductionAttackOutput) => {
+    /**
+     * There are a number of ways to tweak the output of this component:
+     * (1) you can provide a context, which shows up on top, e.g., for displaying the
+     * premise for SNLI.
+     * (2) you can format the original input and the reduced input yourself, to
+     * customize the display for, e.g., NER.
+     */
```
codeviking, line 64: This is another spot where a multi-line `//` comment were probably be more appropriate.

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const original = reducedInput.original;
+    const formattedOriginal = reducedInput.formattedOriginal;
+    return (
+        <div>
+            {reducedInput.context ? <span>{reducedInput.context}</span> : null}
+
+            {formattedOriginal ? <span>{formattedOriginal}</span> : null}
```
codeviking, line 71: Why do we capture `formattedOriginal` as a local variable, but not the others?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    ) : null}
+
+                    {(reducedInput.formattedReduced || []).map((formattedReduced: string) => (
+                        <span key={formattedReduced}>{formattedReduced}</span>
+                    ))}
+                </>
+            ) : (
+                <>
+                    {reducedInput.final.map((reduced: string[]) => {
+                        const noReduction = JSON.stringify(original) === JSON.stringify(reduced);
```
codeviking, line 88: Serializing JSON is faster than parsing it, but this is still likely to be slow.

Would it suffice to count the number of tokens?

ui/src/components/Tokens.tsx
```diff
@@ -0,0 +1,30 @@
+import styled from 'styled-components';
+
+export const ColorizedToken = styled.span<{ backgroundColor?: string }>`
+    background-color: ${({ backgroundColor }) => backgroundColor};
```
codeviking, line 4: I'd use `background` for brevity -- bot for the css property and the property:

```
<ColorizedToken background="g6" />
```

Also, `<Colorized />` doesn't sit quite write with me. Don't we call this `<Highlighted />` elsewhere? Should we stick with that?

ui/src/components/Tokens.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+// green token used to represent addition in HotFlip
+export const GreenToken = styled(ColorizedToken)`
+    background-color: ${({ theme }) => theme.color.G5};
+`;
+
+// transparent token used to have correct spacing of elements in HotFlip
+export const TransparentToken = styled(ColorizedToken)`
+    background-color: 'transparent';
+`;
```
codeviking, line 24: Do we explicitly need to set the background color to transparent? Could we just use a `<span />` and it'd have the same effect?

ui/src/components/Tokens.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+// red token used to represent deletion in InputReduction and replacement in HotFlip
+export const RedToken = styled(ColorizedToken)`
+    background-color: ${({ theme }) => theme.color.R5};
+`;
+
+// green token used to represent addition in HotFlip
+export const GreenToken = styled(ColorizedToken)`
+    background-color: ${({ theme }) => theme.color.G5};
+`;
```
codeviking, line 19: With the addition of the `backgroundColor` property do we still need `<RedToken />` and `<GreenToken />`. If we do, we could use `attrs` I think instead of setting the css property again:

```
const GreenToken = styled(ColorizedToken).attrs(() => ({ backgroundColor: 'G5' })``;
```

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+                                    <p>
+                                        <h6>Original Input:</h6> {originalColored}
+                                    </p>
+                                ) : null}
+
+                                <p>
+                                    <h6>Reduced Input:</h6> {reducedColored}
+                                </p>
+
+                                {noReduction ? <p>(No reduction was possible)</p> : null}
```
codeviking, line 105: Should this be called out as a warning or error of some sort -- maybe using the `<Alert />` component?

ui/src/components/InputReduction.tsx
```diff
@@ -0,0 +1,113 @@
...
+                            </>
+                        );
+                    })}
+                </>
+            )}
+        </div>
+    );
+};
```
codeviking, line 113: The large ternary comparison here with lots of nesting makes it a little hard to parse what's going on.

I'd suggest breaking the individual treatments into local components, and doing something like this:

```
if (reducedInput.formattedReduced) {
   return <PreformattedInputReductionOutput ... />
} else {
   return <DefaultInputReductionOutput />
}
```

The names I've proposed are bad -- but the idea is to:

1. Move specific display logic to individual components.
2. Branch at one level in the parent.

WDYT?

codeviking (approved):
Nice work; I left a few ideas.

ui/src/components/InputReduction.tsx
jonborchardt: i will add a todo, i dont plan on making ANY api changes as part of this work

ui/src/components/InputReduction.tsx
jonborchardt: we should discuss. i do not buy that that is cleaner 
it is essentially short-circuiting which is a nono in my book.
i can see the benefit if you had a deep nest, but in this case, we have a single else
i am going to leave as is unless you feel strongly

ui/src/components/InputReduction.tsx
jonborchardt: ahh.. ok

ui/src/components/InputReduction.tsx
jonborchardt: ill rename, but its output used in a complicated way, it cannot be a component as of now

ui/src/components/InputReduction.tsx
jonborchardt: copy pasta... ill fix up

ui/src/components/Tokens.tsx
jonborchardt: we want to preserve the padding and margin etc, so things match up vertically....

ui/src/components/Tokens.tsx
jonborchardt: is this better in some way?

ui/src/components/InputReduction.tsx
jonborchardt: its not an error.. its just data

ui/src/components/InputReduction.tsx
jonborchardt: done

ui/src/components/InputReduction.tsx
jonborchardt: reverted

ui/src/components/InputReduction.tsx
jonborchardt: renamed

ui/src/components/InputReduction.tsx
jonborchardt: reverted

ui/src/components/InputReduction.tsx
jonborchardt: fixed

ui/src/components/InputReduction.tsx
jonborchardt: yes, done

ui/src/components/Tokens.tsx
jonborchardt: both done

ui/src/components/Tokens.tsx
jonborchardt: leaving for now

ui/src/components/InputReduction.tsx
jonborchardt: leaving as is for now

ui/src/components/InputReduction.tsx
jonborchardt: ya, this blob sucks... ill see about reworking it a bit

jonborchardt: done

PR #701, opened 2021-01-07
hotflip ui code

jonborchardt:
image

note. i have an open issue with the allennlp team about why the revised prediction is not returning


ui/src/components/Hotflip.tsx
```diff
@@ -0,0 +1,65 @@
...
+    final: string[][];
+    original: string[];
+    new_prediction?: string;
+    context?: string;
+}
+
+export const Hotflip = ({ final, original, new_prediction, context }: HotflipAttackOutput) => {
+    const [originalString, flippedString] = highlightFlippedTokens(
+        original,
+        final[0] // just the first one, huh,... yuck
```
aimichal, line 44: `// just the first one, because the others XXX`

XXX = have a lower confidence?

ui/src/demos/reading-comprehension/types.ts
```diff
@@ -163,3 +163,14 @@ export type Prediction =
     | TransformerQAPrediction
     | NAQANetPrediction
     | NMNPrediction;
+
+export const isPrediction = (x: Prediction): x is Prediction => {
+    const xx = x as Prediction;
```
aimichal, line 168: change `x` to `pred` and `xx` to `typedPred`?

Or explain what is happening, because it's not clear (to me).

(I'm probably just be too out of touch with TS tho, so up to you!)

aimichal (approved):

ui/src/components/Hotflip.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+
+import { RedToken, GreenToken, TransparentToken } from './Tokens';
+
+// NOTE: The original HotFlipComponent handled targeted words to flip.  This was used in the
```
codeviking, line 5:
```suggestion
// NOTE: The original HotFlipComponent handled targeted words to flip.  This was only used in the
```

I think that's the case, let me know if it's not.

codeviking (approved):

ui/src/components/Hotflip.tsx
jonborchardt: i have no idea.... there is no conf displayed

ui/src/demos/reading-comprehension/types.ts
jonborchardt: ill update in a followup pr
the x's here dont matter... but ill do as requests, it makes sense

PR #702, opened 2021-01-07
add model description

jonborchardt:
added the description of the selected model

changed the dropdown to use the short description if availiable

pr clean up of typeguards

image

### DIFF
--- a/ui/src/demos/reading-comprehension/types.ts
+++ b/ui/src/demos/reading-comprehension/types.ts
@@ -16,18 +16,18 @@ export interface BiDAFPrediction extends WithTokenizedInput {
     token_offsets: number[][];
 }
 
-export const isBiDAFPrediction = (x: Prediction): x is BiDAFPrediction => {
-    const xx = x as BiDAFPrediction;
+export const isBiDAFPrediction = (pred: Prediction): pred is BiDAFPrediction => {
+    const typedPred = pred as BiDAFPrediction;
     return (
-        isWithTokenizedInput(x) &&
-        xx.best_span !== undefined &&
-        xx.best_span_str !== undefined &&
-        xx.passage_question_attention !== undefined &&
-        xx.span_end_logits !== undefined &&
-        xx.span_end_probs !== undefined &&
-        xx.span_start_logits !== undefined &&
-        xx.span_start_probs !== undefined &&
-        xx.token_offsets !== undefined
+        isWithTokenizedInput(pred) &&
+        typedPred.best_span !== undefined &&
+        typedPred.best_span_str !== undefined &&
+        typedPred.passage_question_attention !== undefined &&
+        typedPred.span_end_logits !== undefined &&
+        typedPred.span_end_probs !== undefined &&
+        typedPred.span_start_logits !== undefined &&
+        typedPred.span_start_probs !== undefined &&
+        typedPred.token_offsets !== undefined
     );
 };
 
@@ -41,16 +41,16 @@ export interface TransformerQAPrediction {
     span_start_logits: number[];
 }
 
-export const isTransformerQAPrediction = (x: Prediction): x is TransformerQAPrediction => {
-    const xx = x as TransformerQAPrediction;
+export const isTransformerQAPrediction = (pred: Prediction): pred is TransformerQAPrediction => {
+    const typedPred = pred as TransformerQAPrediction;
     return (
-        xx.best_span !== undefined &&
-        xx.best_span_scores !== undefined &&
-        xx.best_span_str !== undefined &&
-        xx.context_tokens !== undefined &&
-        xx.id !== undefined &&
-        xx.span_end_logits !== undefined &&
-        xx.span_start_logits !== undefined
+        typedPred.best_span !== undefined &&
+        typedPred.best_span_scores !== undefined &&
+        typedPred.best_span_str !== undefined &&
+        typedPred.context_tokens !== undefined &&
+        typedPred.id !== undefined &&
+        typedPred.span_end_logits !== undefined &&
+        typedPred.span_start_logits !== undefined
     );
 };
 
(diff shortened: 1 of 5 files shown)


ui/src/demos/reading-comprehension/types.ts
```diff
@@ -16,18 +16,18 @@ export interface BiDAFPrediction extends WithTokenizedInput {
     token_offsets: number[][];
 }
 
-export const isBiDAFPrediction = (x: Prediction): x is BiDAFPrediction => {
-    const xx = x as BiDAFPrediction;
+export const isBiDAFPrediction = (pred: Prediction): pred is BiDAFPrediction => {
```
codeviking, line 19: I wonder if you can omit the cast by doing this:

```
(pred: Prediction & Partial<BiDAFPrediction>) => {}
```

I think that means `pred` will have the type `Prediction` and a `T | undefined` for each property of `BiDAFPrediction`.

ui/src/tugboat/components/SelectModel.tsx
```diff
@@ -14,17 +15,38 @@ export const SelectModel = () => {
...
+                                    {m.card.short_description || m.card.description}
+                                </Markdown>
+                            </Select.Option>
+                        ))}
+                    </Select>
+                    {ctx.selectedModel ? (
+                        <Description>
+                            <Markdown>{ctx.selectedModel.card.description}</Markdown>
+                        </Description>
+                    ) : null}
```
codeviking, line 35: I think we should lift this into a separate component, `<SelectedModelDescription />` or `<Description />`.

This provides flexibility. If I implement a demo where I don't want to display the description I can simply omit that control.

This inclination is ultimately derived from the  concept of a *pure function*. I like the idea of *pure components* -- components that do one thing, and don't render things that aren't directly associated with it's name.

ui/src/tugboat/components/SelectModel.tsx
```diff
@@ -14,17 +15,38 @@ export const SelectModel = () => {
...
 };
+
+const Frame = styled.div`
+    border: 1px solid ${({ theme }) => theme.color.N4};
+    background: ${({ theme }) => theme.color.N2};
+    border-radius: ${({ theme }) => theme.shape.borderRadius.default};
+`;
+
+const Description = styled.div`
+    padding: ${({ theme }) => theme.spacing.md};
```
codeviking, line 49: Wouldn't a `<p />` do this for us?

ui/src/tugboat/components/SelectModel.tsx
```diff
@@ -14,17 +15,38 @@ export const SelectModel = () => {
...
+    max-height: 300px;
+    overflow-y: auto;
```
codeviking, line 51: The "More / Less" style treatment seems more natural here than scrolling.  Can we use that instead? I think `antd` provides one that's associated with the `<Typography />` component?

codeviking (commented):
LGTM; I left a few ideas.

codeviking (approved):
Oops, meant to approve.

ui/src/demos/reading-comprehension/types.ts
jonborchardt: ill try it

ui/src/tugboat/components/SelectModel.tsx
jonborchardt: i want it IN the field for look and feel... what if i expose a prop to show/hide selected model desc?

ui/src/tugboat/components/SelectModel.tsx
jonborchardt: i want left and right padding also, so no, i dont think so

ui/src/tugboat/components/SelectModel.tsx
jonborchardt: no, because this will be markdown... more/less wont work for markdown

this only shows up when the screen is narrow, i could even add it to a below breakpoint

ui/src/tugboat/components/SelectModel.tsx
codeviking: Oops. I didn't catch that. I read this as `margin`. That makes sense.

ui/src/tugboat/components/SelectModel.tsx
codeviking: I wouldn't expect that to be a property of the `<SelectModel />`.   My litmus test is to think about what I'd expect with a native `<select />` element -- in which case it'd surprise me if that input did anything other than provide a selectable list of things to the user.

I'd propose a new "compound" component, `<SelectModelAndDescription />`. A better name is ok too :). The real point is we preserve the purity of the `<SelectModel />` component.

ui/src/tugboat/components/SelectModel.tsx
codeviking: Ah, gotcha. 

I wouldn't tie it to a breakpoint, as it's possible there's some weird resolution where we would or wouldn't need it. Letting the browser decide seems ok.

Go ahead as-is -- we can try it out and see how it feels. If we're doing a lot of scrolling, or if people aren't figuring out that they can scroll, then I think we'll want to think about another treatment.

ui/src/demos/reading-comprehension/types.ts
jonborchardt: no, this does not work, we can try again later if we have an uncontrollable urge

ui/src/tugboat/components/SelectModel.tsx
jonborchardt: done

ui/src/tugboat/components/SelectModel.tsx
jonborchardt: done

PR #706, opened 2021-01-07
add new prediction to bottom of hot flip

jonborchardt:
image

### DIFF
--- a/ui/src/components/Hotflip.tsx
+++ b/ui/src/components/Hotflip.tsx
@@ -6,6 +6,12 @@ import { RedToken, GreenToken, TransparentToken } from './Tokens';
 // masked-lm task. I have removed this functionality from this component.  We can either add it
 // back, or make a new HotflipComponent when we port masked-lm.
 
+class EmptyHotflipRequestError extends Error {
+    constructor() {
+        super('A Hotflip request cannot be empty.');
+    }
+}
+
 /**
  * Takes in the input before and after the hotflip attack and highlights
  * the words that were replaced in red and the new words in green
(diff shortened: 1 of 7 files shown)


ui/src/components/Hotflip.tsx
```diff
@@ -34,32 +34,40 @@ const highlightFlippedTokens = (originalInput: string[], flippedInput: string[])
...
+    best_span_str: string;
+    loss: number;
+    passage_question_attention: number[][];
+    passage_tokens: string[];
+    question_tokens: string[];
+    span_end_logits: number[];
+    span_end_probs: number[];
+    span_start_logits: number[];
+    span_start_probs: number[];
+    token_offsets: number[][];
```
codeviking, line 51: This looks like it's just another instance of the stuff that's in `types.ts` (the various `Output` subtypes).

Could we just use those types instead?

ui/src/components/Hotflip.tsx
```diff
@@ -34,32 +34,40 @@ const highlightFlippedTokens = (originalInput: string[], flippedInput: string[])
...
 }
 
-export const Hotflip = ({ final, original, new_prediction, context }: HotflipAttackOutput) => {
+export const Hotflip = ({ final, original, outputs, context }: HotflipAttackOutput) => {
+    const finalIndexToShow = 0;
     const [originalString, flippedString] = highlightFlippedTokens(
         original,
-        final[0] // just the first one, huh,... yuck
+        final[finalIndexToShow]
```
codeviking, line 58: What if `final` is empty?

ui/src/components/Hotflip.tsx
```diff
@@ -34,32 +34,40 @@ const highlightFlippedTokens = (originalInput: string[], flippedInput: string[])
...
-
-            {new_prediction ? (
-                <p>
-                    <h6>New Prediction:</h6> {new_prediction}
-                </p>
+            <h6>Original Input:</h6> {originalString}
+            <h6>Flipped Input:</h6> {flippedString}
+            {outputs.length && outputs[0].best_span_str ? (
+                <>
+                    <h6>New Prediction:</h6> {outputs[finalIndexToShow].best_span_str}
```
codeviking, line 68: I'm confused -- are `outputs` and `final` the same?

codeviking: Also, what if `outputs` is empty?

codeviking (approved):
LGTM; Seems like we may want to handle the case where the various lists we're pulling things out of are empty.

ui/src/components/Hotflip.tsx
jonborchardt: well... you are pointing out a larger issue
i need to fix this up
great find

ui/src/components/Hotflip.tsx
jonborchardt: 2 sibling arrays that are in `sync`

ui/src/components/Hotflip.tsx
jonborchardt: please see new code

ui/src/components/Hotflip.tsx
jonborchardt: fixed

ui/src/components/InputReduction.tsx
```diff
@@ -132,17 +132,13 @@ const DefaultOutput = ({
                 return (
                     <>
                         {!formattedOriginal ? (
-                            <p>
-                                <h6>Original Input:</h6> {originalColored}
-                            </p>
+                            <>
```
jonborchardt, line 135: while in rome, i fixed this

ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -82,6 +83,14 @@ const OutputByModel = ({
     throw new UnexpectedModelError(model.id);
 };
 
+const BasicAnswer = ({ output }: { output: Prediction }) => {
```
jonborchardt, line 86: while in rome, i added this

ui/src/components/Hotflip.tsx
```diff
@@ -31,35 +37,30 @@ const highlightFlippedTokens = (originalInput: string[], flippedInput: string[])
...
 }
 
-export const Hotflip = ({ final, original, new_prediction, context }: HotflipAttackOutput) => {
-    const [originalString, flippedString] = highlightFlippedTokens(
-        original,
-        final[0] // just the first one, huh,... yuck
-    );
+interface Props {
+    final?: string[];
+    original: string[];
```
codeviking, line 48: I might name these to `originalTokens` and `finalTokens` (if that's what they are) for consistency.

ui/src/demos/reading-comprehension/Attacks.tsx
```diff
@@ -62,7 +63,7 @@ export const Attacks = ({ model, input, target }: Props) => {
                 ) : null}
                 {supportedAttackTypes.has(AttackType.HotFlip) ? (
                     <Collapse.Panel key={AttackType.HotFlip} header="HotFlip">
-                        <Attack<Input, HotflipAttackOutput>
+                        <Attack<Input, HotflipAttackOutput<typeof prediction>>
```
codeviking, line 66: Shouldn't this just be `Prediction`? There's no way the compiler can infer anything more specific than that, right?

jonborchardt: i was surprised it let me do it.
if its cleaner ill convert

codeviking (approved):
LGTM.

ui/src/demos/reading-comprehension/Attacks.tsx
codeviking: Yea, I'd just use `Prediction`, as that's what `typeof prediction` will always return...I think?

PR #708, opened 2021-01-08
fix spans

jonborchardt:
the span used by bidaf are into the token list, not the char list, since we dont render tokens, we are no longer using this value

fixes: #684


codeviking (approved):
Sure, is there a reason to return `best_span` too then? It seems confusing to return both, especially if `best_span` is wrong?

jonborchardt:
> Sure, is there a reason to return `best_span` too then? It seems confusing to return both, especially if `best_span` is wrong?

the FE might use it someday... the backend passes lots of values we dont use
the value is not wrong.... its just not useful to the ui

PR #714, opened 2021-01-13
Add an endpoint for returning model cards.

codeviking:
This adds a single, shared endpoint for returning model cards.

We need this because some models (like NMN) use an older version
of `allennlp` or `allennlp_models` (or both) that doesn't have the
`ModelCard` stuff.

So instead we have a single endpoint that uses the latest (1.3)
n' greatest and serves model cards for all pretrained models.
The endpoint will be: `/api/model-card/:pretrained-model-id`.


jonborchardt (approved):
beautiful!

PR #715, opened 2021-01-13
Source the model card from the new endpoint.

codeviking:
This modifies the UI to use the new `/api/model-card` endpoint
for retrieving model cards.

There's a hack in place that maps `nmn` to an id of the
appropriate form. Soon we'll be migrating to ids that all
follow the `{task}-{model}` convention, at which point this
hack can be removed.


jonborchardt (approved):
seems reasonable

PR #719, opened 2021-01-14
make attacks and interperters generic (needed for next task, ner)

jonborchardt:


ui/src/components/Attackers.tsx
```diff
@@ -1,30 +1,33 @@
...
+    target: keyof I & string;
 }
 
-export const Attacks = ({ model, input, target, prediction }: Props) => {
+export const Attackers = <I, O>({
+    model,
+    input,
+    target,
+    prediction,
+    getBasicAnswer,
```
codeviking, line 29: This feels like it'd work great as `children`:

```
<Attackers
   model={model}
   ...>{({ output }) => (
    getBasicAnswer(output)
</Attackers>
```

This means we can do things like:

```
<Attackers
   model={model}>{({ output }) => (
        <h1>{output.answer}</h1>
</Attacker>
```

Functionally the same, but in my mind a little easier to use (and read).

codeviking (approved):
LGTM; One thought.

ui/src/components/Attackers.tsx
jonborchardt: updated

PR #720, opened 2021-01-14
fix interperaters to be generic (needed for NER)

jonborchardt:


ui/src/components/Interpreters.tsx
```diff
@@ -9,22 +9,21 @@ import { Model } from '../tugboat/lib';
...
-import { InterpreterId, WithTokenizedInput } from '../lib';
+import { InterpreterId } from '../lib';
 
-export interface InterpreterData {
-    instance_1: {
-        grad_input_1: number[];
-        grad_input_2: number[];
-    };
+export interface SaliencyData<O> {
+    interpretData: (output: O) => number[];
```
codeviking, line 15: This feels like an awkward place for this.  The name also makes me suspect, as it's *very* generic. I'd propose lifting this out and not giving these layers any knowledge of it.  More on that below.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -56,10 +56,23 @@ export const Main = () => {
...
+                                                        interpretData: (o: InterpreterData) =>
+                                                            o.instance_1.grad_input_2,
+                                                    },
+                                                    {
+                                                        header: 'Passage',
+                                                        tokens: output.passage_tokens,
+                                                        interpretData: (o: InterpreterData) =>
+                                                            o.instance_1.grad_input_1,
+                                                    },
+                                                ]}
```
codeviking, line 75: I'd propose we do this via the `<Interpreter />` children as well. I think it does a better job of maintaining the right encapsulation layers, as it prevents the `<Interpreters />` from having to have any knowledge of the specific data structure being handled:

```
<Interpreters>{({ output }) => (
   <SaliencyThing data={output.instance_1.grad_input_1} />
)}{</Interpreters>
```

This way the user can use any type of output they want, and the mapping function isn't "leaked" by being tunneled through the `<Interpreter />`. bit.

Can you tell I'm bullish on the applicability of this pattern? 😅

codeviking (commented):
I think we could make the mechanism for mapping the returned data to output a little simpler (and more flexible while doing so). LMK what you think.

ui/src/components/Interpreters.tsx
jonborchardt: done

ui/src/demos/reading-comprehension/Main.tsx
jonborchardt: done

PR #721, opened 2021-01-15
add ner and generalize some stuff

jonborchardt:
This adds the entire task for NER
(and some needed generalizations)

image

image

image


ui/src/demos/named-entity-recognition/Predictions.tsx
```diff
@@ -0,0 +1,201 @@
...
+        );
+    } else {
+        // If no entity,
+        // Display raw text.
+        return <span>{token.text}</span>;
+    }
+};
+
+// Lookup table for entity style values
+const entityLookup: {
```
jonborchardt, line 113: if this is used by other tasks, ill lift it up when we get there

ui/src/demos/named-entity-recognition/Usage.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+        <ModelUsage
+            installCommand={installCommand}
+            bashCommand={bashCommand}
+            pythonCommand={pythonCommand}
+            evaluationNote={evaluationNote}
+            trainingNote={trainingNote}
+        />
+    );
+};
```
codeviking, line 70: I'd expect more of this to be present in the model card.  It seems a little silly to jump through all sorts of hoops just to use the `archive_file` attribute and a single example.  

What if we just put this content in a `usage.md` file in each demo that we load and render as HTML? There's a markdown loader for webpack that'd allow us to do this at build-time.

WDYT?

ui/src/demos/named-entity-recognition/types.ts
```diff
@@ -0,0 +1,24 @@
+export interface Input {
+    sentence: string;
+}
+
+export interface WithTokenizedInput {
+    words: string[];
+}
+
+export const isWithTokenizedInput = (x: any): x is WithTokenizedInput => {
+    const xx = x as WithTokenizedInput;
```
codeviking, line 10: Didn't you switch `x` to `pred` in another PR? Should we stick with that?

ui/src/tugboat/lib/Example.ts
```diff
@@ -14,3 +14,15 @@ export function isGroupedExamples(
...
+    constructor() {
+        super(`The examples aren't grouped but are expected to be.`);
+    }
+}
+
+export class GroupedExamplesError extends Error {
+    constructor() {
+        super(`The examples are grouped but are expected not to be.`);
+    }
+}
```
codeviking, line 28: Can we move these to `tugboat/error/index.ts` instead?

codeviking (approved):
Nice! 

I left an idea about handling the usage information a little differently, but it's not a must-do. We might discuss it with the AllenNLP team before proceeding (it'll be harder to change this if we cement this pattern by applying it to all of the demos moving forward).

Otherwise there's one small cleanup item related to errors that are getting moved. Thanks for doing this!

ui/src/demos/named-entity-recognition/Usage.tsx
jonborchardt: i do want to move more to the model card, however, i dont want this to be just md...
as then the example and archive_file is no longer sourced from a single source
further, ner is special, its usage is 'easier' 
check out the reading-comp usage for a better example

im going to NOT do anything here, unless you object

ui/src/demos/named-entity-recognition/types.ts
jonborchardt: sure... its not a prediction though

ui/src/demos/named-entity-recognition/Usage.tsx
codeviking: > is no longer sourced from a single source

Yea, I get that. But I think we already have that problem.

Here's what I see happening as is:

1. NER is retrained.
2. AllenNLP updates `pretrained_model_file`.
3. The notes here don't get updated to reflect the new datasets that are used in #1.

So we already have the split brain problem in some sense. I think we should chat about this with the AllenNLP team before proceeding. Maybe we can make a case for why the usage information should be a part of the model cards (so it's easier for them to update)!

I think it's ok to merge, let's just make sure to follow up with them.

ui/src/demos/named-entity-recognition/types.ts
jonborchardt: done

ui/src/tugboat/lib/Example.ts
jonborchardt: done

PR #722, opened 2021-01-17
Finish up support for shareable URLs (permalinks).

codeviking:
This is Part 2 of 2 (hopefully) of the permalinks effort. It adds
logic that loads and then proceeds to submit the input associated
with a shared URL.

To get this working I had to do a fairly major rearchitecture of
things. Here's the most significant bits that changed:

**Selecting a new model now updates the URL.**

When the user selects a new model using the dropdown, it now adds
the model identifier to the URL path. For instance, selecting BiDAF
takes the user to `/reading-comprehension/bidaf`.

I did this for two reasons:

1. It means we have to persist less data, which means there's less
to restore when loading a shareable URL. Trying to set both the
form input and the selected input made things tricky, since the
bits that manage them are at different levels in the application's
hierarchy. It's definitely possible, it just gets messy fast.

2. It's nice for the end user. The shareable URLs now include the
model identifier, so there's more context that'll help them remember
what the link points to.

**I ripped out the complicated bits of the `usePromise()` API.**

I did this because I discovered a series of bugs related to the
mechanics of the `usePromise()` mechanism. It turns out all the work
I did to try and eliminate race-conditions ultimately made the
API a little unruly *and* made room for some nasty bugs to creep
in.

So I ripped out the complicated bits, but still retained the
`usePromise()` hook and `<Promised />` component, as I think they
still have value. They do a lot less now, but they still save a lot
of boilerplate and seem a lot easier to use now. Plus those bugs
that were around now appear to be gone. Or I might just be tired.

Also, for the record, @aimichal called this when he first saw
these APIs. "Beware ye codeviking who enter here..." couldn't
stop me from treading down the perilous path of premature
generalization.

**A note about the browser history:**

As implemented the history doesn't totally work, but I think it's
ok as-is.

The shareable URLs work great, and can be navigated to via the
browser's history.

If a user loads a shared URL and then submits something else, I
added code that pushes a new history entry into the stack without
the shareable URL parameters. If the user navigates back to this
URL nothing happens. If there's form input or output that's
being displayed it'll continue to be.

I tried really hard to find a way to fix this, but every route I
went down ended up with a big ole' mess of spaghetti. That's why
there's a datestamp at the end of this branch name, as I started
to have a hard time tracking attempts.

Anyway I think it's the right compromise. Overall this feels pretty
nice, and it's worth noting that navigating using the browser
history on the existing demo doesn't work at all -- so it's definitely
an improvement.

**It turns out this whole routing on the client thing is a lot of work...!**


ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -28,13 +23,29 @@ import {
...
+    output: Prediction;
+}
 
-        <DebugInfo input={input} output={output} model={model} />
-    </Output.Section>
-);
+export const Predictions = ({ input, model, output }: Props) => {
+    return (
+        <Output.Section>
+            <Output.SubSection title="Question">{input.question}</Output.SubSection>
```
codeviking, line 41: I forgot to mention this in my giant wall of text aka PR description.  I added the question to the output. This helped me keep track of when the contents of the form don't match the output being shown. I figured it'll probably help our users do that too.

ui/src/AppId.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * An identifier for the application that's used for persisting data associated with shareable
+ * URLs.
+ *
+ * See the `<Share.Link />` and `<Share.Controller />` components for more information.
+ */
+export const AppId = 'allennlp-demo';
```
jonborchardt, line 7: can we use the skiff app id instead?

ui/src/components/TaskCards.tsx
```diff
@@ -9,7 +9,7 @@ interface Props {
 }
 
 export const TaskCards = ({ children }: Props) => (
-    <Promised fetch={fetchTaskCards}>
-        {({ output: tasksById }) => <>{children(tasksById)}</>}
+    <Promised promise={fetchTaskCards} deps={[]}>
+        {(t) => <>{children(t)}</>}
```
jonborchardt, line 13: better var name?

ui/src/demos/reading-comprehension/Predictions.tsx
jonborchardt: i would cut this. many of the answer types display the question....

ui/src/demos/reading-comprehension/Predictions.tsx
```diff
@@ -49,25 +60,25 @@ const OutputByModel = ({
         case ModelId.Bidaf:
         case ModelId.BidafElmo:
         case ModelId.TransformerQa: {
-            if (isBiDAFPrediction(output) || isTransformerQAPrediction(output)) {
-                return <BasicPrediction input={input} output={output} />;
+            if (!isBiDAFPrediction(output) && !isTransformerQAPrediction(output)) {
```
jonborchardt, line 63: i dont prefer the new pattern... but its ok

ui/src/tugboat/components/share/slug.ts
```diff
@@ -0,0 +1,24 @@
+import slug from 'slug';
+
+/**
+ * Removes non-word characters and shortens the provided string to include up to `max` words.
+ */
+function shorten(s: string, max: number): string {
+    const words = s.replace(/[^\w\s]/g, '').split(/\s+/);
```
jonborchardt, line 7: we use slug.js elsewhere

jonborchardt (approved):
i am assuming it works ;-P
left some thoughts

ui/src/AppId.ts
codeviking: We could, but we'd have to wire it through as an environment variable. Hard-coding it here seems preferable to me, as it's simpler.

ui/src/components/TaskCards.tsx
codeviking: It's `t` for `task`, I can use the longer form. I'm just inspired by Go's use of short local variable names of late :-P.

ui/src/tugboat/components/share/slug.ts
codeviking: Ah, I think you meant this comment in relation to the TODO? I'll grep for how it's used and remove the TODO if the usages seem like things that we won't end up replacing.

ui/src/demos/reading-comprehension/Predictions.tsx
codeviking: I made the change so that the errors are more explicit. 

In this scenario the model is expected, but the output doesn't match our expectations for that model. So throwing a `UnexpectedModelError()` felt wrong. This change throws a `UnexpectedOutputError()` instead.

I did this because the other error was being thrown after a change I made, and it was confusing to debug. I ended up here and figured I'd save a future debugger the time it took me to unwind.

LMK what you think.

ui/src/tugboat/components/share/slug.ts
codeviking: The only other usage is in `tugboat/lib/Demos.ts`. So it's safe for us to remove it from the demo's dependencies once we publish `@allenai/tugboat`.

ui/src/components/TaskCards.tsx
codeviking: Updated!

ui/src/demos/reading-comprehension/Predictions.tsx
jonborchardt: makes sense

PR #739, opened 2021-01-18
general fixes in order to make adding oie easier

jonborchardt:


api/allennlp_demo/open_information_extraction/model.json
```diff
@@ -1,4 +1,6 @@
 {
     "id": "open-information-extraction",
-    "pretrained_model_id": "structured-prediction-srl"
+    "pretrained_model_id": "structured-prediction-srl",
```
codeviking, line 3: Wow, that's quite the id shift!

ui/src/components/Attack.tsx
```diff
@@ -64,3 +65,7 @@ export const Attack = <I, O>({
         </>
     );
 };
+
+const SpacedButton = styled(Button)`
+    margin-right: ${({ theme }) => theme.spacing.md};
+`;
```
codeviking, line 71: Should we just use `<antd.Space />`? I always forget about the component, but it means we don't have to write any css?

api/allennlp_demo/open_information_extraction/model.json
jonborchardt: ?

ui/src/components/Interpret.tsx
```diff
@@ -47,3 +48,7 @@ export const Interpret = <I extends { [k: string]: any }, O>({
         </>
     );
 };
+
+const SpacedButton = styled(Button)`
+    margin-right: ${({ theme }) => theme.spacing.md};
+`;
```
codeviking, line 54: Seems worth throwing in a shared location somewhere.

ui/src/components/Attack.tsx
jonborchardt: the issue is that the spinner and the answer content swap for each other....
not sure is space would help

ui/src/demos/named-entity-recognition/Main.tsx
```diff
@@ -41,18 +40,12 @@ export const Main = () => {
                             }>
                             {({ input, model, output }) => (
                                 <Output>
-                                    <Output.Section
+                                    <Output.ShareSection
```
codeviking, line 43:
```suggestion
                                    <Output.ShareableSection
```

Though I'm not a big fan of this generalization. This is a place where I think it's cleaner to keep them separate.

The rationale is that their combination feels a little awkward. Sure, they're often used together, but combining them feels like we're forcing two things into a bucket together that aren't terribly related.  Maybe it's ok and preferable because it saves a few lines, but my gut instinct doesn't like it.

ui/src/demos/named-entity-recognition/config.ts
```diff
@@ -1,10 +1,11 @@
 import { DemoConfig } from '../../tugboat/lib';
+import { ModelId } from '../../lib';
 
 export const config: DemoConfig = {
     group: 'Annotate a sentence',
     title: 'Named Entity Recognition',
     order: 1,
-    modelIds: ['named-entity-recognition', 'fine-grained-ner'],
+    modelIds: [ModelId.ElmoNer, ModelId.FineGrainedNer],
```
codeviking, line 8: Shouldn't `Ner` be `NER`, since it's an abbreviation?

I guess `Elmo` is too, which makes this confusing. This is a place where I'd just use strings instead of an enum, as I think it avoids questions like these :).

codeviking (approved):

ui/src/demos/named-entity-recognition/Main.tsx
jonborchardt: lets try it, this will be in every page, and users of main really dont care about how share works

codeviking: Sure, maybe we update the name though?

PR #741, opened 2021-01-19
Don't refetch things unnecessarily in the UI.

### LINKED ISSUE #734: Endpoint cleanup.
As the API has grown it's gotten a little unwieldy. In particular, the fact that the root request issued to each endpoint returns info about it has forced some odd trade offs. For instance, instead of requesting all tasks by hitting `https://demo.allennlp.org/api/tasks`, we have to hit `https://demo.allennlp.org/api/tasks/all`.

This of course isn't an earth-shattering level of technical debt, but also is surprising. I'd propose we move endpoint info to be served by `/info`, so that the root can be used for whatever is appropriate for the thing in question.

codeviking:
This modifies the UI so that we fetch all of the metadata we need
up front, rather then fetching it per demo. This makes it really
fast to switch from one demo to another, as there's no additional
requests to make. The trade off of course is that there's more going
across the wire in a single request, but the largest of these payloads
is 52 KB, so that should be fine. We ask the user to download a
bunch of JavaScript that's considerably larger than this.

Another benefit of this route is that the requests can all be made
in parallel. So there's an additional speed improvement via that
mechanism too.

As part of this I cleaned a few things up about the the backend too.
The `/api/tasks/all` endpoint is now just `/api/tasks/`, and
`/api/model-card/` became `/api/model-cards/`. This makes the
OCD side of my brain happy, and fixes #734. I also modified
the `/api/info` endpoint to only include information about
the endpoints that do things with models, as the other endpoints
don't really belong there.


api/allennlp_demo/tasks/api.py
```diff
@@ -18,21 +16,10 @@ def __init__(self, name: str = "tasks"):
...
         def tasks():
             tasks = get_tasks()
             return flask.jsonify(tasks)
 
-        @self.route("/task/<string:task_id>", methods=["GET"])
-        def task(task_id: str):
-            for (tid, task) in get_tasks().items():
-                if tid == task_id:
-                    return flask.jsonify(task)
-            raise NotFound(f"No task card with {task_id} found.")
```
codeviking, line 34: We don't use this endpoint so it's safe to remove.

dev/prod_models.conf
```diff
@@ -18,6 +18,18 @@ server {
...
+        rewrite /api/model-cards(/(.*))? /$2 break;
+        proxy_pass http://model_cards:8000;
+        proxy_set_header X-Forwarded-For $remote_addr;
+    }
+
+    location /api/tasks {
+        rewrite /api/tasks(/(.*))? /$2 break;
+        proxy_pass http://tasks:8000;
+        proxy_set_header X-Forwarded-For $remote_addr;
+    }
```
codeviking, line 31: I modified the UI only route so that these "metadata" services run locally too. It makes it easier to test and tweak those bits n' pieces. This might make it easier, for instance, to validate upcoming changes to the model card API.

ui/src/App.js
```diff
@@ -54,28 +56,40 @@ for which the route `/task/<model_name>` serves the <SingleTaskDemo> component,
...
+            <ScrollToTopOnPageChange />
+            <DemoWrapper>
+                <ErrorBoundary>
+                    <Promised 
+                        promise={() => Promise.all([ 
+                            fetchModelInfo(), 
+                            fetchTaskCards(), 
+                            fetchModelCards() 
+                        ])} 
+                        deps={[]}>
```
codeviking, line 70: The new `<Promised />` API makes a lot more sense to me.

ui/src/App.js
```diff
@@ -54,28 +56,40 @@ for which the route `/task/<model_name>` serves the <SingleTaskDemo> component,
...
-                    />
-                ))}
-                <Route path="/:model/:slug?" component={Demo} />
-            </Switch>
-        </VarnishApp>
-    </Router>
+    <VarnishApp layout="left-aligned">
+        <Router>
+            <ScrollToTopOnPageChange />
+            <DemoWrapper>
```
codeviking, line 62: I lifted this out so that the `<Loading />` indicator is displayed in the same spot as it is now. It turns out it there was no reason to nest it as deeply as it was. This should also make things a little snappier, as React will have less to reconcile as we change from one demo to another.

ui/src/components/MultiModelDemo.tsx
```diff
@@ -54,35 +44,33 @@ interface Props {
...
+
+    const cardsById = useContext(ModelCards);
+    const models: Model[] = [];
+    for (const info of demoInfos) {
+        const modelCardId = getModelCardId(info);
+        if (!(modelCardId in cardsById)) {
+            throw new ModelCardNotFoundError(info);
+        }
+        models.push(new Model(info.id, cardsById[modelCardId]));
+    }
```
codeviking, line 66: FYI @jonborchardt This will impact the `<SingleModelDemo />` factoring you're wiring up.

ui/src/demos/named-entity-recognition/Main.tsx
```diff
@@ -23,63 +22,61 @@ import { Version, Input, Prediction, InterpreterData, isWithTokenizedInput } fro
...
+                    </Predict>
+                </Tabs.TabPane>
+                <Tabs.TabPane tab="Model Card" key="Card">
+                    <ModelCard />
+                </Tabs.TabPane>
+                <Tabs.TabPane tab="Model Usage" key="Usage">
+                    <Usage />
+                </Tabs.TabPane>
+            </Tabs>
+        </MultiModelDemo>
```
codeviking, line 80: The only change here is I removed the wrapping `<Content />`, since we don't need it anymore.

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -32,71 +31,69 @@ import {
...
+                    </Predict>
+                </Tabs.TabPane>
+                <Tabs.TabPane tab="Model Card" key="Card">
+                    <ModelCard />
+                </Tabs.TabPane>
+                <Tabs.TabPane tab="Model Usage" key="Usage">
+                    <Usage />
+                </Tabs.TabPane>
+            </Tabs>
+        </MultiModelDemo>
```
codeviking, line 97: Again, just removed the wrapping `<Content />`.

ui/src/components/MultiModelDemo.tsx
jonborchardt: im not making a single model... turns out its not needed

ui/src/css/model.css
```diff
@@ -3,7 +3,7 @@
 }
 
 .model__content {
-  padding: 0.6em 2em 0.875em 2em;
+  padding: 0 0 0.875em 0;
```
jonborchardt, line 6: this will, of course, eventually go away

ui/src/demos/named-entity-recognition/Main.tsx
jonborchardt: this is fine
and the time for it... but as i add more and more demos, changes like these will realllllly identify what could be factored out...

jonborchardt (approved):
lg2m

ui/src/css/model.css
codeviking: Yes! I look forward to it :).

ui/src/demos/named-entity-recognition/Main.tsx
codeviking: Yup. That's how the best abstractions are arrived at.  It's hard to make good guesses beforehand, but as they say, hindsight is 20/20!

ui/src/components/MultiModelDemo.tsx
codeviking: Great!

PR #742, opened 2021-01-19
Add integration tests.

codeviking:
Breaking things this morning reminded me (yet again) of how
complex this application is. To help avoid this, this PR
adds a mechanism for writing and running integration tests.

The tests run the UI against the current production APIs for
doing things with models (predict, interpret and attack requests),
so it's admittedly not a true integration test in that the local
APIs aren't being verified.

That said this would've caught the issue I introduced this morning,
so I think it's probably ok that this isn't perfect. Also, this
gets things rolling so we can make it better over time -- baby steps!


jonborchardt (approved):
seems fine

PR #743, opened 2021-01-19
Jon oie

jonborchardt:
this adds openie

to do this, i needed to make some more stuff general... please check independent commits for clarity

note, we are not using hierplane for this anymore:
image


README.md
```diff
@@ -34,10 +34,28 @@ There's three ways to run things locally:
     docker-compose -f docker-compose.ui-only.yaml up --build
     ```
 
+    or use the script:
+
+    ```
+    ./bin/dockerui
```
codeviking, line 40: I might just call this `start_ui` or `start_ui_only` or something like that. The name `dockerui` makes me think it's a UI for...docker?

In other words, I don't care if the script uses docker or not, I care that it starts a ui -- so let's try to capture that in the name.

README.md
```diff
@@ -34,10 +34,28 @@ There's three ways to run things locally:
...
+    ```
+
    Once that's complete you'll be able to access your local version by opening
    [http://localhost:8080](http://localhost:8080) in a browser. Changes to the code should
    be automatically applied.
 
+   Note: To clean up docker containers, be sure to use:
+
+   ```
+   docker-compose -f docker-compose.ui-only.yaml up --build
```
codeviking, line 50:
```suggestion
   docker-compose -f docker-compose.ui-only.yaml down --build
```

ui/bin/dockerui.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
codeviking, line 2:
```suggestion
set -eu
```

`set -e` -> If a command fails, stop, instead of trying to continue.
`set -u` -> Fail if there's a reference to a variable that isn't defined.

ui/bin/dockerui.sh
```diff
@@ -0,0 +1,8 @@
...
+if [ $1 == "down" ]
+then
+    docker-compose -f ../docker-compose.ui-only.yaml down
+else
+    docker-compose -f ../docker-compose.ui-only.yaml up --build
+fi
```
codeviking, line 8: I think this script will fail if you're not in the repository root. Consider making these scripts path independent like so:

```
dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
docker-compose -f "$dir/../docker-compose.ui-only.yaml down"
```

ui/src/demos/open-information-extraction/Main.tsx
```diff
@@ -1,17 +1,59 @@
...
-            NOTE: This task is under development
-        </Content>
+            <TaskDescription />
+            <Tabs>
+                <Tabs.TabPane tab="Demo" key="Demo">
+                    <SelectExample displayProp="sentence" placeholder="Select a Sentence" />
+                    <Predict<Input, Prediction>
+                        fields={
+                            <>
+                                <Sentence />
```
codeviking, line 32: I wonder if we should package these up in an `Input` namespace / object, so that it's clear they're input controls, not display ones.

```
<Input.Sentence />
<Input.Text />
```

ui/src/demos/open-information-extraction/Main.tsx
```diff
@@ -1,17 +1,59 @@
...
+                                <Submit>Run Model</Submit>
+                            </>
+                        }>
+                        {({ input, model, output }) => (
+                            <Output>
+                                <Output.ShareableSection
```
codeviking, line 38: I'm still struggling with this name.

I wonder if we could do `<Output.WithShareSupport />` or `<Output.WithShareButton />` instead?  I'm just struggling to understand the intent.  Another optiona:

```
<Output  extra={<ShareButton /> />
```

☝️ I like this route, we get to avoid the top-level component by capturing the `<AlignRight />` and `<Share.Link />` in a single component.

WDYT?

ui/src/demos/open-information-extraction/Predictions.tsx
```diff
@@ -0,0 +1,123 @@
...
+                })}
+            </Output.SubSection>
+
+            <DebugInfo input={input} output={output} model={model} />
+        </Output.Section>
+    );
+};
+
+const TokenSpan = ({ token, id }: { token: FormattedToken; id: number }) => {
+    if (token.entity !== undefined) {
```
codeviking, line 46: Consider reversing this. I think this is. case where go's preference for "flat" functions makes a lot of sense:

```
if (token.entity === undefined) {
    return <span>{`${token.text}${' '}`}</span>;
} 

/* Rest of the function body */.
```

It makes it really easy to read and understand the edge-case. As is you have to jump all the way to the bottom of the method to grok how it's handled.

ui/src/demos/open-information-extraction/Predictions.tsx
```diff
@@ -0,0 +1,123 @@
...
+        const tagParts = token.entity.split('-');
+        const [tagLabel, attr] = tagParts;
```
codeviking, line 48: Is it safe to assume there will always be a `-` in the entity? Should we handle the case cleanly where `tagParts.length < 2`?

codeviking (approved):
Great. I left a few ideas.

ui/src/demos/open-information-extraction/Predictions.tsx
jonborchardt: this works as is... testing for existence of attr below

README.md
jonborchardt: done

README.md
jonborchardt: fixed

ui/bin/dockerui.sh
jonborchardt: added

ui/bin/dockerui.sh
jonborchardt: added

ui/src/demos/open-information-extraction/Main.tsx
jonborchardt: done

ui/src/demos/open-information-extraction/Main.tsx
jonborchardt: added shrebutton

ui/src/demos/open-information-extraction/Predictions.tsx
jonborchardt: done

PR #753, opened 2021-01-20
add srl and make things more general

jonborchardt:
image


aimichal (approved):
I had a quick look. Not sure what to pay attention to, but looks okay to me :-)

schmmd:
@jonborchardt looks mostly good, but I would show the shorthand that's in the text output and show the long form on mouseover.  It'll be easier for an expert to scan, and it will avoid the cases where the annotation is longer than the text.

E.g. "in the car" would have "ARGM-LOC" but on mouseover would show "MODIFIER-LOCATION"

For SRL, I would change "Extractions" to "Frames" because that's the word they use, as I recall.

PR #754, opened 2021-01-21
Add Sentry for capturing client-side exceptions.

codeviking:
This adds client-side error capturing powered by
[Sentry.io](https://sentry.io).

I added a query string argument that we can use temporarily
to throw an error to make sure things are working. I'll
remove it after we've verified everything.

I also wired things up so that errors should include the
associated environment and SHA, which should make it easier
to debug things.

Lastly I enabled source maps so that the line-numbers
included in the stack traces that Sentry captures are
useful. I can't test this locally, so I'll need to create
an ad-hoc to do so. There's no risk of doing this since
the source code is already public (and arguably there isn't
even if that's not the case, as it's fairly well known
how to reverse minification and obfuscation these days).

Pass through relevant environment variables at build time.

### DIFF
--- a/dev/check_models_ci.py
+++ b/dev/check_models_ci.py
@@ -7,7 +7,7 @@ import os
 import logging
 from typing import Iterable
 
-WORKFLOW_FILE_PATH = ".github/workflows/ci.yml"
+WORKFLOW_FILE_PATH = ".github/workflows/api_ci.yml"
 logging.basicConfig()
 
 # These are endpoints we have tests for that aren't models. This script skips these.
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -5,6 +5,12 @@ WORKDIR /ui
 COPY package.json yarn.lock ./
 RUN yarn
 
+ARG env
+ARG sha
+
+ENV SENTRY_ENVIRONMENT=${env}
+ENV SENTRY_RELEASE=${sha}
+
 COPY . .
 RUN yarn build
 
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -30,8 +30,6 @@ import '@allenai/varnish/dist/theme.css';
 Sentry.init({
     dsn: "https://59686a41b9664bf2a8bbc51a602428c2@o226626.ingest.sentry.io/5599301",
     autoSessionTracking: true,
-    // This value should be tweaked for production, and should be a value between 0 and 1.
-    tracesSampleRate: 1.0,
     environment: process.env.SENTRY_ENVIRONMENT || 'dev',
     release: process.env.SENTRY_RELEASE || 'none'
 });
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -28,10 +28,10 @@ import '@allenai/varnish/dist/theme.css';
 // Sentry is a tool that captures JavaScript errors at runtime and aggregates them.
 // If you need access, ask someone on the AllenNLP team.
 Sentry.init({
-    dsn: "https://59686a41b9664bf2a8bbc51a602428c2@o226626.ingest.sentry.io/5599301",
+    dsn: 'https://59686a41b9664bf2a8bbc51a602428c2@o226626.ingest.sentry.io/5599301',
     autoSessionTracking: true,
     environment: process.env.SENTRY_ENVIRONMENT || 'dev',
-    release: process.env.SENTRY_RELEASE || 'none'
+    release: process.env.SENTRY_RELEASE || 'none',
 });
 
 /*******************************************************************************
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
+import * as Sentry from '@sentry/react';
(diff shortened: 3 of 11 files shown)


dev/check_models_ci.py
```diff
@@ -7,7 +7,7 @@
 import logging
 from typing import Iterable
 
-WORKFLOW_FILE_PATH = ".github/workflows/ci.yml"
+WORKFLOW_FILE_PATH = ".github/workflows/api_ci.yml"
```
codeviking, line 10: Oops. I broke this.

I noticed `master` was failing. Sorry about that!

ui/src/App.js
```diff
@@ -24,6 +25,17 @@ import './css/hierplane-overrides.css';
 import './css/visualization-types.css';
 import '@allenai/varnish/dist/theme.css';
 
+// Sentry is a tool that captures JavaScript errors at runtime and aggregates them.
+// If you need access, ask someone on the AllenNLP team.
+Sentry.init({
+    dsn: "https://59686a41b9664bf2a8bbc51a602428c2@o226626.ingest.sentry.io/5599301",
```
codeviking, line 31: The portion after the `@` is something [Sentry calls the "Secret Key"](https://docs.sentry.io/product/sentry-basics/dsn-explainer/).

But they also say this:

> The secret part of the DSN is optional and effectively deprecated. While clients will still honor it, if supplied, future versions of Sentry will entirely ignore it.

...and the value has to be embedded in the client anyway. So sure, we could jump through some hoops to make this accessible at runtime instead of baking into the client -- but it'd still be in the JavaScript payload in plain-text, so I don't see any values in those acrobatics.

LMK if y'all think we should be more careful about this value, in which case we'll figure out what feels right (and I'll rotate the value).

ui/src/demos/reading-comprehension/Main.tsx
```diff
@@ -29,6 +29,10 @@ import {
 } from './types';
 
 export const Main = () => {
+    // TODO (@codeviking): Remove this once we've verified Sentry is working.
+    if (document.location.search.indexOf('?throw=true') !== -1) {
+        throw Error('⚡️ This is a test error to verify that Sentry is wired up correctly.');
+    }
```
codeviking, line 35: I'll yank this out after I make sure everything is working on the live site.

ui/src/App.js
aimichal: Looks like `59686a41b9664bf2a8bbc51a602428c2` is the PUBLIC_KEY in the template

> {PROTOCOL}://{PUBLIC_KEY}:{SECRET_KEY}@{HOST}{PATH}/{PROJECT_ID}

No : before the @ in that URL, so no secret key. 👍

ui/src/App.js
```diff
@@ -24,6 +25,15 @@ import './css/hierplane-overrides.css';
 import './css/visualization-types.css';
 import '@allenai/varnish/dist/theme.css';
 
+// Sentry is a tool that captures JavaScript errors at runtime and aggregates them.
+// If you need access, ask someone on the AllenNLP team.
+Sentry.init({
+    dsn: 'https://59686a41b9664bf2a8bbc51a602428c2@o226626.ingest.sentry.io/5599301',
+    autoSessionTracking: true,
+    environment: process.env.SENTRY_ENVIRONMENT || 'dev',
+    release: process.env.SENTRY_RELEASE || 'none',
```
aimichal, line 34: Hmm, is App.js run in the browser? If so, then process.env.* won't be set, so then we'll always have  'dev' and 'none', right?

Or is this file executed during build, where node is invoked and does have those vars you specified in the Dockerfile?

ui/Dockerfile
```diff
@@ -5,6 +5,12 @@ WORKDIR /ui
 COPY package.json yarn.lock ./
 RUN yarn
 
+ARG env
+ARG sha
+
+ENV SENTRY_ENVIRONMENT=${env}
+ENV SENTRY_RELEASE=${sha}
```
aimichal, line 12: These names make it look like this has something to do with the environment and release of sentry to use.

But these mean the environment and release *of our app* which we *report* to sentry.

If that's right, maybe vars like this would be a little clearer, at the expense of slightly longer names:

ALLENNLP_DEMO_ENV_FOR_SENTRY
ALLENNLP_DEMO_RELEASE_FOR_SENTRY

or something.

Just a thought!

aimichal (approved):
I think this all looks good, but I do have some questions. I could be completely missing something obvious tho!

jonborchardt (approved):

ui/src/App.js
codeviking: These are injected at build-time via the `webpack.EnvironmentPlugin()`.  I wired them into the build process.

ui/Dockerfile
codeviking: Yea, that's a valid point. That said these are the canonical values used by `Sentry`, and some clients (their SDKs for server-side exception tracking) automatically read these by name.

So I'm inclined to stick with the convention, as we do with things like `AWS_*` and `GOOGLE_*` and so on.

ui/src/App.js
codeviking: Thanks, my tired brain wasn't reading things correctly. 👍

PR #755, opened 2021-01-21
Fix Sentry in production.

codeviking:
I'm not sure why, but right now errors handled by the
`<ErrorBoundary />` aren't being sent to Sentry. This works
in development, but not in production, which leads me to believe
it has something to do with the fact that I decided to *not*
use `<Sentry.ErrorBoundary />`.

This PR factors things out a bit so that we can use the display
logic we already have in place, but use it within the context
of Sentry's `<ErrorBoundary />` I'll test this out in an ad-hoc
to verify that errors get captured.

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -6,7 +6,7 @@ import { Content, Footer, Header, Layout, VarnishApp } from '@allenai/varnish/co
 import { ScrollToTopOnPageChange } from '@allenai/varnish-react-router';
 
 import { Demos } from './tugboat/lib';
-import { ErrorBoundary, Promised } from './tugboat/components';
+import { ErrorBoundaryView, Promised } from './tugboat/components';
 
 import allenNlpLogo from './components/allennlp_logo.svg';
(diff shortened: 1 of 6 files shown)


ui/src/App.js
```diff
@@ -88,7 +91,15 @@ const App = () => (
...
                                             {demos.all().map(({ config, Component }) => (
                                                 <Route key={config.path} path={config.path}>
-                                                    <Component />
+                                                    <Sentry.ErrorBoundary
+                                                        fallback={({ error, resetError }) => (
+                                                            <ErrorBoundaryView
+                                                                error={error}
+                                                                resetError={resetError}
+                                                            />
+                                                        )}>
```
codeviking, line 100: Adding a second error boundary means this error will get cleared if the user uses the left-hand navigation.

This seems preferable, as otherwise there's things the user can click that don't do anything -- which isn't a great UX.

jonborchardt (approved):

PR #756, opened 2021-01-21
Handle legacy permalinks cleanly.

codeviking:
Right now if a user loads a legacy permalink, they get an error
without any info about why the URL doesn't work. It's unclear
if it will ever work, and what they should do.

This adds code that instead handles this situation by cleanly.
We show a warning to the user and redirect them to the current
demo they're trying to use in a clean state.

As part of this I removed the `resetError()` mechanism, as it
doesn't really work in practice (if they reload the same URL
over and over, they might get the same error repeatedly).

Here's how it looks:

Screen Shot 2021-01-21 at 12 24 18 PM


jonborchardt (approved):
beautiful!

PR #757, opened 2021-01-21
add dep and con pars tasks

jonborchardt:
this uses hierplane for now


ui/src/@types/hierplane.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * hierplane does not have types
+ */
+declare module 'hierplane' {
+    const Tree: (spec: any) => any;
```
codeviking, line 5: We should be able to at least type this as a `React.PureComponent` or something like that.  For the props I'd make `tree` a `{}`.

It's better than the `any` route, and shouldn't be too much work.

ui/src/components/Hierplane.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    .hierplane__visualization-verbs__prev {
+        transform: rotateY(-180deg);
+    }
+
+    .hierplane__visualization-verbs__label {
+        user-select: none;
+    }
+
+    /* THESE ARE HACKS PUT HERE BY CODEVIKING. */
```
codeviking, line 45:
```suggestion
    /* THESE ARE HACKS PUT HERE BY @codeviking */
```

ui/src/demos/constituency-parsing/Predictions.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const Predictions = ({ input, model, output }: Props) => {
+    // TODO: this is a placeholder visualization. We will be removing Hierplane someday soon!
+    return (
+        <Output.Section>
+            {output.hierplane_tree ? (
+                <Output.SubSection>
+                    <Hierplane tree={output.hierplane_tree} theme="light" />
+                </Output.SubSection>
+            ) : null}
```
codeviking, line 22: Should we at least print the output when the tree isn't returned, maybe a `<code>{JSON.stringify(output, null, 2)}</code>`.  Maybe that's already taken care of by the debug output, in which case we could just let the user know to look there: `Unable to render output. See debug output below`.

codeviking (approved):
Nice, moving right along!

ui/src/@types/hierplane.d.ts
jonborchardt: done

ui/src/demos/constituency-parsing/Predictions.tsx
jonborchardt: added

PR #758, opened 2021-01-21
activate tasks and remove ones we wont do

jonborchardt:
just publishing the next 3 tasks


schmmd (approved):
Approved, assuming you are just removing boilerplate for things you had expected to migrate but don't need to.

jonborchardt:
> Approved, assuming you are just removing boilerplate for things you had expected to migrate but don't need to.

yes

PR #759, opened 2021-01-21
add enums and api config settings for remaining tasks

jonborchardt:


api/allennlp_demo/coref/model.json
```diff
@@ -1,5 +1,6 @@
 {
     "id": "coreference-resolution",
     "pretrained_model_id": "coref-spanbert",
-    "attackers": ["input_reduction"]
+    "attackers": [],
```
jonborchardt, line 4: current ui does not show an attack... odd this was here

codeviking (approved):

PR #761, opened 2021-01-22
Jon coref

jonborchardt:
image

### DIFF
--- a/ui/src/components/DebugInfo.tsx
+++ b/ui/src/components/DebugInfo.tsx
@@ -10,7 +10,7 @@ export interface Props {
 }
 
 export const DebugInfo = ({ input, output, model }: Props) => {
-    const title = 'API Output';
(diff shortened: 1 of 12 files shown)


ui/src/components/DebugInfo.tsx
```diff
@@ -10,7 +10,7 @@ export interface Props {
 }
 
 export const DebugInfo = ({ input, output, model }: Props) => {
-    const title = 'Debug Output';
+    const title = 'CLI Output';
```
codeviking, line 13: Why are we updating this to be called CLI Output? Are we sure this actually matches the I/O of the allennlp CLI?

My gut tells me it's different, which might cause end users to be confused. I'd stick with Debug.

codeviking (commented):
LGTM; Though I think renaming "Debug Output" to "CLI Output" will probably cause some user confusion.

PR #769, opened 2021-01-25
Jon menu and fields

jonborchardt:
1- updates menu to have a better structure
2- make inputs to be more generic


ui/src/demos/constituency-parsing/Main.tsx
```diff
@@ -29,7 +29,7 @@ export const Main = () => {
                     <Predict<Input, Prediction>
                         fields={
                             <>
-                                <Field.Sentence />
+                                <Field.Input property="sentence" />
```
codeviking, line 32: How about we use `name` instead of `property`? I feel like `property` is a bit overloaded, given the use of React `props`.

ui/src/demos/coreference-resolution/Main.tsx
```diff
@@ -29,7 +29,7 @@ export const Main = () => {
                     <Predict<Input, Prediction>
                         fields={
                             <>
-                                <Field.Document />
+                                <Field.TextArea property="document" />
```
codeviking, line 32: I think `Document` should probably just be `Passage`? Does the backend call this a `document`?

I prefer the existing names. We're not trying to abstract a `TextArea` or an `Input`, `antd` already does that for us. We're trying to abstract inputs that are typical in the types of demos we build.  These are things like `Sentence`s, `Question`s and `Passage`s.  This is important, as we might decide that a `Sentence` is better as a single-line `Input`, and then decide later it's a `TextArea` and this shouldn't matter to demo developers.  They just know they want whatever we consider the canonical display for a prompt for a `Sentence`.

So I'd vote we redact this change, and in this case consider `Passage` over `Document`, as the two seem redundant.

ui/src/demos/masked-language-modeling/config.ts
```diff
@@ -1,7 +1,7 @@
 import { DemoConfig } from '../../tugboat/lib';
 
 export const config: DemoConfig = {
-    group: 'Other',
+    group: 'Generate a passage',
```
codeviking, line 4: I think it generates a sentence, right?

ui/src/groups.ts
```diff
@@ -5,8 +5,9 @@ import { DemoGroup } from './tugboat/lib';
  * as part of the DOM. They'll be faster, look better and we can change colors n' such via
  * CSS.
  */
+import addIcon from './icons/add-14px.svg';
 import annotateIcon from './icons/annotate-14px.svg';
-import otherIcon from './icons/other-14px.svg';
+import parseIcon from './icons/parse-14px.svg';
 import passageIcon from './icons/passage-14px.svg';
 import questionIcon from './icons/question-14px.svg';
```
codeviking, line 12: We should inline this (since SVGs are valid React). Do we at least have a `TODO` for doing this?

codeviking (changes requested):
I'm not a fan of the new names -- let me know what you think of my critique.

ui/src/groups.ts
jonborchardt: there is already a todo

ui/src/demos/masked-language-modeling/config.ts
jonborchardt: this label is from ms

ui/src/demos/constituency-parsing/Main.tsx
jonborchardt: reverted

ui/src/demos/coreference-resolution/Main.tsx
jonborchardt: reverted

ui/src/demos/masked-language-modeling/config.ts
codeviking: /shrug. Ok!

codeviking (approved):

PR #770, opened 2021-01-26
te task

jonborchardt:
BTW: i think we should remove this example...

image


ui/src/demos/textual-entailment/Predictions.tsx
```diff
@@ -0,0 +1,258 @@
...
+                It is <strong>somewhat likely</strong> that {judgment}.
+            </div>
+        );
+    } else {
+        summaryText = <div>The model is not confident in its judgment.</div>;
+    }
+
+    const formatProb = (n: number) => {
+        return parseFloat((n * 100).toFixed(1)) + '%';
+    };
```
codeviking, line 123: Consider using `Intl` instead:

```
let f = new Intl.NumberFormat(navigator.language, { style: 'percent', maximumFractionDigits: 1 });
console.log(f.format(0.1231)); // outputs "12.3%"
```

schmmd:
@jonborchardt why do you think we should remove that example?

ui/src/demos/textual-entailment/Predictions.tsx
```diff
@@ -0,0 +1,258 @@
...
+};
+
+const Columns = styled.div`
+    display: grid;
+    grid-template-columns: auto 1fr;
+    gap: ${({ theme }) => theme.spacing.xl};
+
+    @media ${({ theme }) => belowOrEqualTo(theme.breakpoints.lg)} {
+        grid-template-columns: 1fr;
+        gap: ${({ theme }) => theme.spacing.md};
```
codeviking, line 181: You should test this in Safari. I learned the other day that Safari doesn't support `gap`, so we might have to use `grid-gap` instead?

ui/src/demos/textual-entailment/Predictions.tsx
```diff
@@ -0,0 +1,258 @@
...
+ */
+interface TernaryGraphProps {
+    contradiction: number;
+    neutral: number;
+    entailment: number;
+}
+
+const TernaryGraph = ({ contradiction, neutral, entailment }: TernaryGraphProps) => {
+    const containerWidth = 290;
+    const containerHeight = 260;
```
codeviking, line 196: These don't actually reflect the dimensions of the container though, right? Maybe these should just be named `fixedWidth` and `fixedHeight`, respectively?

ui/src/demos/textual-entailment/Predictions.tsx
```diff
@@ -0,0 +1,258 @@
...
+    height: ${({ height }) => `${height}px`};
+    left: ${({ left }) => `${left}px`};
+    top: ${({ top }) => `${top}px`};
+`;
+
+const TernaryPlot = styled.img.attrs({ src: TePlotSrc })<{ top: number; left: number }>`
+    position: absolute;
+    margin: -7px 0 0 -7px;
+    width: 14px;
+    height: 14px;
```
codeviking, line 255: This looks like a centering trick. Could we use `display: grid` and `align-items` or `justify-content` instead?

codeviking (approved):
LGTM; Left a few ideas.

jonborchardt:
> @jonborchardt why do you think we should remove that example?

because it has the word 'God'

ui/src/demos/textual-entailment/Predictions.tsx
jonborchardt: navigator.language 
NICE!

jonborchardt: done

ui/src/demos/textual-entailment/Predictions.tsx
jonborchardt: done

ui/src/demos/textual-entailment/Predictions.tsx
jonborchardt: renamed

ui/src/demos/textual-entailment/Predictions.tsx
jonborchardt: no because this is absolutely positioned

codeviking: Ah, but is that necessary?

PR #772, opened 2021-01-26
adding viz for masked lm

jonborchardt:
this also cleans up the inputs in tugboat
note: there is a todo that needs to happen before i merge… but this is ready to review

image


ui/src/demos/masked-language-modeling/Main.tsx
```diff
@@ -1,17 +1,86 @@
...
+                <Tabs.TabPane tab="Demo" key="Demo">
+                    <SelectExample displayProp="text" placeholder="Select a Text" />
+                    <Predict<Input, Prediction>
+                        fields={
+                            <>
+                                <Field.Text validator={Validator} />
+                                <Field.Hidden
+                                    name="sentence"
+                                    value="The doctor ran to the emergency room to see [MASK] patient."
+                                />
```
codeviking, line 37: Why is the sentence hidden? Isn't this something the user provides?

ui/src/demos/masked-language-modeling/Main.tsx
```diff
@@ -1,17 +1,86 @@
...
+            </Tabs>
+        </MultiModelDemo>
     );
 };
+
+const Validator = async (_: RuleObject, text: string) => {
+    if (!text || text.indexOf('[MASK]') === -1) {
+        return Promise.reject(new Error('The sentence must include one or more "[MASK]" tokens.'));
+    }
+};
```
codeviking, line 86: What if we gave this a more specific name, like `validateHasAtLeastOnceMaskToken` or something like that?

Also, does `[MASK]` need to be a separate word, or can it be used like this:

```
some[MASK]
```

Right now I think this would be allowed, but it's possible that's not valid.

ui/src/demos/masked-language-modeling/Main.tsx
jonborchardt: this is a hack because im blocked by the backend, it will be resolved before i merge
text will become sentence

ui/src/demos/masked-language-modeling/Main.tsx
jonborchardt: ill rename...
and it is valid to have mask on a word, the api treats it as a new word.

jonborchardt: done

codeviking (approved):

codeviking (approved):
We need to update `api/Dockerfile` too, so that things are consistent. Once that's done, LGTM.

PR #774, opened 2021-01-28
vqa task visualization

jonborchardt:
lots of odd looking changes in here
it is likely easier to look at the individual commits
@codeviking pleases take a look at the changes to tugboat (examples, prediction, etc)

Note: i will not be merging this until after the allennlp release at the end of the day because we need changes to from the api.

happy to go over this in person if its easier


ui/src/demos/visual-question-answering/Main.tsx
```diff
@@ -1,17 +1,107 @@
...
+import { Predictions } from './Predictions';
+import { Input, Prediction } from './types';
+
+import baseballSrc from '../exampleImages/baseball_game.jpg';
+import busStopSrc from '../exampleImages/bus_stop.jpg';
+import kitchenSrc from '../exampleImages/kitchen.jpg';
+import livingRoomSrc from '../exampleImages/living_room.jpg';
 
 export const Main = () => {
+    // Fields on the for to force an update to the value, this is needed because the input control
```
codeviking, line 32:
```suggestion
    // Fields on the form to force an update to the value, this is needed because the input control
```

ui/src/demos/visual-question-answering/Main.tsx
```diff
@@ -1,17 +1,107 @@
...
+    // does not know about the form, and the form is not available at field construction time.
+    const [fieldChanges, setFieldChanges] = useState<FormFieldDict>();
```
codeviking, line 34: I'd consider calling these `overrides` (or at least, that's what comes to mind now).

ui/src/demos/visual-question-answering/Main.tsx
```diff
@@ -1,17 +1,107 @@
...
+                            <>
+                                <Field.Image
+                                    value={selectedExample?.image}
+                                    onChange={(v) => setFieldChanges(v)}
+                                />
+                                <Field.Question />
+                                <Submit>Run Model</Submit>
+                            </>
+                        }
+                        fieldChanges={fieldChanges}>
```
codeviking, line 88: Yea, I think this would be easier to follow if it were:

```
const [ overrides, setOverrides ]
```

...and then here it was:

```
overrides={overrides}
```

ui/src/demos/visual-question-answering/Main.tsx
```diff
@@ -1,17 +1,107 @@
...
-        </Content>
+            <TaskDescription />
+            <Tabs>
+                <Tabs.TabPane tab="Demo" key="Demo">
+                    <SelectExample
+                        displayProp="snippet"
+                        placeholder="Select an Example"
+                        overrides={examples}
+                        onChange={(ex?: Example) => setSelectedExample(ex)}
+                    />
```
codeviking, line 76: Rather than passing `examples` and overriding `setSelectedExamples`, let's just expose an `examples` property on `MultiModalDemo` and pass it through.

Then modify the `examples` context to only fetch them when they're not provided.  This way that mechanism stays intact, and other demos can override the list of examples if they'd like too (which seems like something that'd be generally useful).

ui/src/@types/assets.d.ts
```diff
@@ -9,3 +9,7 @@ declare module '*.svg' {
     const content: any;
     export default content;
 }
+declare module '*.jpg' {
+    const content: any;
+    export default content;
+}
```
codeviking, line 15: I wish TypeScript would just include this by-default. Or maybe as a lib that's packaged with TypeScript but can be added.

ui/src/demos/visual-question-answering/Main.tsx
```diff
@@ -1,17 +1,107 @@
-/*
-    Note: this file is in development and should only be viewed as example code direction
-*/
+/**
+ * NOTE: VQA is a complex task and is a bad example to fork from unless you need an Image input
+ * control.
+ */
```
codeviking, line 4: It's great that you thought about this up-front. Nicely done!

ui/src/demos/visual-question-answering/Predictions.tsx
```diff
@@ -0,0 +1,84 @@
...
+            <DebugInfo input={input} output={output} model={model} />
+        </Output.Section>
+    );
+};
+
+const AnswerTable = (p: any) => Table<Answer>(p);
+
+const StyledTable = styled(AnswerTable)`
+    table {
+        min-width: 98% !important; // boo! there seems to be an error in the width of the inner table
```
codeviking, line 82: Does 98vw work too? I like `vw` since it avoids some of the issues `%` has (as `%` is relative to it's offset container, I think, whereas `vw` is always about the viewport).

That said it also might *not* work.

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
+import React, { useState, useEffect } from 'react';
+import styled from 'styled-components';
+import { Upload, message } from 'antd';
+import { UploadChangeParam } from 'antd/lib/upload';
+import { UploadFile } from 'antd/lib/upload/interface';
+import { LoadingOutlined, UploadOutlined } from '@ant-design/icons';
+
+export interface ModelParams {
```
codeviking, line 8:
```suggestion
export interface UploadedImage {
```

Or maybe `ImageData`, or something like that.  In my opinion there's nothing model specific about this, so a more general name for the type makes sense.

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+    imgSrc?: string;
+    imageName?: string;
```
codeviking, line 10:
```suggestion
    fileName?: string;
```

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
+import React, { useState, useEffect } from 'react';
+import styled from 'styled-components';
+import { Upload, message } from 'antd';
+import { UploadChangeParam } from 'antd/lib/upload';
+import { UploadFile } from 'antd/lib/upload/interface';
+import { LoadingOutlined, UploadOutlined } from '@ant-design/icons';
+
+export interface ModelParams {
+    imgSrc?: string;
```
codeviking, line 9:
```suggestion
    src?: string;
```

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+    imageName?: string;
+    image?: File;
+    image_base64?: string;
```
codeviking, line 12:
```suggestion
    b64EncodedData?: string;
```

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+                            image: undefined,
+                        });
+                        setImageLoadingAndSendEvent(false);
+                    }
+                );
+            }
+        } else if (info.file.status === 'error') {
+            // If the image is too large, then info.file.error.message will
+            // look like "cannot post api/permalink/noop 413'"
+            const errorIs413 = info.file.error.message.match("^cannot post .* 413'$");
```
codeviking, line 93: Could you check for a response with the 413 status code instead of matching against the error message?

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+            file.name = imgSrc;
+            s = {
+                imgSrc: imgSrc,
+                imageName: file.name,
+                image: file as File,
+            };
+        }
+        setStateAndSendEvent(s);
+        setImageLoadingAndSendEvent(false);
+    }
```
codeviking, line 137: I've been avoiding `async` in the codebase and using promises instead. I think we should do so here too.

It's unfortunate, as I really like `async / away`. But there's two pitfalls:

1. Exception handling becomes less explicit, and it's easier to leave uncaught exceptions in place. Promises make it harder to forget to do this, or rather make it easier to spot a missing `catch()`.

2. React's `useEffect()` hook doesn't support `async` functions. So we have to use promises in this API, or add anonymous `async` function that wraps the function body, which seem silly.

WDYT?

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+    const setStateAndSendEvent = (s: ModelParams) => {
```
codeviking, line 139:
```suggestion
    const setStateAndTriggerOnChange = (s: ModelParams) => {
```

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: backend needs it as is

ui/src/demos/visual-question-answering/Main.tsx
jonborchardt: yep, will do

ui/src/demos/visual-question-answering/Predictions.tsx
jonborchardt: ill try

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: doh, i meant to do this before making the pr!

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: k

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: k

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: i dont think so, this was an odd hack from the copied ocde

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+                        <DraggerImg src={localState.imgSrc} />
+                    ) : null}
+                    {!imageLoading && !localState.imgSrc ? (
+                        <DraggerMessage>
+                            <UploadOutlined /> Upload an Image
+                        </DraggerMessage>
+                    ) : null}
+                </Dragger>
+            </div>
+        </React.Fragment>
```
codeviking, line 174: The `noop` thing is a silly hack, and has confused me several times now. I'm also pretty sure @aimichal and I spent some time trying to figure out why it was there.

The native drag APIs aren't too hard to use. I've written code before that uses the events.  Consider ripping out `Dragger` and just binding to the `onDrag*` events yourself.  The code will be cleaner, and we can drop this whole `noop` thing that's bound to confuse a few others along the way.

Your future self (and the AllenNLP team) will appreciate it :).

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: i agree... but i dont want to creep the task, ill add a ticket to do this work

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: ill consider it, and AT LEAST leave a comment about it

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+    const reader = new FileReader();
+    reader.onload = (event) => {
+        const img = new Image();
+        img.src = event.target ? (event.target.result as string) : '';
+        img.onload = () => {
+            const canvas = document.createElement('canvas');
+            const curPixels = img.width * img.height;
+            const scale = Math.sqrt(maxPixels) / Math.sqrt(curPixels);
+            canvas.width = img.width * scale;
+            canvas.height = img.height * scale;
```
codeviking, line 209: So, if `scale` is `1` we still do work? 

Maybe we just skip the work when we can? This will save the user some batter life, and make things a tad faster?

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,283 @@
...
+        };
+        reader.onerror = (error) => {
+            console.log(error);
+            if (onError) {
+                onError('Error compressing image.');
+            }
+        };
+    };
+    reader.readAsDataURL(file);
+};
```
codeviking, line 242: This looks like it just scales the image's dimensions down as to make sure it's below a given size.

I'd consider calling this `downscaleImage` or something like that. I'd also put the

ui/src/tugboat/components/SelectExample.tsx
```diff
@@ -49,7 +58,11 @@ export const SelectExample = ({ displayProp, placeholder }: Props) => {
             <FieldItem label="Example Inputs">
                 <Select
                     value={ctx.selectedExample ? ctx.selectedExample[displayProp] : undefined}
-                    onChange={(id) => ctx.selectExample(examplesById[`${id}`])}
+                    onChange={(id) => {
+                        const selValue = examplesById[`${id}`];
+                        onChange && onChange(selValue);
+                        return ctx.selectExample(selValue);
+                    }}
```
codeviking, line 65: As noted, I don't think you need any of the changes in this file. I'd just wire some predefined examples through the `MultiModalDemo` bit, so that if the context is initialized with examples, it doesn't query new ones.

ui/src/tugboat/components/form/Form.tsx
```diff
@@ -64,6 +75,16 @@ const FormImpl = <I, O>({ action, children, fields, shared }: FormImplProps<I, O
...
 
+    // Update fields from outside callers.
+    useEffect(() => {
+        if (fieldChanges) {
+            // The antd API wants a `RecursivePartial<I>` here, which I can't figure out how to
+            // satisfy. What's interesting is the method only assigns values for properties whose
+            // name matches that of a form field -- so `Partial<I>` should be ok.
+            form.setFieldsValue({ ...form.getFieldsValue(), ...(fieldChanges as any) });
+        }
+    }, [fieldChanges]);
```
codeviking, line 86: Consider calling this `overrides`.

ui/src/tugboat/components/form/Image.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return (
+        <FieldItem label="Image" name="image" rules={[{ required: true }]}>
+            <ImageUpload
+                onChange={(img) => {
+                    blobToString(img.image)
+                        .then((str) => {
+                            img.image_base64 = str;
+                            props.onChange({
+                                image: img,
+                            });
```
codeviking, line 26: If you rename `img` to `image` you can do:

```
props.onChange({ image })
```

Not a big deal though.

ui/src/tugboat/components/form/Image.tsx
```diff
@@ -0,0 +1,34 @@
...
+                        })
+                        .catch((e) => console.log(e));
```
codeviking, line 28: We should display something to the user, and use Sentry's APIs for capturing the error and sending it to our backend.

Just grep for `Sentry.` to find other examples of where we do this.

ui/src/tugboat/components/form/Form.tsx
```diff
@@ -17,11 +17,16 @@ export type FormOutputView<I, O> = (io: {
     output: O;
 }) => React.ReactNode | JSX.Element;
 
+export interface FormFieldDict {
+    [fieldId: string]: any;
+}
```
codeviking, line 22:
```suggestion
export interface Fields {
    [name: string]: any
}
```

ui/src/tugboat/components/form/Hidden.tsx
```diff
@@ -8,12 +8,18 @@ import { FieldItem, TextArea } from './controls';
...
-        <TextArea />
-    </FieldItem>
-);
+export const Hidden = ({ name, value }: Props) => {
+    return (
+        <FieldItem
+            hidden
+            name={name}
+            initialValue={JSON.stringify(value)}
+            dependencies={name ? [name] : undefined}>
```
codeviking, line 21: What are dependencies?

codeviking (approved):
Nice work. This one was tough!

I think there's an opportunity to make a few things simpler.

ui/src/demos/visual-question-answering/Main.tsx
jonborchardt: done

ui/src/demos/visual-question-answering/Main.tsx
jonborchardt: done

ui/src/demos/visual-question-answering/Main.tsx
jonborchardt: done

ui/src/demos/visual-question-answering/Predictions.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: this is only called when scale will be < 1

jonborchardt: and as is... you COULD use it to scale up

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: its called compressImage.... is that not reasonable?

ui/src/tugboat/components/SelectExample.tsx
jonborchardt: done, check new code

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/form/Form.tsx
jonborchardt: done

ui/src/tugboat/components/form/Image.tsx
jonborchardt: done

ui/src/tugboat/components/form/Image.tsx
jonborchardt: done

ui/src/tugboat/components/form/Hidden.tsx
jonborchardt: oops, removing

jonborchardt: i tried soooo many things before pulling the ripcord and using onchange events

ui/src/tugboat/components/form/Form.tsx
jonborchardt: done

ui/src/tugboat/components/form/Hidden.tsx
jonborchardt: done

ui/src/tugboat/components/SelectExample.tsx
jonborchardt: note, i still need this change because it is how the image input is 'told' of example changes

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+                );
+            } else {
+                if (onError) {
+                    onError('Invalid context.');
+                }
+            }
+        };
+        reader.onerror = (error) => {
+            Sentry.captureException(error);
+            console.log(error);
```
codeviking, line 240: You can probably remove this.  If you want to keep it, I'd use `console.error()` instead of `console.error` instead: 

```suggestion
            console.error(error);
```

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+    maxFileBytes,
+    newFileName,
+    onSuccess,
+    onError,
+}: {
+    file: File;
+    maxFileBytes: number;
+    newFileName?: string;
+    onSuccess: Function;
+    onError: Function;
```
codeviking, line 195: What is this new `Function` type I see? I'm so used to us doing `() => ...`?

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+                if (onError) {
+                    onError('Invalid context.');
+                }
+            }
+        };
+        reader.onerror = (error) => {
+            Sentry.captureException(error);
+            console.log(error);
+            if (onError) {
+                onError('Error compressing image.');
```
codeviking, line 242: Consider passing the source error too, so that calling code can do something with it:

```suggestion
                onError('Error compressing image.'. error);
```

I also might consider making a specific error, and passing `error` as the cause:

```
class ImageCompressionError {
   constructor(readonly cause: Error) {
         super(`Error compressing image: ${cause}`);
   }
}

throw new ImageCompressionError(error);
```

ui/src/tugboat/components/form/Image.tsx
```diff
@@ -0,0 +1,38 @@
...
+                    blobToString(image.image)
+                        .then((str) => {
+                            image.image_base64 = str;
+                            props.onChange({
+                                image,
+                            });
+                        })
+                        .catch((e) => {
+                            Sentry.captureException(e);
+                            console.log(e);
```
codeviking, line 31:
```suggestion
                            console.error(e);
```

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+                            imageSrc: URL.createObjectURL(compressedFile),
+                            fileName: compressedFile.name,
+                            image: compressedFile,
+                        });
+                        setImageLoadingAndSendEvent(false);
+                    },
+                    () => {
+                        message.error(
+                            `${info.file.name} file upload failed: ${info.file.error.message}`
+                        );
```
codeviking, line 84: I've been using `antd`'s `notification` API for errors elsewhere. Is `message` preferable? 

Maybe use `notification` too, so it's consistent? (I do this for errors creating permalinks).

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+    file: File;
+    maxFileBytes: number;
+    newFileName?: string;
+    onSuccess: Function;
+    onError: Function;
+}) => {
+    const path = require('path');
+    const originalFileName = path.basename(file.name);
+    const fileName = newFileName || `${originalFileName}_c.jpg`;
+    // console.log(`Compressing ${file.name} (${file.size}Bytes)`);
```
codeviking, line 200: Stray debug statement?

ui/src/tugboat/components/ImageUpload.tsx
```diff
@@ -0,0 +1,288 @@
...
+            if (ctx) {
+                ctx.drawImage(img, 0, 0, canvas.width, canvas.height);
+                ctx.canvas.toBlob(
+                    (blob: Blob | null) => {
+                        if (blob) {
+                            const compressedFile = new File([blob], fileName, {
+                                type: 'image/jpeg',
+                                lastModified: Date.now(),
+                            });
+                            // console.log(`Generated ${fileName} (${compressedFile.size}Bytes)`);
```
codeviking, line 224: Another?

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: 2 vs 3
https://www.albertgao.xyz/2016/08/11/how-to-declare-a-function-type-variable-in-typescript/

ui/src/tugboat/components/form/Image.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/ImageUpload.tsx
jonborchardt: done

ui/src/tugboat/components/form/Image.tsx
jonborchardt: done

PR #798, opened 2021-02-05
Rename `<MultiModelDemo />` to `<TaskDemo />`.

codeviking:
This PR does two things:

1. Patch `<MultiModelDemo />` so that when there is only one model,
   we don't add the selected model ID to the URL path. This prevents
   us from having URLs like this:
   https://demo.allennlp.org/open-information-extraction/open-information-extraction

2. Rename `<MultiModelDemo />` to `<TaskDemo />`, since it handles
   a single model well now too.

### DIFF
--- a/ui/src/components/Demo.tsx
+++ b/ui/src/components/Demo.tsx
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 
 import { Model, Task, Example } from '../tugboat/lib';
-import { MultiModelDemo as TBMultiModelDemo } from '../tugboat/components';
+import { Demo as TBDemo } from '../tugboat/components';
 
 import { AppId } from '../AppId';
 import { TaskCard, getModelCardId, ModelInfo } from '../lib';
@@ -41,12 +41,12 @@ interface Props {
 }
 
 /**
- * An AllenNLP demo featuring multiple models, one of which can be selected at a time.
+ * An AllenNLP demo for a particular task.
  *
  * This component exists primarily to handle the process of converting AllenNLP's specific notion
- * of a model (which is queried via API routes) to the shape expected by the tugboat package.
+ * of a model (which is queried via API routes) to the shape expected by the tugboat module.
  */
-export const MultiModelDemo = ({ ids, taskId, children, examples }: Props) => {
+export const Demo = ({ ids, taskId, children, examples }: Props) => {
     const infos = useContext(ModelInfoList);
     const included = new Set(ids);
     const demoInfos = infos.filter((info) => included.has(info.id));
@@ -69,13 +69,9 @@ export const MultiModelDemo = ({ ids, taskId, children, examples }: Props) => {
 
     return (
         <RedirectLegacyPermalinks modelIds={ids}>
-            <TBMultiModelDemo
-                models={models}
-                task={asTugBoatTask(task)}
-                appId={AppId}
-                examples={examples}>
+            <TBDemo models={models} task={asTugBoatTask(task)} appId={AppId} examples={examples}>
                 {children}
-            </TBMultiModelDemo>
+            </TBDemo>
         </RedirectLegacyPermalinks>
     );
 };
--- a/ui/src/tugboat/components/Demo.tsx
+++ b/ui/src/tugboat/components/Demo.tsx
@@ -6,12 +6,12 @@ import { Config, Models, CurrentTask, Examples } from '../context';
 import { Example, Model, Task } from '../lib';
 import { ModelNotFoundError, NoModelsError } from '../error';
 
-interface SelectedModelRouteProps {
+interface ModelRouteProps {
     models: Model[];
     children: React.ReactNode;
 }
 
-const SelectedModelRoute = ({ models, children }: SelectedModelRouteProps) => {
+const SelectedModelRoute = ({ models, children }: ModelRouteProps) => {
     if (models.length === 0) {
         throw new NoModelsError();
     }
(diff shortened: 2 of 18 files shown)


ui/src/components/Demo.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 
 import { Model, Task, Example } from '../tugboat/lib';
-import { MultiModelDemo as TBMultiModelDemo } from '../tugboat/components';
+import { Demo as TBDemo } from '../tugboat/components';
```
jonborchardt, line 4: I propose we dont rename this file from MultiModelDemo to Demo... its overloaded
maybe TaskDemo or TaskInfo or TaskPage?

ui/src/tugboat/components/Demo.tsx
```diff
@@ -6,12 +6,12 @@ import { Config, Models, CurrentTask, Examples } from '../context';
 import { Example, Model, Task } from '../lib';
 import { ModelNotFoundError, NoModelsError } from '../error';
 
-interface SelectedModelRouteProps {
+interface ModelRouteProps {
```
jonborchardt, line 9: this is really TaskRouteProps.... we have model and task mixed up i am seeing....

ui/src/tugboat/components/Demo.tsx
```diff
@@ -6,12 +6,12 @@ import { Config, Models, CurrentTask, Examples } from '../context';
...
     models: Model[];
     children: React.ReactNode;
 }
 
-const SelectedModelRoute = ({ models, children }: SelectedModelRouteProps) => {
+const SelectedModelRoute = ({ models, children }: ModelRouteProps) => {
```
jonborchardt, line 14: SingleModelTaskRoute

ui/src/tugboat/components/Demo.tsx
```diff
@@ -35,6 +35,24 @@ const SelectedModelRoute = ({ models, children }: SelectedModelRouteProps) => {
     );
 };
 
+const SingleModel = ({ models, children }: ModelRouteProps) => {
```
jonborchardt, line 38: SingleModelTask

ui/src/tugboat/components/Demo.tsx
```diff
@@ -35,6 +35,24 @@ const SelectedModelRoute = ({ models, children }: SelectedModelRouteProps) => {
...
+    if (models.length === 0) {
+        throw new NoModelsError();
+    }
```
jonborchardt, line 41: `else if(models.length > 1) { throw}`

ui/src/tugboat/components/Demo.tsx
```diff
@@ -35,6 +35,24 @@ const SelectedModelRoute = ({ models, children }: SelectedModelRouteProps) => {
...
+
+    // There can be only one...Model.
+    const selectedModel = models[0];
```
jonborchardt, line 44: could probably convert to a typeguard that throws... just a silly idea

ui/src/tugboat/components/Demo.tsx
```diff
@@ -59,12 +77,16 @@ export const MultiModelDemo = ({ models, task, children, appId, examples }: Prop
             <CurrentTask.Provider value={{ task }}>
                 <Examples.Provider
                     value={{ examples: examples || task.examples, selectedExample, selectExample }}>
-                    <Switch>
-                        <Route path={modelPath}>
-                            <SelectedModelRoute models={models}>{children}</SelectedModelRoute>
-                        </Route>
-                        <Redirect to={firstModelPath} />
-                    </Switch>
+                    {models.length > 1 ? (
```
jonborchardt, line 80: this surprises me...
i would expect to just router to X, and have X decide which component to use
i propose removing this logic from the router

jonborchardt (commented):
left some thoughts
i think we should address many of them before merging

ui/src/components/Demo.tsx
codeviking: Sure, let's do `<TaskDemo />`.

ui/src/tugboat/components/Demo.tsx
codeviking: Yup, will update.

ui/src/tugboat/components/Demo.tsx
codeviking: Good call.

ui/src/tugboat/components/Demo.tsx
codeviking: Sorry, I don't follow.

The router sets up an explicit path for each model when there's more than one. We end up with routes like `/:task/:modelId`.  If we do that when there's one model, we end up with URLs like this:

https://demo.allennlp.org/open-information-extraction/open-information-extraction

...which seems a little silly.

So I think it make sense to remove the router when we don't need it.

jonborchardt (approved):

ui/src/tugboat/components/Demo.tsx
jonborchardt: discussed offline, this routing code is great as is

ui/src/components/Demo.tsx
codeviking: Done.

ui/src/tugboat/components/Demo.tsx
codeviking: I think it's ok as is, it's a local symbol, and the props are shared across both local components (one of which is a route handler).

ui/src/tugboat/components/Demo.tsx
codeviking: Yea, I think the above idea (throwing) is fine (and nice and simple).

PR #815, opened 2021-02-08
Jon fixes2

jonborchardt:
fixes 
812
808
771


ui/src/components/DebugInfo.tsx
```diff
@@ -28,17 +29,27 @@ export const DebugInfo = ({ input, output, model }: Props) => {
...
                 </Collapse.Panel>
-            </Collapse>
+            </NoOverflowCollapse>
         </Output.Section>
     );
 };
+
+const NoOverflowCollapse = styled(Collapse)`
+    overflow: hidden;
+`;
```
codeviking, line 55: Why is `overflow: hidden` wanted here?  Personally, I prefer `overflow: scroll`, as that usually fixes the visual issue but still makes the content accessible.

I don't like it when websites hide content in the name of things look nice when what I'm there for is the content.

But I suspect this is correcting for something, and that there's likely a root-cause for that something that we should try to fix instead. Maybe it's a [grid blowout](https://css-tricks.com/preventing-a-grid-blowout/), for instance, which can be tricky to spot but are easily fixed.

Let me know what issue you're solving and then maybe I can help dig into what the root cause is.

ui/src/demos/language-modeling/Predictions.tsx
```diff
@@ -38,18 +38,25 @@ interface SingleResultProps {
 const SingleResult = ({ pred, input }: SingleResultProps) => {
     const form = useContext(Form);
     const completion = cleanTopTokensForDisplay(pred);
+    const displayCompletion =
+        completion.slice(-1) === '.' || completion.slice(-1) === '↵'
+            ? completion
+            : completion.concat(' ...');
```
codeviking, line 44:
```suggestion
            : completion.concat(' …');
```

This uses the utf-8 ellipsis instead of three periods. This is nicer for accessibility clients, search engines and is a single character.

ui/src/demos/language-modeling/Predictions.tsx
```diff
@@ -38,18 +38,25 @@ interface SingleResultProps {
 const SingleResult = ({ pred, input }: SingleResultProps) => {
     const form = useContext(Form);
     const completion = cleanTopTokensForDisplay(pred);
+    const displayCompletion =
+        completion.slice(-1) === '.' || completion.slice(-1) === '↵'
```
codeviking, line 42: Consider using `.endsWith(str)` instead. I think it's a bit truer to the intent.

codeviking (approved):
This is fine as-is, though I'd recommend digging into the reason you're adding `overflow: hidden` a bit. I worry it might be rendering content inaccessible that users might be interested in in the interest of visual fidelity.  I suspect there's something else wrong style-wise that we could fix instead, which might resolve other side-effects too.

ui/src/components/DebugInfo.tsx
jonborchardt: this makes the scroll inside the container work

jonborchardt: making auto works

ui/src/demos/language-modeling/Predictions.tsx
jonborchardt: done

PR #816, opened 2021-02-09
use external lib for tugboat

jonborchardt:
replaces all references to the internal tugboat code with references to the new external lib


ui/src/App.tsx
```diff
@@ -31,7 +30,8 @@ Sentry.init({
 
 const DEFAULT_PATH = '/reading-comprehension';
 
-const demos = Demos.load();
+const ctx = require.context('./demos', true, /\/index\.ts$/);
```
jonborchardt, line 33: this is new

ui/src/components/ModelUsage.tsx
```diff
@@ -73,10 +72,8 @@ export const ModelUsage = (props: Props) => {
                 <p>
                     About the dataset:{' '}
                     <MaybeLink
-                        props={{
-                            name: props.modelCard.evaluation_dataset.name,
-                            url: props.modelCard.evaluation_dataset.url,
-                        }}
+                        name={props.modelCard.evaluation_dataset.name}
```
jonborchardt, line 75: fix

ui/src/components/ModelUsage.tsx
```diff
@@ -97,10 +94,8 @@ export const ModelUsage = (props: Props) => {
...
```
jonborchardt, line 97: fix

ui/src/App.tsx
codeviking: Can we add a comment here, similar to what existed previously, that explains the expected structure?

IIRC I had some comments in the `Demos.load()` method that described the mechanism. Given this is a "magic" aspect of webpack that even we struggle to understand, I think we should leave a description here that helps other unwind what's going on.

ui/src/components/Attack.tsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
 import { Button } from 'antd';
-
-import { Promised } from '../tugboat/components';
-import { Models } from '../tugboat/context';
-import { NoSelectedModelError } from '../tugboat/error';
+import { Promised } from '@allenai/tugboat/components';
+import { Models } from '@allenai/tugboat/context';
+import { NoSelectedModelError } from '@allenai/tugboat/error';
```
codeviking, line 6: Any thought in rolling things into a top level barrel file, so this could be a single line?

```
import { Promised, Models, NoSelectedModelError } from '@allenai/tugboat'
```

WDYT?

ui/src/demos/constituency-parsing/types.ts
```diff
@@ -1,4 +1,4 @@
-import { emory } from '../../tugboat/lib';
+import { emory } from '@allenai/tugboat/lib';
```
codeviking, line 1: I wonder if we're "leaking" an abstraction here, and this should be renamed in `@allenai/tugboat` to be something more general, like `persistence` or `share`.

The motivation being that applications shouldn't really care or know if we use emory or not. Granted, I think emory has staying power, so this might not be an important distinction to make.

@aimichal WDYT?

ui/src/demos/coreference-resolution/Predictions.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
+import { Output, NestedHighlight, withHighlightClickHandling } from '@allenai/tugboat/components';
```
codeviking, line 2: Oof. The `withHighlightClickHandling()` thing is rough. I was hoping we'd do some cleanup of the whole `<Highlight />` mechanism before breaking things out.

The rationale being that if we end up using `@allenai/tugboat` elsewhere, then it makes it harder for us to fix this pattern -- because it'll end up encoded in multiple places.

I'd consider leaving the highlight stuff in the AllenNLP demo. This way there's some healthy friction that forces us to "fix" it before using it elsewhere.

ui/yarn.lock
```diff
@@ -7,6 +7,18 @@
...
+  version "0.1.2"
+  resolved "https://registry.yarnpkg.com/@allenai/tugboat/-/tugboat-0.1.2.tgz#7064eeb0e43c278c090d9cc6b41fb4738b8f037d"
+  integrity sha512-xieWAMCL0GrMFbP6XxfJeiLu64e6ZNWJ4t6YPyUFduUTuZyeGGu3Cf1FvfV4ElYZIVqfB+Yr7izPuUB9P4f8CQ==
+  dependencies:
+    colormap "^2.3.1"
+    commonmark "^0.29.2"
+    react-router "^5.0.1"
+    react-router-dom "^5.0.1"
+    react-syntax-highlighter "^15.3.0"
+    slug "^4.0.2"
```
codeviking, line 20: Does the AllenNLP code still need `slug`? Can we remove it from it's dependency manifest?

ui/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@allenai/eslint-config-varnish": "^1.0.1",
+    "@allenai/tugboat": "^0.1.2",
```
codeviking, line 7: Woohoo! 🛥️

codeviking (approved):
LGTM; I left some ideas.

ui/src/components/Attack.tsx
jonborchardt: making a ticket to consider... there are pros and cons of doing this

ui/src/demos/constituency-parsing/types.ts
jonborchardt: we can rename whenever... just let me know

ui/src/demos/coreference-resolution/Predictions.tsx
jonborchardt: agreed, ill make a ticket

ui/yarn.lock
jonborchardt: ill check, good catch

PR #828, opened 2021-02-12
Add script for building and managing local containers.

codeviking:
This adds a new script, `./demo`, that can be used to start (and
build) a local container, see a list of those that are runnign (
and those that aren't), and stop those that are running.

At this point things can be started and stopped, and all share
a network, but aren't yet routable from the host machine. I plan
to add that bit in another PR, as this is already a big chunk
of code.

### DIFF
--- a/demo
+++ b/demo
@@ -1,0 +1,400 @@
+#!/usr/bin/env python3
+
+"""
+This is a script intended to make running a local version of the AllenNLP Demo easier.
+
+The AllenNLP demo is a series of individual services that together make the experience
+available at https://demo.allennlp.org. Weaving these services together locally to make
+changes takes a surprising amount of acrobatics, so rather than expecting you to become
+an expert in docker, we wrap the necessary incanations in a series of simple commands
+provided by this script.
+
+The easiest way to learn more about this script and how it works is likely by viewing
+it's usage information, which you can see by running this command:
+
+    ./demo --help
+
+It's worth noting that this script isn't used anywhere other than local development
+environments, and probably shouldn't be. This means it's safe to make changes without
+worrying about breaking the house of cards that makes the live site work.
+"""
+
+import argparse
+import subprocess
+import logging
+import os
+import sys
+import shlex
+import time
+
+from enum import Enum
+from typing import List, Tuple, Optional
+from pathlib import Path
+
+prefix = "allennlp-demo"
+network = prefix
+
+def name(svc: str) -> str:
+    """
+    Returns a name to use when running a container for the specified service. This makes it
+    easy to determine if one is already running, so that we don't spin up the same endpoint
+    multiple times.
+    """
+    return f"{prefix}-{svc}"
+
+def tag(svc: str) -> str:
+    """
+    Returns the Docker tag to use for images built for the provided service.
+    """
+    # We use `:latest`, which should be avoided for deploying images for use in production,
+    # but is ok in this case.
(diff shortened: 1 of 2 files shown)


demo
```diff
@@ -0,0 +1,400 @@
...
+provided by this script.
+
+The easiest way to learn more about this script and how it works is likely by viewing
+it's usage information, which you can see by running this command:
+
+    ./demo --help
+
+It's worth noting that this script isn't used anywhere other than local development
+environments, and probably shouldn't be. This means it's safe to make changes without
+worrying about breaking the house of cards that makes the live site work.
```
aimichal, line 19: Should it be called `local-demo` or `demo-local`?

demo
```diff
@@ -0,0 +1,400 @@
...
+            start(svc)
+
+    elif args.command == "stop" or args.command == "down":
+        if len(args.service) == 0:
+            stop_all()
+        else:
+            for svc in args.service:
+                stop(svc)
+
+    elif args.command == "restart" or args.command == "hup":
```
aimichal, line 386: I think it will be simpler to start with a single string for each command; right now we can each choose "start" or "up, "stop" or "down", "restart" or "hup" (which i think means hang up, right?), etc. That can be confusing for people helping each other with it.

demo
```diff
@@ -0,0 +1,397 @@
...
+        for svc in args.service:
+            restart(svc)
+
+    elif args.command == "status" or args.command == "ps":
```
aimichal, line 390: Should there be a final else that shows usage?

demo
```diff
@@ -0,0 +1,397 @@
...
+    # We use `:latest`, which should be avoided for deploying images for use in production,
+    # but is ok in this case.
+    return f"{name(svc)}:latest"
+
+class Status(Enum):
+    """
+    An Enum reflecting the various states of a container.
+    """
+    RUNNING = 0
+    STOPPED = 1
```
aimichal, line 58: Does RUNNING really mean running, or started but crashed?

Sorry to be picky, it probably doesn't matter, but i didn't follow all the code too closely so it might.

aimichal (approved):
oh, and i mean 👍

demo
codeviking: That'd be more clear, but I like the brevity. Let's see what the AllenNLP folks think.

demo
codeviking: Yea, good idea.

I didn't know what to use for `restart` so I chose `hup` arbitrarily.  But your'e right -- I should drop the aliases for now. People can easily add them later if there's a command they execute a lot that's hard to remember or type.

demo
codeviking: `Argparse` handles that by default.

If you try to provide a command that's not in the list of known ones, it outputs an error:

```
❯ ./demo foo
usage: demo [-h] {start,up,stop,down,restart,hup,status,ps} ...
demo: error: argument command: invalid choice: 'foo' (choose from 'start', 'up', 'stop', 'down', 'restart', 'hup', 'status', 'ps')
```

demo
codeviking: Nah, it's a good question.

If it crashes the container process will be terminated, so `RUNNING` does in fact mean it's up and hasn't exited.

When I originally wrote this I figured there'd be more states -- but it ended up just being binary. The process is either up or not, and if it's up we use the health check mechanism to determine if it's responding to HTTP requests as we'd expect.

PR #830, opened 2021-02-12
‘fix’ safari issue

jonborchardt:


codeviking (changes requested):
I don't think we should hack around this in this way.

I'll TAL on Monday and figure out what's causing the prop update to not propagate properly.  We'll either yank out the underlying syntax highlighter and find a better one, or maybe even fix their bug. Who knows.

codeviking:
I more or less took the same route, but I just used the `key` attribute which I think is a little simpler.
https://github.com/allenai/allennlp-demo/pull/844

It's worth noting that functionally they're the same though, so you made the right call here.

PR #841, opened 2021-02-15
Add a local reverse proxy to the local demo script.

codeviking:
The primary, functional change in this PR adds a local reverse
proxy that handles forwarding requests to either local services
or the production site as appropriate.

But this change also includes a sizeable refactor of the script
(sorry). When I first took a stab at this I started by simply
adding another layer if branching to the existing functions,
but that got really messy and unsustainable quickly. So I decided
to take the plunge and use an object-oriented API instead. I think
the result is a lot easier to follow, and should make it a bit
easier to add additional containers as the demo evolves.

It's also probably worth noting that the script is getting a bit
long, but I like keeping things self-contained and think it's still
worth the tradeoff. Let me know what you all think.

### DIFF
--- a/demo
+++ b/demo
@@ -189,8 +189,7 @@ class Container:
         # If `p.poll()` returns a value, the process has terminated. We intentionally don't timeout,
         # we leave that to the user and Ctrl^C.
         while p.poll() is None and not self.is_healthy():
-            delay = 5
-            time.sleep(delay)
+            time.sleep(5)
 
         # If the service still isn't healthy, then it failed to start.
         if not self.is_healthy():
--- a/demo
+++ b/demo
@@ -528,11 +528,13 @@ class Proxy(Container):
         """
         if self.status() != Status.UP:
             raise RuntimeError(f"The proxy isn't up.")
+        print("Updating proxy", end="...")
         subprocess.check_call(
             shlex.split(f"docker exec {self.fqn} nginx -s reload"),
             stdout=subprocess.DEVNULL,
             stderr=subprocess.DEVNULL
         )
+        print("Success")
 
     def upstart(self) -> List[ProxyRoute]:
         """
@@ -542,14 +544,12 @@ class Proxy(Container):
 
         The name comes from the notion of an "upsert" in programs that use databases.
         """
-        print("Updating proxy", end="...")
         routes = self.routes()
         self.update_config(routes)
         if self.status() == Status.UP:
             self.reload()
         else:
             self.start()
-        print("Success")
         return routes
 
     def print_routes(self, routes: List[ProxyRoute]):
--- a/demo
+++ b/demo
@@ -130,7 +130,7 @@ class Status(Enum):
 
 class Container:
     """
-    A single, runnable part of the AllenNLP Demo that's managed by docker.
+    A single, runnable part of the AllenNLP Demo that's managed with Docker.
     """
     def __init__(self, cid: str):
         # The container id, which is unique ID for the thing within the scope of the demo.
@@ -711,4 +711,3 @@ if __name__ == "__main__":
         sys.exit(1)
 
     args.func(args)
-
--- a/demo
+++ b/demo
@@ -655,12 +655,12 @@ def status(args: argparse.Namespace):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         prog="demo",
-        description="a utility for running the AllenNLP demo locally"
+        description="A utility for running the AllenNLP demo locally."
     )
 
     subparsers = parser.add_subparsers(
         title="command",
-        description="The command to execute.",
+        description="the command to execute.",
         dest="command"
     )
 
--- a/demo
+++ b/demo
@@ -102,6 +102,23 @@ def create_network():
         stdout=subprocess.DEVNULL
     )
 
+def printFlush(*args, **kwargs):
+    """
+    Prints a message immediately to stdout that's meant to convey the start of an action.
+    Three trailing dots are appended as a suffix to indicate things are in flight.
+    """
+    kw={ "end":"...", **kwargs, "flush":True }
+    print(*args, **kw)
+
+def printLnFlush(*args, **kwargs):
+    """
+    Print a message immediately to stdout with a trailing newline. Use this for regular output,
+    or following a `printFlush()` call to indicate that the previously started action has
+    completed.
+    """
+    kw={ **kwargs, "flush":True }
+    print(*args, **kw)
+
 class NotImplementedError(Exception):
     def __init__(self):
         super().__init__("Method not implemented.")
@@ -154,12 +171,12 @@ class Container:
         """
         Builds the container.
         """
-        print(f"Building {self.cid}", end="...", flush=True)
+        printFlush(f"Building {self.cid}")
         bp = subprocess.run(shlex.split(self.build_cmd()), encoding="utf-8", capture_output=True)
         if bp.returncode != 0:
-            print("Error", flush=True)
+            printLnFlush("Error")
             raise BuildError(self.cid, bp.stdout, bp.stderr)
-        print("Success", flush=True)
+        printLnFlush("Success")
 
     def start_cmd(self) -> str:
         """
@@ -172,10 +189,10 @@ class Container:
         Starts the container.
         """
         if self.status() == Status.UP:
-            print(f"Container {self.cid} is already running.", flush=True)
+            printLnFlush(f"Container {self.cid} is already running.")
             return
 
-        print(f"Starting {self.cid}", end="...", flush=True)
+        printFlush(f"Starting {self.cid}")
         create_network()
 
         p = subprocess.Popen(
@@ -194,22 +211,22 @@ class Container:
         # If the service still isn't healthy, then it failed to start.
         if not self.is_healthy():
             stdout, stderr = p.communicate()
-            print("Error", flush=True)
+            printLnFlush("Error")
             raise StartError(self.cid, stdout, stderr)
 
-        print("Success", flush=True)
+        printLnFlush("Success")
 
     def stop(self):
         """
         Stops the container.
         """
         if self.status() == Status.DOWN:
-            print(f"Container {self.cid} isn't running.", flush=True)
+            printLnFlush(f"Container {self.cid} isn't running.")
             return
 
-        print(f"Stopping {self.cid}", end="...", flush=True)
+        printFlush(f"Stopping {self.cid}")
         subprocess.check_call(shlex.split(f"docker stop {self.fqn}"), stdout=subprocess.DEVNULL)
-        print("Success", flush=True)
+        printLnFlush("Success")
 
     def healthcheck_url(self) -> str:
         """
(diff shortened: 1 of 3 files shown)


demo
```diff
@@ -1,5 +1,15 @@
 #!/usr/bin/env python3
 
+"""
+This script is a WIP. There's still some things TODO:
+
+- Add a command for viewing and following logs for a container.
+- Do something smart for handling `/info` requests locally. Either we have a smart local
+  proxy or we modify the UI to query info endpoint by endpoint. I lean towards the latter.
+- Test everything.
```
aimichal, line 9: :-)

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
-    Builds and starts the provided service. If the service is already running this does nothing.
+    An Enum reflecting the various states of a container.
     """
-    log = logging.getLogger("start")
+    UP = "Up"
+    DOWN = "Down"
+
+class Container:
+    """
+    A single, runnable part of the AllenNLP Demo that's managed by docker.
```
aimichal, line 133:
```suggestion
    A single, runnable part of the AllenNLP Demo that's managed with Docker.
```

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+        """
+        Returns the command, as a string, for building the container.
+        """
+        raise NotImplementedError()
+
+    def build(self):
+        """
+        Builds the container.
+        """
+        print(f"Building {self.cid}", end="...", flush=True)
```
aimichal, line 157: That's funny, haven't seen end used this way. Usu. it's as `end=""` to skip the newline. :-P

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+
+    def start(self):
+        """
+        Starts the container.
+        """
+        if self.status() == Status.UP:
+            print(f"Container {self.cid} is already running.", flush=True)
+            return
+
+        print(f"Starting {self.cid}", end="...", flush=True)
```
aimichal, line 178: There's a lot of this going on: `print("foo", flush=True)` and `print("bar", end="...", flush=True)`.

What about two helpers? `printlnFlush("foo")` (with newline) and `printFlush("bar...")` (no newline)?

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            encoding="utf-8"
+        )
+
+        # If `p.poll()` returns a value, the process has terminated. We intentionally don't timeout,
+        # we leave that to the user and Ctrl^C.
+        while p.poll() is None and not self.is_healthy():
+            delay = 5
+            time.sleep(delay)
```
aimichal, line 193:
```suggestion
            time.sleep(5)
```

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+        determined by making a HTTP request to the URL returned by healthcheck_url(). A non-200
+        response indicates that the container is not healthy.
+        """
+        # We execute the command in the container since it might not be routable from the host.
+        p = subprocess.run(
+            shlex.split(f"docker exec {self.fqn} curl --fail {self.healthcheck_url()}"),
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL
+        )
+        return p.returncode == 0
```
aimichal, line 234: healthcheck_url raises an exception, so is_healthy does too, right?

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+            return Status.UP
+        return Status.DOWN
+
+    @staticmethod
+    def all() -> List[Tuple['Container', Status ]]:
+        """
+        Returns a list of all containers that make up the AllenNLP Demo and their status.
+        """
+        out = subprocess.check_output(
+            shlex.split(f"docker ps --format {{{{.Names}}}} --filter name=^{prefix}"),
```
aimichal, line 254: Since prefix is "allennlp-demo" this will catch "allennlp-demos" right?

If the prefix is "allennlp-demo-" (trailing hyphen) then the container-name-building code will have to skip the hyphen.

Trade offs, tradeoffs. :-)

You can leave it, just thinking out loud.

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
+            encoding="utf-8"
+        )
+        running = set(out.splitlines())
+        containers = []
+        for container in [ UI(), Proxy() ]:
+            containers.append((container, Status.UP if container.fqn in running else Status.DOWN))
+        return [ *APIEndpoint.all(), *containers ]
```
aimichal, line 261: this method should return `List[Tuple['Container', Status ]]:`, but the elements it returns include `List[Tuple['APIEndpoint', Status]]`. Is this a bug, or are the types equivalent?

Oh I see, APIEndpoint is a subclass of Container. Tricky.

Why not also ask UI and Proxy for its containers?

demo
```diff
@@ -98,220 +102,561 @@ def create_network():
...
-        log.info(fmt.format(svc, status_label))
+        print(fmt.format(container.cid, status.value), flush=True)
 
 if __name__ == "__main__":
-    logging.basicConfig(level=os.getenv("LOG_LEVEL", logging.INFO), format="%(message)s")
-
     parser = argparse.ArgumentParser(
         prog="demo",
-        description="A utility that makes it easier to run a local version of the AllenNLP demo."
+        description="a utility for running the AllenNLP demo locally"
```
aimichal, line 659: So humble. This is a monster application. I mean that in a good way, it covers a lot of ground 👍

aimichal (approved):
Nice. Got a couple questions.

demo
codeviking: I'm "testing", but not thoroughly. This is a reminder to do so 😅 .

demo
codeviking: Yup! It's a little tricky, but seems to work nicely. We'll see how this plays out in practice of course!

demo
codeviking: Sure, good idea.

demo
codeviking: Oops, thanks. I used to print this out.

demo
codeviking: If `healthcheck_url()` isn't implemented, yes.

The `Container` class is an abstract class, so the `NotImplementedError()` serves to make sure that subclasses implement everything.

demo
codeviking: Yea, good point. I like including the `-` in the prefix so it's more specific.  The side effects of globbing more things into this list would be tricky to figure out.

demo
codeviking: Ah, it does so.  Just above this we do this:

```
for container in [ UI(), Proxy() ]:
            containers.append((container, Status.UP if container.fqn in running else Status.DOWN))
```

`UI()` and `Proxy()` are containers with `n` of `1`, so there's no need for a static method for creating them.

demo
dirkgr: It would be good to have a smoke test for this script in CI. We often end up with severely broken utility scripts after a while. The only way to prevent that is to have an automated test. It doesn't have to do much. If it just prints the help that's already a huge improvement, because it will make sure the imports still work.

demo
```diff
@@ -98,296 +102,611 @@ def create_network():
...
-    Returns a URL that can be used in the container for the provided service to determine if it's
-    healthy. This URL is not one that's routable from the host.
-    """
-    if svc == "ui":
-        return "http://localhost:3000"
-    else:
-        return "http://localhost:8000"
+class NotImplementedError(Exception):
+    def __init__(self):
+        super().__init__("Method not implemented.")
```
dirkgr, line 107: Python has a built-in `NotImplementedError`.

demo
dirkgr: You don't want to use `logging` for that?

demo
codeviking: Yea, that makes sense.

I actually intended to replace the existing integration test mechanism, that uses `docker-compose`, with this mechanism.

Right now we have integration tests that start the UI (pointed at production) using `docker-compose` and then run a [single, silly test](https://github.com/allenai/allennlp-demo/blob/master/it/cypress/integration/rc_spec.ts). If we use the script to start the UI in this instance, we *should* make sure that most of the code here works.

I could also go a bit further and either start an API endpoint too as part of that test, or add a separate smoketest all together for this script that does a few things like that (and uses `curl` to make sure things respond with non-200 responses as expected).

WDYT?

dirkgr (commented):
I might have more comments later, but right now we have to meet about something else ...

demo
codeviking: Oops, thanks!

codeviking:
For some reason GitHub won't let me respond inline to this, so I will here:

> You don't want to use logging for that

I did originally, but opted out as:

1. I couldn't figure out to omit the trailing `\n` when using `self.log()`.
2. It felt wrong to try to do ☝️ , as logs aren't supposed to be used for that type of output.

Does that makes sense to you too?

dirkgr:
`logging` is not a way to get what you want. You would use `logging` to provide diagnostic output, and some sort of progress update, for a part of a batch job. I tend to think of all my code that way, but I sense that you're thinking of this as a UI within the confines of the command line. For that, `print()` is the way to go.

demo
```diff
@@ -98,296 +102,611 @@ def create_network():
...
+        return f"""
+            docker run
+                --rm
+                --name {self.fqn}
+                --env FLASK_ENV=development
+                --network {network}
+                --volume {self.path}:/app/allennlp_demo/{self.cid}
+                --volume {paths.api_modules / "common"}:/app/allennlp_demo/common
+                --volume {home / ".allennlp"}:/root/.allennlp
+                --volume {home / ".cache/huggingface"}:/root/.cache/huggingface
```
dirkgr, line 359: This right here is an underappreciated aspect of this script. Not having to download all the models every time? Yes please!

demo
```diff
@@ -98,296 +102,611 @@ def create_network():
...
+        """
+        routes = []
+        for endpoint, status in APIEndpoint.all():
+            if status != Status.UP:
+                continue
+            urlpath = endpoint.urlpath()
+            routes.append(ProxyRoute(
+                urlpath,
+                f"http://{endpoint.fqn}:8000",
+                [f"rewrite {urlpath}(/(.*))? /$2 break"],
```
dirkgr, line 454: Does this work if `urlpath` contains regexy characters?

dirkgr (approved):
Only minor comments

demo
codeviking: Done, thanks!

demo
codeviking: Done!

demo
codeviking: Good question, we'll find out when that happens :).

My guess is it'd break something, because we reference the capturing groups in the rewrite.

dirkgr: Wait, you could have fixed this ahead of time: https://docs.python.org/3/library/re.html#re.escape

codeviking: Ah, nice find!

I'm not too worried about this case. The URL path is produced by the model id.  I think it's pretty unlikely for the model id to contain any regexpy characters, since it's meant to be a human readable string and to be URL safe.

That said it's also just a function call, so if you think I definitely should throw this in let me know and I'll do so.

PR #847, opened 2021-02-18
By default, only show status for containers that are running.

codeviking:
This is a small change to the `status` command. Now we only show
the status for containers that are running by default. If the
`--all` flag is added we show the status of everything.

### DIFF
--- a/demo
+++ b/demo
@@ -708,7 +708,7 @@ def status(args: argparse.Namespace):
     printLnFlush(fmt.format("Container", "Status"))
     printLnFlush(fmt.format("-" * spaces, "------"))
     for container, status in containers:
-        if args.only is not None and args.only.lower() != status.value.lower():
+        if status == Status.DOWN and not args.all:
             continue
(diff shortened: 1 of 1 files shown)


demo
```diff
@@ -674,7 +674,7 @@ def status(args: argparse.Namespace):
     printLnFlush(fmt.format("Container", "Status"))
     printLnFlush(fmt.format("-" * spaces, "------"))
     for container, status in containers:
-        if args.only is not None and args.only.lower() != status.value.lower():
+        if status == Status.DOWN and not args.all:
```
aimichal, line 677: Took me a second to parse this.

How about "if args.all or status != Status.DOWN: println..."? I think that'll read more fluently, though it's not the fail-fast conditional inversion that's popular in Go.

Just a thought!

aimichal (approved):

demo
codeviking: Yea, that makes sense -- good call.

PR #848, opened 2021-02-18
Restore the restart command.

codeviking:
I removed the restart command in a previous PR for simplicity,
but in doing a little work today using the tool discovered I
think it's pretty useful. This PR brings it back.

### DIFF
--- a/demo
+++ b/demo
@@ -662,6 +662,32 @@ def stop(args: argparse.Namespace):
 
     proxy.print_routes(proxy.upstart())
 
+
+def restart(args: argparse.Namespace):
+    """
+    Restarts the specified containers by stopping them, building and then starting them.
+    """
+    try:
+        cids: List[str] = args.container
+
+        # If no ids are provided, restart all running containers
+        if len(cids) == 0:
+            cids = [ c.cid for c, s in Container.all() if s == Status.UP ]
+
+        for cid in filter_cids(cids):
+            container = Container.from_id(cid)
+            container.stop()
+            container.build()
+            container.start()
+        proxy = Proxy()
+        proxy.print_routes(proxy.upstart())
+    except (StartError, BuildError) as err:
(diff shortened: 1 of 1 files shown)


demo
```diff
@@ -662,6 +662,32 @@ def stop(args: argparse.Namespace):
...
+
+        # If no ids are provided, restart all running containers
+        if len(cids) == 0:
+            cids = [ c.cid for c, s in Container.all() if s == Status.UP ]
+
+        for cid in filter_cids(cids):
+            container = Container.from_id(cid)
+            container.stop()
+            container.build()
+            container.start()
```
aimichal, line 681: I wonder if it makes sense to stop all matching containers first, then build, then start.

The reason is that if I have 5 containers running, and I want to restart them all, it might be a confusing experience if (say) the UI container is rebuilt and restarted, but makes calls to an API that hasn't been built yet. (This is actually theoretical, I have no idea if that applies to local dev for AllenNLP Demo.)

On the other hand, if you stop all and build the first, but it fails, then nothing is running. Which may be okay, but may be not? I dunno.

Also, do these .stop() and .start() operations show what they're doing? (I think they do, just checking.)

aimichal (approved):

demo
codeviking: > Also, do these .stop() and .start() operations show what they're doing? (I think they do, just checking.)

Yup, they do:

```
❯ ./demo restart
Stopping ui...Success
Building ui...Success
Starting ui...Success
Updating proxy...Success

Path                     	Target
-------------------------	-------------------------
http://localhost:8080/   	Local (ui)
http://localhost:8080/api	https://demo.allennlp.org
```

> I wonder if it makes sense to stop all matching containers first, then build, then start.

Good point. I think you're right -- stopping the world first prevents any confusion when something only partially works because the changes triggered by a restart are in flight. I'll update the order of operations shortly!

PR #849, opened 2021-02-18
Fetch info model by model.

codeviking:
This modifies the front-end to query info for each model on a
per-model basis. This is probably a bit slower, but also means
that when running the stack locally the info is provided by
the local version that's running instead of production.

If we find it's too slower for users we can probably find other
ways to speed things up.

I also removed the aggregate `/api/info` route. Nothing uses it
after this change, and it involves all sorts of complicated
machinery, so it felt ok to remove.


demo
```diff
@@ -471,16 +471,25 @@ class Proxy(Container):
...
+                    urlpath,
+                    f"http://{endpoint.fqn}:8000",
+                    [f"rewrite {urlpath}(/(.*))? /$2 break"],
+                    local_cid=endpoint.cid
+                ))
+            else:
+                # We can't use a single catch-all that forwards all requests to `/api`
+                # to production because some models share a prefix. For instance, `/api/bidaf`
+                # and `/api/bidaf-elmo` would both get sent to the same place if we made
+                # a rule matching the former. We could affix a trailing `/` to resolve this,
```
aimichal, line 487:
```suggestion
                # a rule matching `/api`. We could affix a trailing `/` to resolve this,
```

demo
```diff
@@ -471,16 +471,25 @@ class Proxy(Container):
...
+                # but that makes NGINX behave in weird ways with respect to whether the 
+                # trailing URL path is included in the forwarded request or not. So we avoid
+                # all of these headaches by simply having a route for everything that goes
+                # where we want it to.
+                routes.append(ProxyRoute(urlpath, "https://demo.allennlp.org"))
```
aimichal, line 492: How will this work with Skiff app environments?

aimichal (approved):

demo
codeviking: You mean, if someone were to deploy an ad-hoc and want to point their local machine at that instance?

Right now that's not supported, but we could probably make a way to do that. Something like:

```
./demo proxy bidaf https://bidaf-adhoc.apps.allenai.org
```

aimichal: Fix it when/if it comes up, no worries!

PR #852, opened 2021-02-19
Add a header for debugging.

codeviking:
This adds a custom header, `X-Served-By` that allows local clients
to determine what backend is handling a request. My intuition is
this will debug issues.

### DIFF
--- a/demo
+++ b/demo
@@ -534,7 +534,7 @@ class Proxy(Container):
 
                     # Add a header that clients can use to determine where a response is
                     # coming from.
-                    add_header X-Served-By {route.upstream};
+                    add_header Via {route.upstream};
 
                     {extra}
                 }}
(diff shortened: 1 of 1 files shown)


demo
```diff
@@ -531,6 +531,11 @@ class Proxy(Container):
                     {rewrites}
                     proxy_pass {route.upstream};
                     proxy_set_header X-Forwarded-For $remote_addr;
+
+                    # Add a header that clients can use to determine where a response is
+                    # coming from.
+                    add_header X-Served-By {route.upstream};
```
aimichal, line 537: Old habits are hard to shake:

> Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012 ...

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

aimichal (approved):

demo
codeviking: Hah, yup!

Ok, maybe I'll use the [Via](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via) header instead?

codeviking: I did that.

PR #855, opened 2021-02-19
Finalize the new demo script.

codeviking:
This makes the new demo script official by updating the documentation
to reference it, updating CI to use it and by removing the
`docker-compose` files which shouldn't be needed anymore.


aimichal (approved):
# 👏 

Super nice!

PR #868, opened 2021-02-25
Ignore empty directories.

codeviking:
If we delete a module using `git` it'll leave behind an empty directory
which can make the `demo` script fail in confusing ways. This fixes
this by silently ignoring directories that are empty.

### DIFF
--- a/demo
+++ b/demo
@@ -459,6 +459,8 @@ class APIEndpoint(Container):
                 continue
             if p.name == "common" or p.name.startswith(".") or p.name == "__pycache__":
                 continue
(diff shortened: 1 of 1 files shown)


dirkgr (approved):
Sure, that's good. But you could be more specific even and ignore directories that don't contain a `model.json` file?

codeviking:
> Sure, that's good. But you could be more specific even and ignore directories that don't contain a model.json file?

I could do that, but I figure there might be a scenario where the user has added a Python module that they intend to run using this script, but forgotten to add the `model.json` file. This way the demo script will fail and note the absence of that script.

PR #869, opened 2021-02-25
fix collapse panel issue

jonborchardt:
antd has a broken component, this is a quick hacky fix to deal with that
simply forcing the content area of a panel to be 100% tall

### DIFF
--- a/ui/src/components/CollapsePanel.tsx
+++ b/ui/src/components/CollapsePanel.tsx
@@ -1,0 +1,13 @@
+import styled from 'styled-components';
+import Collapse from 'antd/es/collapse';
+
+/**
+ * The latest version of antd has a bug with setting the height of a dynamic collapse panel.
+ * This is a hammer to fix that.
+ * TODO: we should remove this when we can.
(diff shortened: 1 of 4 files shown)


ui/src/components/CollapsePanel.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+import Collapse from 'antd/es/collapse';
+
+/**
+ * The latest version of antd has a bug with setting the height of a dynamic collapse panel.
+ * This is a hammer to fix that.
+ * TODO: we should remove this when we can.
+ */
```
codeviking, line 8: Can we link to the associated GitHub issue here?

codeviking (approved):
Are we sure this isn't a side-effect of the styles we're applying to the parent?

jonborchardt:
> Are we sure this isn't a side-effect of the styles we're applying to the parent?

yes.  ant is now setting an inline style based on computed content... its crazy
im sure they will revert/fix soon

PR #870, opened 2021-02-27
Fix the run command.

codeviking:
This fixes a bug Dirk discovered in the run command. The
`--entrypoint` option was being passed to the command being
ran in the container instead of `docker run`.

### DIFF
--- a/demo
+++ b/demo
@@ -220,8 +220,9 @@ class Container:
             cmd = shlex.split(f"{self.exec_cmd()} {entrypoint}")
         else:
             cmd = shlex.split(self.start_cmd())
-            cmd.append("--entrypoint")
-            cmd.append(entrypoint)
+            # The tag to run is always the last argument, so we insert the `--entrypoint`
+            # before then.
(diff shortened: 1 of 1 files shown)


dirkgr (approved):

demo
```diff
@@ -220,8 +220,9 @@ class Container:
             cmd = shlex.split(f"{self.exec_cmd()} {entrypoint}")
         else:
             cmd = shlex.split(self.start_cmd())
-            cmd.append("--entrypoint")
-            cmd.append(entrypoint)
+            # The tag to run is always the last argument, so we insert the `--entrypoint`
+            # before then.
+            cmd = cmd[:len(cmd)-1] + [ "--entrypoint", entrypoint ] + cmd[len(cmd)-1:]
```
dirkgr, line 225: `cmd[:len(cmd)-1]` could just be `[cmd[0]]`, no?

dirkgr (approved):

demo
dirkgr: I'll try that in my own branch.

PR #888, opened 2021-03-12
Upgrade all demos to 2.0

dirkgr:
This can't be merged until https://github.com/allenai/allennlp-models/pull/227 is merged.

Still missing:
 - [ ] Order of questions for RC needs to be changed so that SQuAD questions come first


demo
```diff
@@ -160,8 +160,8 @@ class Container:
         """
         Builds the container.
         """
-        printFlush(f"Building {self.cid}")
-        bp = subprocess.run(shlex.split(self.build_cmd()), encoding="utf-8", capture_output=True)
+        printLnFlush(f"Building {self.cid}")
+        bp = subprocess.run(shlex.split(self.build_cmd()))
```
dirkgr, line 164: This makes the demo script print the build steps to console.

dirkgr:
Actually, we can change the order of the questions independently of this PR. @jonborchardt, can you give it a review? I just merged the corresponding change in -models.

demo
jonborchardt: 👍

ui/src/demos/reading-comprehension/types.ts
```diff
@@ -163,59 +163,18 @@ export const isNAQANetPredictionArithmetic = (
...
-        typedPred.program_nested_expression !== undefined
-    );
-};
-
 export type Prediction =
     | BiDAFPrediction
     | TransformerQAPrediction
-    | NAQANetPrediction
-    | NMNPrediction;
+    | NAQANetPrediction;
```
jonborchardt, line 169: there is a glob of code written in ui
all of these files should be deleted:
https://github.com/allenai/allennlp-demo/tree/master/ui/src/demos/reading-comprehension/nmn

this makes me so sad, that took me a long time to write!

jonborchardt (approved):
be sure to also clean up https://github.com/allenai/allennlp-demo/tree/master/ui/src/demos/reading-comprehension/nmn

im so sad this is getting deleted! i JUST wrote it!

ui/src/demos/reading-comprehension/types.ts
dirkgr: 😢

dirkgr:
So since I was already bringing back NMN, I also brought back semantic parsing. Though it still doesn't have a front-end, and I think that's by design.

dirkgr:
@jonborchardt, do you want to review this again?

jonborchardt (approved):
lg2m, did you test that each demo returns predictions and attacks work?

dirkgr:
I just tested predictions. But @schmmd has said for a long time that he wanted to test the demo :-P

Should I spin up an environment?

dirkgr:
I would [spin it up right now](https://marina.apps.allenai.org/a/allennlp-demo/e/create), but I don't want it running all night while nobody is looking anyways.

schmmd:
@dirkgr why don't you deliver to prod, and then I can take a look there ;-)

dirkgr:
I deployed this to https://allennlp-demo.demomodelsweep.apps.allenai.org/, and it does not work one bit. Glad I tried it first. I'm not sure what's wrong, but I don't have time this week anymore, so I'm putting this aside until next week.

dirkgr:
We believe this is a problem with the ad-hoc, not a problem with this code. So we're shipping it!

PR #896, opened 2021-03-18
Ordered RC examples

dirkgr:
This makes it so examples for the reading comprehension tasks are ordered the way they appear in the tasks file. At least, that's what it'll do when it's done.

This is part of a multi-repo nightmare. We need
 * https://github.com/allenai/allennlp/pull/5059
 * https://github.com/allenai/allennlp-models/pull/229
 * https://github.com/allenai/allennlp-demo/pull/896

all at the same time.


dirkgr:
@jonborchardt, this currently stalls out when trying to read the response from the tasks service, as expected. The format is actually defined in tugboat, not here.

codeviking:
> The format is actually defined in tugboat, not here.

Maybe we should move the definition to back to the AllenNLP demo. I'm not sure it makes sense to live in Tugboat, given it's an AllenNLP specific definition of what a task is.

@jonborchardt, what do you think?

codeviking:
@dirkgr I'm not seeing an error when I pull this down and run things locally...but the examples for RC appear to be in the same order as they're in now. So either I'm doing something wrong, or there's an uncommitted change or some other ghost-in-themachine somewhere in the mix?

dirkgr:
Sorry, I had some untracked files in the directory.

dirkgr:
I ran this just now and it worked right away. I don't know why I thought it was broken before. Will now try to orchestrate the reviews across three repos 🤪.

dirkgr:
I'm getting this Cypress error. Apparently it can't find the description for reading comprehension tasks in the output of the main page.

When I `curl`, I just get a warning that I need to enable JavaScript. When I pull the main page with a browser, the description of RC tasks is there. I am not sure how to proceed. What is Cypress actually doing?

codeviking:
Did the text change?

The [test loads the page in a browser, and looks for the text that's defined here](https://github.com/allenai/allennlp-demo/blob/master/it/cypress/integration/rc_spec.ts#L11).

If the text is present but this isn't finding it, then something else is wrong. If the text changed we might just need to update
that string.

I'll try this out shortly and let you know what I figure out.

codeviking:
It appears to be something that's CI specific, as they work locally. I cut https://github.com/allenai/allennlp-demo/issues/903, someone from ReViz will look into this next week, but for now it's ok to ignore the failures as they're spurious.

dirkgr:
Sweet, it's live and it works!

PR #912, opened 2021-03-29
add temp logging to ui tests

jonborchardt:


codeviking:
We weren't able to reproduce the spurious failures, so it's ok to close this for now.

Issue #876, opened 2021-03-01
Handle lack of gradient input field cleanly.


codeviking:
Our code assumes the existence of the `grad_input_1` field, but sometimes it's not present, according to https://sentry.io/organizations/ai2/issues/2218627224/?project=5599301&query=is%3Aunresolved&sort=user&statsPeriod=14d).

We should try to figure out why this is happening and probably make the UI code more defensive (if it's something we should expect). It's also possible that it's something we can fix in the backend.

codeviking:
@leo-liuzy Are you taking a look? If not let us know and ReViz on-call can take a look.

PR #913, opened 2021-03-30
Jon fix missing grad

jonborchardt:
this pr makes the instance grad interface shared, more general, and deals with missing values.
fixes: https://github.com/allenai/allennlp-demo/issues/876

note, the first 5 commits are reverted...


codeviking (approved):
Type guards for the win. Nicely done.

PR #936, opened 2021-04-14
Jon fix noop

jonborchardt:
this should fix dirk's issue


dirkgr (approved):
LGTM, but tests are failing.

dirkgr:
Looks like it's not a real failure.

PR #938, opened 2021-04-22
Dependabot roundup

dirkgr:
This combines a bunch of recent dependabot changes. There are some more, but they did not merge cleanly, so I didn't bother. Dependabot will reproduce them if necessary.

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,4 +1,4 @@
 git+git://github.com/allenai/allennlp.git@v2.3.1
-git+git://github.com/allenai/allennlp-models.git@v2.3.1
+git+git://github.com/allenai/allennlp-models.git@v2.3.0
 Flask==1.1.2
 pytest==6.2.3
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,4 +1,4 @@
 git+git://github.com/allenai/allennlp.git@v2.1.0
 git+git://github.com/allenai/allennlp-models.git
 Flask==1.1.2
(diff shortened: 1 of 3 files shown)


codeviking (approved):

api/requirements.txt
```diff
@@ -1,4 +1,4 @@
 git+git://github.com/allenai/allennlp.git@v2.1.0
 git+git://github.com/allenai/allennlp-models.git
```
epwalsh, line 2: Why do we have allennp pinned to a release and allennlp-models pinned to nothing?

dirkgr: I don't remember. But we could pin both to the latest to get around this.

api/requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+git://github.com/allenai/allennlp.git@v2.1.0
-git+git://github.com/allenai/allennlp-models.git
+git+git://github.com/allenai/allennlp.git@v2.3.1
+git+git://github.com/allenai/allennlp-models.git@v2.3.1
```
epwalsh, line 2: ?

```suggestion
git+git://github.com/allenai/allennlp-models.git@v2.3.0
```

dirkgr: Oh, I didn't even check. I thought you always ship those in tandem.

epwalsh: We decided a while ago that we could do patch releases in the core repo when needed without a corresponding models release

PR #967, opened 2021-04-27
Fix the demo script.

### LINKED ISSUE #966: The demo script isn't working anymore.
After running `./demo start` I see this in a browser:

Screen Shot 2021-04-27 at 2 53 58 PM

It looks like the endpoints that are supposed to be sending requests to production are failing:

Screen Shot 2021-04-27 at 2 54 33 PM

codeviking:
This fixes the demo script by setting the server name
when negotiating a SSL connection. I'm not sure why
things stopped working -- my guess is something changed
in Google LoadBalancers (which Skiff uses) which makes
the SSL handshake more restrictive (which seems like
a good thing).

This fixes #966.

### DIFF
--- a/demo
+++ b/demo
@@ -33,6 +33,7 @@ from typing import List, Tuple, Optional
 from pathlib import Path
 from urllib.request import urlopen
 from urllib.error import URLError
+from urllib.parse import urlparse
 from http.client import HTTPResponse, RemoteDisconnected
 
 # When we spawn docker things they need a name that prevents them from colliding with other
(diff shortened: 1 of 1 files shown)


dirkgr (approved):

demo
```diff
@@ -582,11 +583,20 @@ class Proxy(Container):
...
+                extra += f"""
+                    proxy_ssl_name {host};
+                    proxy_ssl_server_name on;
+                """
             locs.append(f"""
                 location {route.path} {{
                     {rewrites}
-                    proxy_pass {route.upstream};
                     proxy_set_header X-Forwarded-For $remote_addr;
+                    proxy_pass {route.upstream};
```
codeviking, line 599: Oops. This is noise. I'm going to revert this so this doesn't get interpreted later as if it's impactful.

PR #977, opened 2021-05-06
varnish dropdown is fixed

jonborchardt:
@aimichal varnish is ready for you
@dirkgr you can upgrade antd as desired now


codeviking (approved):
👍

PR #1049, opened 2021-07-12
Add a pod disruption budget.

codeviking:
This change does two things:

1. Adds a `PodDisruptionBudget`, which ensures that a single replica
   is always available. This prevents downtime while updating the
   cluster version.
2. Changes the deployment spec so that Kubernetes can deploy 2X the
   number of replicas. This should make things a little faster and
   improve capacity during deploys.


aimichal (approved):
:-)

epwalsh (approved):

PR #1054, opened 2021-07-20
Jon update deps

jonborchardt:


aimichal (approved):
I assume the demo still runs after these changes?

PR #1225, opened 2022-07-29
Remove Sentry.

codeviking:
We're closing the account.

https://github.com/allenai/reviz/issues/440


jonborchardt (approved):

codeviking:
The demo is failing because `@allenai/tugboat` depends on `@sentry/react`. So I'll have to remove it first there, publish a new version, and then use that version here.

codeviking:
This is close to working. But there's a JS error because a polyfill is missing now:

```
Module not found: Error: Can't resolve 'path' in '/ui/node_modules/@allenai/tugboat/components
```

We might be able to include a shim locally. But we should probably fix this upstream -- either by including the shim there or removing the `path` dependency.

jonborchardt (approved):

PR #1284, opened 2022-11-30
Migrate Ingress to use networking.k8s.io/v1 API

mewil:


codeviking (approved):
This looks good. But I think there's another part to the demo. I believe we instantiate an ingress per backend, with some fancy path munging embedded there.

https://github.com/allenai/allennlp-demo/blob/main/api/allennlp_demo/common/.skiff/common.libsonnet#L247-L288

That said I'm pretty sure this as is would work. So you could update that in another pass.

mewil:
Thanks for spotting that! I've gone ahead and update that in this PR too.

codeviking (approved):
I think we can simplify the path. Otherwise LGTM.

